PR #1, opened 2015-10-03
Fix in CallExternalCommand: Exception message "Error calling external…

viper3400:
Exception message "Error calling external command" was shown although CallExternalCommand has finished successful.

This is reproducible by creating a simple batch file containing:

_copy %1 C:\TEMP_

When adding this as external command the output of "CallExternalCommand" will be something like:

D:\WORK\greenshot\Greenshot\bin\Debug>copy "C:\Users\Jan\AppData\Local\Temp\2015_10_03_16_08_28_Greenshot_Running_Microsoft_Visual_Studio.png" C:\TEMP
        1 Datei(en) kopiert.

So output string is not empty and MatchCollection uriMatches = URI_REGEXP.Matches(output) is executed. I don't analysed what this regular expression should check, in fact uriMatches != null but uriMatches.Count is 0. There is no match, so I suppose uriMatches.Count should be > 0, not >= 0?

When accessing uriMatches[0] an System.ArgumentOutOfRangeException was raised:

Message = "Das angegebene Argument liegt außerhalb des gültigen Wertebereichs.\r\nParametername: i"

In addition log4net message has been written before exportInformation.ErrorMessage was set, so exception message was not in log file.

### DIFF
--- a/GreenshotExternalCommandPlugin/ExternalCommandDestination.cs
+++ b/GreenshotExternalCommandPlugin/ExternalCommandDestination.cs
@@ -121,7 +121,7 @@ namespace ExternalCommand {
 						if (config.OutputToClipboard) {
 							ClipboardHelper.SetClipboardData(output);
 						}
-						if (uriMatches != null && uriMatches.Count >= 0) {
+						if (uriMatches != null && uriMatches.Count > 0) {
 							exportInformation.Uri = uriMatches[0].Groups[1].Value;
 							LOG.InfoFormat("Got URI : {0} ", exportInformation.Uri);
 							if (config.UriToClipboard) {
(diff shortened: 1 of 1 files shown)


Lakritzator:
In 1.3 I changed a lot of the code anyway, but for 1.2 this is certainly something usefull.
Thanks, I will merge it and build 1.2.7 together with a couple of other fixes.

Lakritzator:
So, it's made available.

Sorry for the long wait, I am a bit occupied with Greenshot 1.3.x
This should fix some old .NET 2.0 problems by upgrading to 4.5 and more importantly: make it a LOT easier to develop plug-ins/addons/extension (whatever one wants to call it) for Greenshot.
For now it's a highly volatile branch... but in a few weeks it should be better.

Hope to hear more of you :)

PR #10, opened 2016-03-14
language-cs-CZ.xml update

PetrToman:
Sending and update of CZ language file. [language-cs-CZ.zip](https://github.com/greenshot/greenshot/files/172354/language-cs-CZ.zip)


Lakritzator:
Not really a useful pull request :)

PR #15, opened 2016-03-23
Change the GPL text to GPLv3

Lakritzator:
Text comes from here: https://www.gnu.org/licenses/gpl-3.0.txt


Lakritzator:
Any reason not to change?

PR #37, opened 2016-11-22
kab localization

belkacem77:
Kabyle localization


Lakritzator:
Impressive how many languages there are in the world, I never heard of this one.
Happy to merge this for you, many thanks!!

I will need to merge this manually into our current release, release/1.2.9, due to the fact that develop has a lot of changes... and I will need to modify the installer.

But your translations will be inside 1.2.9 RC4 which should be coming soon.

We will also change the process of supplying translations. At least we should separate it from the code :smile:

Thanks again, the rest of the pull requests will be merged too

Lakritzator:
Oh, I didn't see this yet... but what about the main translation (language-kab-dz.xml), without that there will be no way to select the language.

belkacem77:
Hi Lakritzator,

Kabyle is a north African Language. One of Berber languages.
Le iso code is kab.
Many tools and solutions are localized into Kabyle among them Firefox,Linux Mint, Odoo....

May be we will get involved in coding later.
Thanks for merge.

PR #54, opened 2017-01-25
add default effects when capture done.

xucongli1989:
Add default border with color.
ps:you can ignore 'readme.md' and unuseful for yours  when merging.


Lakritzator:
This is really cool, thanks a lot!!!!!!!

I will review this, right after releasing my bugfix 2.
You did commit a lot of changes which have nothing to do with the change itself, and I need to check the effect that your changes have on the configuration. Also I currently don't want another dependency (Json.NET) for the 1.2.x version so I will need to check if it works with the build in Json code.
Another thing is the UI, as we are heavily building a new Greenshot I would prefer not to have any UI changes which mean translations etc... but I will first look at it.

Thanks again.

Lakritzator:
I didn't have time to see it live YET, but I expect there will be issues when someone uses multiple destinations. e.g. Clipboard & file
The capture will probably have the "border" effect applied multiple times and depending on the order, the first is okay the second has it 2x, the third 3x etc

xucongli1989:
you can try it from:https://github.com/xucongli1989/greenshot/releases

follow your mind to use my commits,and it does not matter.

PR #74, opened 2017-09-30
Being able to set window title in the image editor.

AndersBillLinden:
When I have a lot of image editors open, which is the case when I have created some screenshots that I dont want to close, all of those editors window titles reads "Greenshot image editor". Then I have to search for the right window in the taskbar by clicking them one by one. This change makes it possible to set a window title on each of them.

### DIFF
--- a/Greenshot/Forms/ImageEditorForm.Designer.cs
+++ b/Greenshot/Forms/ImageEditorForm.Designer.cs
@@ -119,6 +119,7 @@ namespace Greenshot {
 			this.saveElementsToolStripMenuItem = new GreenshotPlugin.Controls.GreenshotToolStripMenuItem();
 			this.loadElementsToolStripMenuItem = new GreenshotPlugin.Controls.GreenshotToolStripMenuItem();
 			this.pluginToolStripMenuItem = new GreenshotPlugin.Controls.GreenshotToolStripMenuItem();
+			this.titleToolStripMenuItem = new GreenshotPlugin.Controls.GreenshotToolStripMenuItem();
 			this.helpToolStripMenuItem = new GreenshotPlugin.Controls.GreenshotToolStripMenuItem();
 			this.helpToolStripMenuItem1 = new GreenshotPlugin.Controls.GreenshotToolStripMenuItem();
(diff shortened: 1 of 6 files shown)


Lakritzator:
Hey,

thanks for sharing this with us!

I understand your issue, but having to name the editor title manually is not really something I would like our users to do. In fact, I think not having a title is plainly a "bug"... but it should have an automatically generated title! (The title will change as soon as you save the capture to a file)

Besides that, having another dialog is a lot of work for us, as we need to translate it in a lot of languages 😄 

So I hope you can agree on the fact that we should fix this differently?

Best wishes,
Robin

AndersBillLinden:
When I used greenshot for 20 screenshots of different minecraft recipes, it had been a great thing to be able to tell the windows apart in the taskbar, so the change is totally demand driven.

AndersBillLinden:
Maybe the title could be automatically generated but also settable. After all, the icon itself already tells its a greenshot image editor. But what should such a title encompass? The screenshot dimensions?

AndersBillLinden:
...or maybe greenshot could detect which window the screenshot came from and set the title to something like "Outlook, 50x60"

Lakritzator:
That is exactly what it should already do, but it's implemented on the wrong location in the capture flow and besides is the current implementation not very smart when a capture has multiple windows on the result.

AndersBillLinden:
In the case I mentioned, automatic title generation had not been so useful compared to custom title. If anyone sais I am using the software to things it was not intended to, I am not late to agree. But I am not alone using screenshots as a "memo".

AndersBillLinden:
Some programs adds a trailing text after a hyphen to the dynamically set title, like "My webpage - Google Chrome", or if you have the vs extension "Customize Visual Studio Title", you get a title like "FooSolution - Microsoft Visual Studio". If you think that would be a good way to recognize the window.

Lakritzator:
I am currently blocked on a code signing issue, which prevents updates to Greenshot all together.
But to track this issue, I created https://greenshot.atlassian.net/browse/FEATURE-1099
It would be nice if you registered yourself as a watcher.

PR #75, opened 2017-10-18
Do not show thank you page if installer was run silently

SmuSmu:
I understand your points in http://getgreenshot.org/faq/how-can-i-avoid-greenshot-opening-a-browser-window-at-the-end-of-the-installation-process/

But can we stop showing this site for silent deployments. Reason is I as admin want to deploy always the latest version. But I do not want to disturb people. So I want so do a silent update. But if I do this with your installer a unwanted window will pop up.

BTW
Current situation is that I still deploy Version 0.8.0 where this stuff was not shown. I think thats something you also don`t like to see. ;-)

### DIFF
--- a/Greenshot/releases/innosetup/setup.iss.template
+++ b/Greenshot/releases/innosetup/setup.iss.template
@@ -651,7 +651,7 @@ end;
 Filename: "{code:getNGENPath}\ngen.exe"; Parameters: "install ""{app}\{#ExeName}.exe"""; StatusMsg: "{cm:optimize}"; Flags: runhidden runasoriginaluser
 Filename: "{code:getNGENPath}\ngen.exe"; Parameters: "install ""{app}\GreenshotPlugin.dll"""; StatusMsg: "{cm:optimize}"; Flags: runhidden runasoriginaluser
 Filename: "{app}\{#ExeName}.exe"; Description: "{cm:startgreenshot}"; Parameters: "{code:GetParamsForGS}"; WorkingDir: "{app}"; Flags: nowait postinstall runasoriginaluser
-Filename: "http://getgreenshot.org/thank-you/?language={language}&version={#Version}"; Flags: shellexec runasoriginaluser
+Filename: "http://getgreenshot.org/thank-you/?language={language}&version={#Version}"; Flags: shellexec runasoriginaluser skipifsilent
 
 [InstallDelete]
 Name: {app}; Type: dirifempty;


kantholy:
+1

Lakritzator:
@kantholy (and @SmuSmu):
Just so things are clear: I **currently** will not merge this PR.

By doing so, we will miss revenue! Without revenue, I need to spend my time on different things to keep my head above water. We would also have been able to move Greenshot forward a lot faster, if there would be more money to free up more time.

Before you read on, I do not want to provoke something but I do want to point out something that is mostly ignored:

This change is mainly useful for commercial environments, but most companies are using Greenshot **without** supporting us financially. It's even so that a LOT (most!) of the reported bug tickets & feature requests clearly come from "commercial " users and these make up most of our time.

Just consider this yourself, although I don't like finger pointing, but it's very convenient that Greenshot costs nothing.. right? So let's roll it out on thousands of PC's, yes there are even companies with a lot more who use Greenshot without ever having spent us a single dime, and if there are "issues" they expect support for it...

Now, we am more than happy to solve issues, what e do not want to do is reduce the little income we make with Greenshot. 

What would be a possible solution is to change the Greenshot license so commercial usage costs, and at the same time we can merge this PR. Would that be okay?
I might be looking to much inside my "box", and I am very interested in other suggestions & opinions!

We are also **considering** going into the Microsoft App-Store, would that interest you?

I also want to add a solution which can be used right away! If you want a "quick" solution, you can always use the .zip and "deploy" it yourself:
http://getgreenshot.org/faq/in-which-cases-should-i-use-the-zip-package-instead-of-the-installer/

BTW.
Not a good idea to use 0.8.0, but if you want to use it that is up to you... we don't support it. We don't care much about what a company installs for their users, we don't get any money for it.
We of course like to give our users the best experience, but **we** don't force 0.8 upon them.

Best wishes,
Robin

hudsonm62:
Hi @Lakritzator,

Any chance of this being looked at again?
The zip is no longer available as of 1.3. Which is fine, but it now makes auto deployment unrealistic with some tools. (i.e. reporting as failed install when browser not available, time outs, etc.)

Using something like /verysilent should, indeed, be very silent. Whilst I entirely understand the need for revenue, intentionally making a flag like /verysilent behave in a way that contradicts expectations can create friction for system admins and may discourage wider enterprise adoption — which in turn _could_ reduce opportunities for financial support from commercial users.

Perhaps theres a middle ground we could find here?

---

Related issues: #610 #470 #609

jklingen:
@hudsonm62 JFYI we'll probably bring back the ZIP, but our top priority was to publish the long-awaited release at all.

Neustradamus:
I think, it must no be automatic, a page is opened only if user accept it.

A checkbox needs to be added.

FF-Brown:
> I **currently** will not merge this PR.

The answer is pretty clear, y'all.

PR #77, opened 2018-01-25
a new plugin for a Chinese  picture bed website

harlanc:
My requirements ：
When I record my blogs,I need screenshots.So I need to select a figure bed service to store my figures.QINIU is one of the most popular figure bed website in china.It is free.So it is my select.

But followings are my steps to use a picture when I write my blog.

1. cutting a picture.serval steps
2. and then upload to websites,serval steps manually.
3.get the picture URL and paste it into my blog.

so complicated.

ScreenShot is my daily screenshot tools.So I implement a Plugin for QINIU.


harlanc:
I think if this plugin is added in screenshot,then it will be more popular in china.

Lakritzator:
Very cool, but we have a lot of bad experience with Imgur.
As there are many people using Greenshot, there are many unforseen side effects:

1. If there is a problem at Imgur, we are getting issues which we cannot solve while (see next)
2. It seems Imgur does not feel responsible for helping when there are problems, there support is very poor and there are no clear channels for reporting issues
3. Due to the huge amount of users, Greenshot is rate limited which in it's turn makes our users report issues to us

In the end, we wanted to help our users by supplying access to the service but we are spending a lot of time tracking the problems. This time goes away from the time we need to develop, and also doesn't bring us anything.

So we decided to only include new services when we can make sure these are not using up our time.

It seems Qiniu is very chinese, which we don't speak... I have a hard time figuring out the terms and conditions of the service....

harlanc:
The missing libraries are added.

Lakritzator:
Hi Harlan,

my concern is not the SDK, but the Qiniu server... Currently we have > 100.000.000 users, as we cannot know in advance how many might use Qiniu this can take the whole server down. Qiniu might block Greenshot, which will generate a lot of work for me and my colleagues as we need to answer the problem tickets which are reported.

So we need to know the usage policy of Qiniu, if this is rate limited.
We also need to know what kind of support Qiniu can give us, when our users can't upload.
We make no money with the service, so it needs to be free too...

About the libraries, we don't like dlls directly inside our repository. Don't you use nuget?

Best wishes,
Robin

PR #79, opened 2018-02-23
Removed redundant check for lineThickness > 0

toddryan:
There was an additional check for lineThickness > 0 before checking if shadow is true and drawing. We already know that lineThickess is greater than zero due to the check at line 57


toddryan:
Over 2 years for a response.
My contribution likely does not apply anymore.

PR #82, opened 2018-03-18
OneDrive integration plugin

Vistritium:
Hello,

I have created plugin that supports uploading to OneDrive.
I saw https://greenshot.atlassian.net/browse/SUPPORT-78 apparently it was not completed so I decided to write one myself.
I have tested it on my private and company account.
When building the plugin you need to supply clientId in `OneDriveCredentials`. You can create app on https://apps.dev.microsoft.com
I based this plugin on Dropbox plugin.


Lakritzator:
I will look at this later, I worked on develop (which should become Greenshot 1.3) offline for some time now, the conflicts are not so dramatically...

Vistritium:
Ok. I can try to integrate those changes into develop, just let me know if this would be useful. Also I can help with Dropbox plugin, also let me know.

Lakritzator:
Have a look at my changes... there is not much left of the old code...
I think the best example is Dapplo.Addons.Imgur

I am using a new HTTPclient, Dapplo.HttpExtensions, which I made while flurl was not doing stuff correctly (maybe it is now!). This should actually make it easier to fix the Dropbox bug, so I will first check that myself.

I am using MVVM (CaliburnMicro) for the UI, which is a must for the configuration of every addon!
It should be very simple to copy / paste from a different addon.
The configuration framework itself is slightly simplified, and also the language support is improved.

You can use MEF (Inversion of Control or Dependency Injection, whatever you want to call it) to get your Information injected, which makes things less "Spaghetti"-like, more modern. You can just add attributes, and don't have to work with stuff in the AssemblyInfo.cs etc... like before.

I still have left out a lot of changes, I actually went a bit to far with changing Greenshot offline, while they would not make it possible to release Greenshot this year 😞 
There still are a lot of loose ends, which will be clear as I will keep pushing a lot of changes.
But due to the number of pull-requests I would be missing out on, I needed to move the changes back online.

The steps that are needed:
* Create IOneDriveLanguage.cs
* Create IOneDriveConfiguration.cs
* Create ViewModel / View
* Remove OneDrivePlugin.cs, move the upload to the destination
* Modify OneDriveDestination to have
* Modify the OneDrive utility
-> You can use Json.Net via the HttpExtensions, currently Greenshot doesn't bring this itself so you need to use Costura (see Dapplo.Addons.Imgur!)

I will be more than happy to do this all for you, it would be enough if you make sure your code is rebased or pasted into the current develop branch, even if it doesn't compile... I just don't want to copy it myself, because this would remove you from the contribution list.

P.S.
It still is a bit sluggish, I need to optimize the startup a bit...

Best wishes,
Robin

Vistritium:
I have pasted and updated the plugin code. Currently I managed to make project to compile but I'm facing some runtime errors on startup. Probably something is still wrong with onedrive's plugin code because when I remove this plugin app starts.

More info about error: https://gist.github.com/Vistritium/57ceb533b274484d45d98a3988e54b32
But I didn't spend too much time trying to debug it yet, maybe something trivial.

I'm happy to continue working on this plugin. However if you want to take over then it is no problem for me as well.

Lakritzator:
Seems to be a composition issue, where something is not exported correctly...
I will probably get it working 😄 
And I will let you know, you are very much welcome to continue work, but let me get it working first.

Lakritzator:
Greenshot is starting again, but I didn't test the destination yet.
It is late, and I need to sleep!

P.S. Credential files are no longer needed, you can add a file called "Greenshot-constants.ini" to the same location as the .exe or drop it into %APPDATA%\Greenshot

The contents are:
```
[OneDrive]
ClientId=....
ClientSecret=....
```

Vistritium:
The destination probably is not working yet.
Thanks for info regarding credentials, I'll clear that up.

Lakritzator:
I already removed the Credentials class yesterday, that is why I had to inform you of the constants file :smile:

I already had a go with the destination, and there seems to be an issue with the Authentication, which might be on MY side... I believe it's the same issue in the Imgur plugin when using Oauth.

I will keep you informed...

Vistritium:
I'm going to look today/tomorrow into OneDrive destination.

PR #84, opened 2018-03-30
Fix typo for DIB in language files

smileytechguy:
Correct "Device Independend Bitmap" to "Device Independent Bitmap"


Lakritzator:
Thanks!!!

PR #85, opened 2018-04-07
OneDrive addon - make destination work

Vistritium:
Hi, some changes to make Destination work. The plugin is not complete yet but destination should work.

1) Changed authorization endpoints back to login.microsoftonline.com. Many live endpoints are deprecated (idk if login too) and they probably don't work with work and school microsoft accounts. More info:
 - https://docs.microsoft.com/en-us/onedrive/developer/rest-api/concepts/migrating-from-live-sdk
 - https://docs.microsoft.com/en-us/onedrive/developer/rest-api/getting-started/authentication

2) Removed ClientSecret from configuration. The native applications need not to use secret id when contacting Microsoft oauth authorization api. More info:
 - https://docs.microsoft.com/en-us/onedrive/developer/rest-api/getting-started/app-registration

3) Make ClientId writable. Otheriwse it gets wiped out on every save.
4) Fix upload endpoint - it needs a ":" at the end.....(spent so much time debugging this!!!) :P

TODO list:
1) Fix configuration UI
2) Currently links are private, add option to select anonymous or organization links:
 - Add option in UI
 - Additional calls needed for another type of link


Lakritzator:
Cool, thanks for the work!!!

That with the colon : is described here: https://docs.microsoft.com/en-us/onedrive/developer/rest-api/concepts/addressing-driveitems

What do you mean with wiped out of the client-id? Is there a bug in my configuration frameworl?

Will merge it right away! 😄

Vistritium:
> What do you mean with wiped out of the client-id?
I'm not sure how is this supposed to work, but when I add clientId to `%APPDATA%\Greenshot` it is read correctly, but when app saves the addon's state (for example after retrieving token) it removes the clientId (if this is marked as read only) and on next app start it get's default value.

Vistritium:
Ok, I think I know what I did wrong. I should have saved it to `Greenshot-constants.ini` not to `Greenshot.ini`

Lakritzator:
Yes, use that file, so you always use your own values.

But with that information I got what was wrong... I should only prevent writing the default value, but allow reading & writing. Fixed with a commit I just pushed.

Thanks!!

Lakritzator:
I had some time to try it out, but I still get an error:
Http response 400 (BadRequest) for https://login.microsoftonline.com/common/oauth2/v2.0/token, details from website:
{"error":"unauthorized_client","error_description":"AADSTS70001: Application with identifier 'xxxxxxx' was not found in the directory xxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxxxx\r\nTrace ID: 83b8c1d0-7991-4f26-97a9-0e8b32480900\r\nCorrelation ID: f0fede14-eabe-4586-b2f5-410354bdf23a\r\nTimestamp: 2018-04-08 10:48:01Z","error_codes":[70001],"timestamp":"2018-04-08 10:48:01Z","trace_id":"83b8c1d0-7991-4f26-97a9-0e8b32480900","correlation_id":"f0fede14-eabe-4586-b2f5-410354bdf23a"}

Any ideas? How did you create the client-id?

Vistritium:
I created an app on
https://apps.dev.microsoft.com/
Make sure that there is "Native Application" platform added and `Files.ReadWrite` Microsoft Graph Permissions added:
greenshot-microsoft-app

Lakritzator:
Ah yes, got it working! Thanks!!! :)
That was the reason why I didn't get it working with the Graph API.

I want to get rid of the embedded browser, it's awful! (It's Internet Explorer, there is currently no other way)

As we want to have a consistant way of working with OAuth authentication for all services, I want to try to open in the default browser and have the service (OneDrive in this case) redirect via the Greenshot website. The code will be passed via the window title (or another way), this is called OOB. This only works reliable when I have the script to call Greenshot under my control, that is why I want to redirect to the Greenshot site.

The other thing is, ofcourse, we can get a little traffic... and reduce our costs a bit.
Also we can inform people when they have an old version, or that there are service issues, or for instance that the service will no longer be available. Our website is static, there is no database or other logging, and it's very high available as it's cached around the world (Cloudflare).

Anyway just so you know, I will try getting the OOB working for all services, OneDrive will be a part of this.

Vistritium:
Ok, thanks for letting me know. 
But frankly, this sounds like weird idea :P (out of curiosity) did you see such solution somewhere? I usually see embedded browsers being used like this one (for native apps).

> and have the service (OneDrive in this case) redirect via the Greenshot website
Regarding this, I'm not sure if this is bug or I'm doing something wrong but currently I cannot set some custom redirect urls in Microsoft Apps although it probably should be possible. For example cannot set `https://getgreenshot.org`:
redirect_url_cannot_set
Weird..

Lakritzator:
OOB is used, but not often. The problem is that most people don't know how a browser can communicate with a local application. You can for instance use a local "HTTP" server, but this is tricky as you can't always use the same port to communicate. Another trick is to track title changes, so a browser script can just change the title to contain the information! (in this case a the title should contain the GUID for the authorization so Greenshot can find the info.)

This is easy to implement, with a library I wrote...

Now, about the redirect URI, that is unfortunate...
I had a look, what they want to probably mean with the {scheme}://{domain}/{path} is probably custom uri-schemes.. An application can register it's own scheme, which calls the application.  (this is a third way for a browser to call an application)

An example would be the uri: greenshot://oauth/onedrive
Which will call greenshot, if registered properly, with the URI as a commandline argument.

I need to check if there is a way to redirect to Greenshot... if not, we can use the URI "urn:ietf:wg:oauth:2.0:oob", which will not bring the initially wanted functionality of the redirect, but I can still open the browser with a Greenshot Uri, which will call OneDrive 😁 
The "OOB" should bring the functionality I mentioned elsewhere.

PR #86, opened 2018-04-08
Onedrive addon improvements - UI, public&organization link type

Vistritium:
OneDrive improvements:

UI:
* Removed unused language properties (maybe they are needed but I just didn't know?)
* Added OneDriveLinkType selection
* Added reset credentials button

Destination:
* Support public and organization links - additional call was needed to retrieve them

### DIFF
--- a/src/Greenshot.Addon.OneDrive/Entities/OneDriveGetLinkRequest.cs
+++ b/src/Greenshot.Addon.OneDrive/Entities/OneDriveGetLinkRequest.cs
@@ -1,0 +1,36 @@
+﻿#region Greenshot GNU General Public License
+
+// Greenshot - a free and open source screenshot tool
+// Copyright (C) 2007-2018 Thomas Braun, Jens Klingen, Robin Krom
+// 
+// For more information see: http://getgreenshot.org/
+// The Greenshot project is hosted on GitHub https://github.com/greenshot/greenshot
+// 
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 1 of the License, or
+// (at your option) any later version.
+// 
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+// 
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+#endregion
+
+using Newtonsoft.Json;
+
+namespace Greenshot.Addon.OneDrive.Entities
+{
+    public class OneDriveGetLinkRequest
+    {
+        [JsonProperty("type")] 
+        public string Type;
+        
+        [JsonProperty("scope")] 
+        public string Scope;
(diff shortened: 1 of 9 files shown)


Vistritium:
If the user is logged into personal microsoft account and he celect the organization link type, the endpoint will return error:
```
{
    "error": {
        "code": "invalidRequest",
        "message": "Organization-scoped links are not available in OneDrive Personal.",
        "innerError": {
            "request-id": "b7d0b243-9eee-4c9a-bf5e-da05930b2839",
            "date": "2018-04-08T10:17:31"
        }
    }
}
```
Currently the app doesn't propagate the error message to the user. Maybe we can do it somehow? I wonder how to do it in clean way.

Lakritzator:
Cool!
I'll have a look at the error handling, what I can do to propegate the error.

If you have some ideas to improve the code which "you" as a destination developer has to implement, let me know. I hope that the current state ist better than that of 1.2.x?

Vistritium:
> I hope that the current state ist better than that of 1.2.x?
Definitely! :) less code and better abstractions (for example for oauth2) and also real json support are cool!
> If you have some ideas
Hmm, since you asked. I'm wondering if properties like `UploadFormat`, `UploadJpegQuality` or `FilenamePattern` should be domain of the addons. Won't users want to set this value once and expect it to be consistent across addons?

Other topic:
Can you take a look at the https://github.com/greenshot/greenshot/blob/9490ebd270a7e4f2b897e4a9a3ef1f3fc8873ee3/src/Greenshot.Addon.OneDrive/ViewModels/OneDriveConfigViewModel.cs#L115
Button should get disabled after click but I'm not sure how to achieve it.

Lakritzator:
Just to inform you, and have the information somewhere where I can find it at a later time...

I have added some of the basics of my idea to make it possible for a destination to have it's own output settings.

I have added new interfaces which a destination can use to extend it's configuration, and these interface contains all the output information. (Format, colors, quality etc etc)
There is one (IFileConfiguration) for the information and one (IDestinationFileConfiguration) which adds a boolean if the global settings  are used or not.

Currently I have used the new way in the Imgur and OneDrive destinations, they can use extension methods I wrote for ISurface, to get different information or output to a stream. The "what" needs to be exported is decided by supplied information like this:
```
surfaceToUpload.WriteToStream(imageStream, CoreConfiguration, _oneDriveConfiguration);
```
If the later is not supplied, the WriteToStream uses the global IFileConfiguration, if the later is supplied and UseOwnSettings is set to true, the second IFileConfiguration information is used.

This opens up the possibility to have separate effects, formats etc, for destinations, and have a consistant behaviour of outputting bitmaps.

Currently I didn't implement all the output settings in the UI **yet**, but the destination doesn't have to do anything to get all of them as soon as I extend the UI... The FileConfigPartViewModel / FileConfigPartView (name might be changed) takes care of this.

This goes in the ConfigView of the destination
```
<ContentControl x:Name="FileConfigPartViewModel"/>
```

This goes in the ConfigViewModel of the destination, e.g. your OneDriveConfigViewModel:
```
        [Import]
        public FileConfigPartViewModel FileConfigPartViewModel { get; private set; }

        public override void Initialize(IConfig config)
        {
            FileConfigPartViewModel.DestinationFileConfiguration = OneDriveConfiguration;
```

Vistritium:
Cool!

PR #89, opened 2018-07-05
Fix Traditional Chinese(Taiwan) translation

Lin-Buo-Ren:
The previous translation translates "hosted by" wrong, this patch fixes it.

Signed-off-by: 林博仁(Buo-ren, Lin) <Buo.Ren.Lin@gmail.com>


Lakritzator:
Thanks!!!

PR #102, opened 2018-11-30
Add [img] BBCode tag to the link on the clipboard

PhocaCz:
Hi,

will be great to add new parameter to Imgur plugin - Add [img] BBCode tag to the link on the clipboard.

The following PR includes changes for adding this new parameter, but you should know:

1) I am not C# developer, for me it is impossible to test the code (I really don't know if the code is right)
2) It's just a suggestion, of course it can be improved for more tags - not only IMG BBCODE tag.

Thank you,
Jan


Lakritzator:
Hey thanks!

It actually looks quite complete, from a code standpoint.
Properties should start with a capital, there are coding conventions for this, but not being a C# developer, I understand you missed that :smile: and I am actually quite surprised about the quality.

The only thing that is missing for me, is the "text" in the translation file, the checkbox will not have any text with it :grin: 

I need to think about the general idea behind this, and if this really adds value.
Can you elaborate on why you need or why you wanted to have this?

PhocaCz:
Hi,
thank you for taking a look at it.

Mostly, if you upload an image or screenshot to imgur, you do it, because you want to paste it somewhere - in some forum page or e.g. on Github page with help of some specific tags:

Imgur

So, you just upload an image to imgur and then you paste BBCode tag (or another code like Markdown) of this image to the text.

Real example:  In phpBB forum pages, you have an option to answer posts quickly (Quick Reply) and there are not available icons of BBCode tags , so pasting image link including the BBCode tags can be very helpful.

phpBB,

PhocaCz:
Hi,
is there a possibility to release source code of version 1.2.10.6

On Github, there is a version 1.2.10 which does not include the patch for 

**BUG-2235 | Imgur authentication issues due to imgur api change**

It seems like this release:
https://github.com/greenshot/greenshot/releases/tag/Greenshot-RELEASE-1.2.10.6
still includes the version 1.2.10 without the patch in the source code ZIP file.

I have tried to build the greenshot from this release but get the:
"An error occured while uploading to Imgur (403)" when uploading to imgur.

Built:
Built

Stable:
Stable

Thank you. 

Jan

Lakritzator:
Hi Jan
the code is in there!
You will NOT get 1.2.10.6 in your version just by building, as the assembly version (that which you see) is set on the build server during the build. This is specify here: https://github.com/greenshot/greenshot/blob/release/1.2.10/appveyor12.yml

You will also not be able to use any of the cloud services when you build yourself, as the credentials for this services are also included during the build. We can't really put these secrets into the git repository, can we? 😄 

So you will need to register with Imgur and specify you own api key in this file https://github.com/greenshot/greenshot/blob/release/1.2.10/GreenshotImgurPlugin/ImgurCredentials.cs if you want to use it.

Hope this helps.

Best wishes,
Robin

PhocaCz:
Hi, :smile:
great, thank you very much for the info.

Jan

PR #131, opened 2019-06-04
Removed external test dependency

wookiefriseur:
Removed capture test being dependent on some Notepad window. Test was only successfull if this window existed (had to open it manually).

Code from [your dapplo repo](https://github.com/dapplo/Dapplo.Windows/blob/450fae4b4a139b2bf6063bbbc6c40772355d4c0d/src/Dapplo.Windows.Tests/WindowsEnumeratorTests.cs#L85)

PS: Great seeing you working on Greenshot 2.0. Keep up the good work. Thanks to your recent changes, I was finally able to finish a build. With this small change all tests are successfull now.

### DIFF
--- a/src/Greenshot.Tests/CaptureTests.cs
+++ b/src/Greenshot.Tests/CaptureTests.cs
@@ -20,6 +20,7 @@
 using System.Drawing.Imaging;
 using System.IO;
 using System.Linq;
+using System.Reactive.Linq;
 using System.Threading.Tasks;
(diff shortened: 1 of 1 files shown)


Lakritzator:
Thanks, I'm still working on the dotnet core 3.0 migration, which is taking its time due to it being bleeding edge... so I didn't fix all the tests yet... But being able to test captures etc, does make it a bit easier to change stuff without worrying. 

Thanks for the change!!

PR #142, opened 2019-09-25
Update README.md

dikadali:
Added MahApps.Metro hyperlink and removed duplicate section : "For users the major changes since 1.2.x are:"


Lakritzator:
Sometimes one should read what they write, thanks for fixing this!

PR #150, opened 2019-10-25
Fix msg for FormatException

Christian-Schulz:
Small correction.
Argument with index 1 does not exists in argument list.


Lakritzator:
Good catch, thanks

PR #157, opened 2019-11-29
Fixing textbox issue with polish letter ą

marcin1990:
Fixing issue in textbox when typing ALTGR+A (polish letter ą) raised in threads BUG-2288, BUG-2474.


Lakritzator:
Oh, very cool!!! Thanks a lot, will merge this right away.

PR #190, opened 2020-04-17
Font size field type fix

KillyMXI:
This looks like a copy-paste error. Font size is supposed to be a float, judging from the rest of the code. Wrong type here causes exceptions.

Side note: both DecimalIntConverter and DecimalFloatConverter also look like someone wasn't paying attention. `Decimal --> (explicit) Int16 --> (implicit) Float` conversion causes no issues only due to the fact all numbers effectively are small integers.


Lakritzator:
Wow... how the hell did you find those bugs 👍 
I looked at the code, you are so right... also with the side note.

DecimalFloatConverter should use Convert.ToSingle(o);
DecimalIntConverter  should use Convert.ToInt32(o);

You are more than happy to change those in the same PR, if you like.
As you found them, but I have no problem doing the changes myself...

PR #200, opened 2020-04-26
Propagate DPI Changes down to Drawable Containers and Adorners and Resize Grippers Accordingly

jklingen:


Lakritzator:
Looks ok

Greenshot/Drawing/Adorners/MoveAdorner.cs
```diff
@@ -147,12 +147,13 @@ public override void Paint(PaintEventArgs paintEventArgs)
 
 			targetGraphics.SmoothingMode = SmoothingMode.None;
 			targetGraphics.CompositingMode = CompositingMode.SourceCopy;
-			targetGraphics.PixelOffsetMode = PixelOffsetMode.Half;
+			targetGraphics.PixelOffsetMode = PixelOffsetMode.None;
```
jklingen: @Lakritzator is there a special reason this needed/needs to be set to `Half`? :thinking:  it caused a weird drawing bugs when I added the 1px border (repaint artefacts + a single missing pixel). Got those fixed by changing to `None`, I don't think it broke anything else.

Lakritzator: I'm pretty sure this was the standard, as Half normally improves the quality.
See about it here: https://photosauce.net/blog/post/image-scaling-with-gdi-part-3-drawimage-and-the-settings-that-affect-it

Lakritzator:
Looks ok, but let's not merge before the Zoom is, just in case we have a merge conflict.

What I just remembered, we might need to change the obfuscation pixel size too depending on the DPI. But not for the screen DPI, instead for the DPI of the capture! Otherwise it will still be readable 😁

Lakritzator:
Jens, can you update this so we can merge it?

Greenshot/Drawing/Adorners/MoveAdorner.cs
jklingen: > None=Bad, Half=Good

Hmm, maybe I am missing something, but I am not sure whether this article really applies to our scenario here. The blog post is all about scaling images, and so are most other examples I found dealing with `PixelOffsetMode.Half`. Actually, as we are drawing geomatrical shapes using integer coordinates, I'd rather be surprised if we really had to think about pixel offset, smoothing and interpolation at all... :thinking: 

I noticed that `TargetAdorner` had none of them set, so I tried to discard all three from `ResizeAdorner` and `MoveAdorner` too. The result looks as I'd expect it to look. So maybe we don't need them in this scenario at all?

jklingen:
Sorry, took some time before I found the time :frowning_face: 

> Looks ok, but let's not merge before the Zoom is, just in case we have a merge conflict.

Just resolved all merge conflicts in here.
 
> What I just remembered, we might need to change the obfuscation pixel size too depending on the DPI.

Well, the user needs to visually check and maybe re-adjust the obfuscation result anyway, so I don't think this is necessary. The same problem occurs for large font sizes - the default obfuscation settings won't be helpful if you are trying to obfuscate something in a headline :grin: 


I think this PR is ready to merge if you're okay with [what I did to the `Adorners`](https://github.com/greenshot/greenshot/pull/200#discussion_r449689029).

Lakritzator:
Oh, somehow the only case I looked at works but I have a lot of still small adorners and some drawing artifacts... was merged to premature due to lazyness from my side to check... oops.

Lakritzator:
Fixed it...

jklingen:
@Lakritzator  thanks for fixing this (and making the code more beautiful :smiley: )

just out of interest - how did you reproduce the drawing artifacts? I couldn't find any

Lakritzator:
@jklingen I found the artifacts as I actually test with a DPI of 200%, due to the high resolution of the 12" screen of my Surface Pro.
I noticed that the size of the grippers are becoming much bigger than the safety margin that I used to redraw the screen.

There are other issues with the DPI support! I'm fixing them via https://github.com/greenshot/greenshot/issues/227

There is DPI support build into the .NET Framework which came with 4.7.2 if I recall, this is used to propage the DPI changes via the Windows Forms code. This only works if it's activated in the app.conf file, I should have kept my own implementation as it's a bit picky, but I still prefer to have the supported way. I have found that there are many issues, with the sizes of the icons/fonts etc, if you use different displays which don't have the same DPI, and trying to make it work.

Issue #185, opened 2020-04-14
Plans/timeline for the editor?


KillyMXI:
I'm having trouble to get an answer in the place of origin, so I have open this issue in hope to finally have a bit of maintainers attention!

There is an interest to have a certain feature in the editor. I would like to step in and implement it.

But, because the existing editor is marked as legacy and there is an ongoing work to move to WPF, I'm not sure it makes sense to work with the existing editor. I also see a couple of experimental projects that don't seem to be any close to be shipped, but if there is something going on under the radar - that might also be helpful to know.

So, I would like to know your plans and estimates for the editor.

----

https://greenshot.atlassian.net/browse/FEATURE-84 - no answer for 3 weeks. You better do away with Jira if you provide no support there. So far it looks like writing to /dev/null.

https://github.com/greenshot/greenshot/issues/25#issuecomment-603229039 - no answer either.

Lakritzator:
I'm sorry you can't wait... but honestly, I'm ill and for the last 4 weeks we have to entertain a 4yo while the kindergarten is closed. Work didn't stop either, and groceries etc take much more time. In fact my son was banging on the door, while I started to answer this,  as he can't sleep... (8:40 in the evening) So I'm now sitting in his room next to his head typing this before I go to bed...

> You better do away with Jira if you provide no support there. So far it looks like writing to /dev/null.
You better stop nagging, Greenshot is free so why do you assume you got the right for support or a certain response time? Greenshot is a free time project, and I work on it when I have time and energy, and as time is short I rather do something fun or small. If you like things to go faster, sponsor us e.g. on open-collective or me here at github, this can make me work less and have more time for Greenshot.

Also do you really think that writing all over the place makes me come out of my bed, or interrupt my playtime with my kid, or that I answer during the work I also have to do to make a living? I see all comments and issues, in Jira and Github come by, sometimes I can't answer all.

Don't get me wrong, I really like to have some help to get the zoom working.
I saw your JIRA, comment and issue... answering takes a lot more time than a simple bug.
My priority is to get the 1.3 release out of the door, which is already costing me all my leftover energy.

As Greenshot 1.3 might still be used many years by people not liking 2.0 due to reasons yet unknown (dotnet core, size, looks), it depends on the effort one needs to put into a zoom, but it might make sense. As the editor is a lot of manual painting, I'm not really sure if it would be so easy. But you are welcome to propose a solution!

[...]

Issue #25, opened 2016-05-11
Zoom while editing is needed


Layarion:
Using the beta

I'm trying to Draw a rectangle around a very small letter in the cleanest, aligned, and sharpest looking way possible...

but i am having a hell of a time because the 8 black re-sizing blocks make it impossible to see what i'm doing. please for the love of god, allow me to zoom in OR give me a magnifier just like when i'm taking screenshots.

Lakritzator:
https://github.com/greenshot/greenshot/blob/develop/CONTRIBUTING.md

Please monitor the ticket here: https://greenshot.atlassian.net/browse/FEATURE-84
(yeah it's old... but we want to rewrite the editor and than it's in it)

gardian12:
Hello together,
I like to renew the request as described in [https://greenshot.atlassian.net/browse/FEATURE-84](url).
This would be very helpful in many cases.
Is there any chance to prioritize this?

Lakritzator:
@gardian12 Thank you for your interest in Greenshot.

Unfortunately it's not helpful for us to have people reporting or "pinging" for issues / status and priority in multiple places. The issues here at GitHub are for developer / technical code issues.

You can if you think it helps, add your comment to FEATURE-84 like "what is the status"... or "I find this also important".

But:
We still have a huge backlog transforming our application to modern technologies which will be easier to maintain or make it easier for others from the community to support us.

I work on Greenshot in my free time, which is defined by the time I don't work and not spent with my family, limited by the amount of chores I need to do for my house, garden, car etc.

Freeing up time, to speed up the development, is only possible if I work less or have to do less chores -> this is only possible if Greenshot makes more money. As Greenshot is free, and a lot of companies don't donate, we only get a limited amount -> we can only slowly go forward.

We are looking at solving this, one way would be: https://opencollective.com/greenshot
These ways to help us free time, are available under the "Sponsor" button at GitHub. Currently it's not possible to sponsor a single person as I'm still on the waiting list.

There is not really a way to prioritise this, unless someone from the community says "here is a pull-request" for this feature. I really need to get rid of technical debts, meaning we use technologies which are no longer supported or block us to go forward. So I first want to have a running & stable Greenshot, with dotnet core 3.0, before I will add or change features. Dotnet core 3.0 is still in preview and will not be available before September, meaning there most likely won't be new version of Greenshot before that time. To be clear, there is also no guarantee that I will be able to finish before that time. "Greenshot is as is, and comes when it comes."

[...]

PR #201, opened 2020-04-26
Zoom for the editor (Feature-84)

KillyMXI:
Zoom implementation

- https://greenshot.atlassian.net/browse/FEATURE-84
- #25 
- #185 

image

Notes:

- Shortcuts Ctrl+ `+`,`-` were used for [another feature](https://greenshot.atlassian.net/projects/FEATURE/issues/FEATURE-919). Considering these shortcuts are wastly more common for the Zoom feature, I suggest to move the previously existing feature to Ctrl+ Shift + `+`,`-`

- ~~Shortcuts for Actual size (100%) and Best Fit (Ctrl+`/` and Ctrl+`*`) are chosen for the proximity on the keyboard and because of possible collisions for other choices. Of the software I use, Ctrl+`/` for actual size is also used in XnView MP.~~

- "Zoom" menu instead of "View" menu - because there are no other View features so far, I think it's more accessible this way.

- Zoom display on the bottom right shows the same menu on click. Menu lists all possible values for quick access. I think this UI solution makes a good balance between usability and code complexity.

Update:

**Summary of behavior changes:**

- New Zoom submenu in the main menu;
- Zoom value display in the status bar, opens the same Zoom submenu on click;
- New shortcuts <kbd>Ctrl</kbd>+<kbd>+</kbd> and <kbd>Ctrl</kbd>+<kbd>-</kbd> for Zoon In and Out, <kbd>Ctrl</kbd>+<kbd>0</kbd> for actual size, <kbd>Ctrl</kbd>+<kbd>9</kbd> for Best Fit (find a zoom value where no scroll is needed to see the entire image);
- Feature previously existing on the same shortcuts is now moved to <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>+</kbd> and <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>-</kbd> - add empty borders to image, trim borders;
- Clipboard paste (text and images) is now aligned based on currently visible area of image;
- Editor window, when opened, is sized to fit on screen;


jklingen:
Hey, thanks for the PR :+1:  I'll do some testing.

I agree that we should re-assign <kbd>Ctrl</kbd> + <kbd>+</kbd> and <kbd>-</kbd>, the other feature is not widely known anyway.

Personally I'd prefer <kbd>Ctrl</kbd> + <kbd>0</kbd> for 100%, I think that's what most popular browsers do and should known for most users.

KillyMXI:
Yeah, I've got another response about <kbd>Ctrl</kbd>+<kbd>0</kbd> already.
Personally, I would prefer to avoid using digits for zoom shortcuts.
When I was considering <kbd>Ctrl</kbd>+<kbd>0</kbd>, I figured I don't like it when icon says `1` inside a magnifier.
Changing the icon to `=` inside a magnifier (which is also available in Fugue set) will make it possible to confuse with <kbd>=+</kbd> key on keyboard.
And then, there might be more natural uses for shortcuts with digits.

jklingen:
Thanks again, so I did some testing, the implementation works great in most situations - good job :bow: I have just found some smaller issues, and a few points open for discussion (this includes the shortcuts already mentioned above). Adding these as tasks here (we could also consider to make separate tickets for these, in case any of them takes longer or needs a lot of discussion):

**Issues:**

- [ ]  Dual monitor setup: on secondary screen, zoom to best fit zooms to 0%; after moving window back to primary screen zoom to best fit always zooms to 100%

- [ ]  Zoom factor < 100% with text box on screen: selecting the textbox seems to redraw everything on top and left of it, making text look pixelated (also happens when dragging tail of a speechbubble) - probably not an issue with your implementation, but we should have a look at it.

- [ ]  Zoom 50% with speech bubble: tail can only be dragged on the upper left quadrant of the screenshot - with Zoom set to 75%, the draggable area is the the top / left 75% of the screenshot


**Open for discussion:**

- [ ]  Should the window be resized with the screenshot? I somehow expected it to remain at the same size, similar to a browser or Paint. Especially the window should not grow out of the visual area, because then the zoom controls in the bottom right are suddenly gone (yeah, there are still other ways, but it is confusing).

- [ ]  Likewise: should "zoom to best fit" zoom window to fit screen, or screenshot to fit window? Personally, I expected the latter.

- [ ]  Maybe add Ctrl + Mousewheel support, too?

- [ ]  Shortcut Actual Size - Change to <kbd>Ctrl</kbd> + <kbd>0</kbd>? Note: on German keyboard, <kbd>Ctrl</kbd> + <kbd>/</kbd> and <kbd>Ctrl</kbd> + <kbd>*</kbd> only work on keypad (because otherwise both <kbd>/</kbd> and <kbd>*</kbd> require holding down <kbd>Shift</kbd>.

- [ ]  If above shortcut is changed: Shortcut Best Fit could be Ctrl + 9 (next to 0 and numerically, 9 is "best fit" for a single character, sort of ;-) )

mischlrebl:
@KillyMXI  Great PR!

Ctrl + 0 would make sense, but I agree that digit shortcuts sould be reserved for other features: line thickness, colors, ...

What about the spacebar? Nowhere used so far and I do not see a use case

jklingen:
@Lakritzator would you like to have a look at the code? You're the one with the deepest insight into the current codebase.

KillyMXI:
> Dual monitor setup

I have to figure out something to test this. I was hoping I handled this well. But this sounds like .Net might not return what I expect it to.

> Zoom factor < 100% with text box on screen: selecting the textbox seems to redraw everything on top and left of it, making text look pixelated (also happens when dragging tail of a speechbubble)

Can't reproduce this. Maybe sensitive to DPI settings? Have to check on a different device.

>  Zoom 50% with speech bubble: tail can only be dragged on the upper left quadrant of the screenshot

How I missed this... I'll look into it.

> Should the window be resized with the screenshot? I somehow expected it to remain at the same size, similar to a browser or Paint.

I would prefer it to take least amount of screen space. The main goal of Greenshot is capturing what's going on in other apps, after all.

> Especially the window should not grow out of the visual area

And this is a good point. I'll look into it.

> Likewise: should "zoom to best fit" zoom window to fit screen, or screenshot to fit window?

Latter is kinda pointless for me. As I mentioned above, I don't see a point of Greenshot editor taking more space than it needs to show the screenshot.
My idea behind this feature:
Screenshot first opened in 100% zoom, showing it "As Is". And this shortcut allows to fit it completely on screen. There is another related question:

- [ ] Should the editor open with 100% zoom (exact capture, but might not fit on screen completely) or with Best Fit (full overview, but less details). If latter, I most likely need to make a provision for a stop value (of 100%), so it won't zoom in before requested explicitly.

> Maybe add Ctrl + Mousewheel support, too?

I'm not a fan of it. I'll wait for more opinions.

> Note: on German keyboard ...

Yeah, I'm kinda afraid of what might be happening on different layouts. Good point for <kbd>Ctrl</kbd>+<kbd>0</kbd>.

> Shortcut Best Fit could be <kbd>Ctrl</kbd>+<kbd>9</kbd>

Agree, it looks best from the proximity on the main keyboard. Although it hurts me from how otherwise senseless it is.

> ... digit shortcuts sould be reserved for other features: line thickness, colors, ...

Yeah, I had that in mind.
I was wondering how important it is to keep <kbd>0</kbd> with other digits. Sequential features usually start from 1...
At this point I don't think <kbd>9</kbd> will also be an issue. Human memory is only good for 3-5 things, anything above that will probably be unusable. I would only ensure there is a gap between sequential features at 1-... and specific features at the end or digits row.

Lakritzator:
I just checked out the PR, and am now working with it.
Didn't look at the code yet, just wanted to have an initial feeling about it.

It gives me a good vibe, and I'm **very** impressed so far!!!

Most of the issues have been stated by Jens already. About the zoom with the ctrl + mousewheel, I myself am a big fan of this!

I will now take a look at the code.

Lakritzator:
Ah the first issue I found, is that the range of the speech bubble tail is limited when the zoom < 100%

Lakritzator:
When I am at < 100% the crop doesn't crop the correct (selected) area.

KillyMXI:
Yeah. jklingen pointed at this already. My bad, will fix soon.

Lakritzator:
Sorry, I somehow missed that... :)
Please, by all means, take your time!!! People have been waiting for ages allready.

KillyMXI:
> When I am at < 100% the crop doesn't crop the correct (selected) area.

This is mysterious. I think I had it at some moment, but I can't reproduce it since.

Lakritzator:
I like best fit, but it doesn't work after I rotate the capture.

KillyMXI:
Hmm. Can't reproduce again. Might be same cause with that crop issue?

jklingen:
I can habe a deeper look at the first two issues I reported, the ones you cannot reproduce. Probably tomorrow.

Greenshot/Drawing/Adorners/AbstractAdorner.cs
```diff
@@ -116,6 +116,23 @@ public virtual Rectangle Bounds
 			}
 		}
 
+		/// <summary>
+		/// Return the bounds of the Adorner as displayed on the parent Surface
+		/// </summary>
+		protected virtual Rectangle SurfaceBounds
```
Lakritzator: I'm not quite happy with the naming here, as it would suggest that these are the bounds of the surface itself. Maybe BoundsOnSurface fits better??

Lakritzator:
I'll see how to reproduce the rotate issue!

Greenshot/Drawing/Adorners/AbstractAdorner.cs
KillyMXI: I agree

Greenshot/Drawing/CropContainer.cs
```diff
@@ -56,7 +56,15 @@ private void Init()
 		/// We need to override the DrawingBound, return a rectangle in the size of the image, to make sure this element is always draw
 		/// (we create a transparent brown over the complete picture)
 		/// </summary>
-		public override Rectangle DrawingBounds => new Rectangle(0,0,_parent?.Width??0, _parent?.Height ?? 0);
+		public override Rectangle DrawingBounds {
+			get {
+				if (_parent?.Image is Image image) {
+					return new Rectangle(0, 0, image.Width, image.Height);
+				} else {
+					return new Rectangle();
```
Lakritzator: I like Rectangle.Empty instead of constructing one.

Greenshot/Drawing/Surface.cs
```diff
@@ -1371,6 +1404,25 @@ public Image GetImageForExport()
 			return GetImage(RenderMode.EXPORT);
 		}
 
+		private Rectangle ZoomClipRectangle(Rectangle rc, double scale)
```
Lakritzator: You can make this, and the next method, static as they don't use any of the instance variables.

Lakritzator:
I finally managed to do a review.

First this: I'm amazed at what you did! You managed to understand the not quite simple Greenshot editor, and made a not trivial change to it. And on top of this, you managed to do this in such a way that it even looks like how we would do it.

Chapeau! Well done!!!

I was only able to do minor nit-picking, all my remarks have little value...

KillyMXI:
> You could say, it works as designed. But I'm not sure if this makes sense...

Anything between full hundreds (e.g. 150% or 133%) can't be shown "pixel-perfect" - it looks really distorted.
I think with screenshots it's more important to show accurate representation of the original rather than allowing to show any size but with smoothing.

I just checked what Paint.NET does. It enables smoothing for 150%.
I can make the same in principle, but I'm not sure it's what should be done in the screenshot editor.

Greenshot/Drawing/CropContainer.cs
KillyMXI: Ah ok, right

Greenshot/Drawing/Surface.cs
KillyMXI: I'm so used to VS suggestions on this I didn't pay much attention myself. And got caught when it didn't catch these methods for some reason

Lakritzator:
You brought up a good point (therefore the 👍 )

Right now you convinced me that it doesn't make sense, but I will talk about it with the guys and I have the right to change my mind. 😄

Lakritzator:
Just a side note... it will not be an easy merge into develop 😢 
But the changes are quite simple, I don't know why we didn't get to this yet.

KillyMXI:
In fact, I started this in the develop branch and switched to release/1.3 half way through 😅

Lakritzator:
That was the right choice, 2.0 is far from finished. Maybe partly due to some of the dotnet core 3.x short comings but mainly because there are still a lot of bugs and things to do.

KillyMXI:
Pushed a bunch of fixes. I will continue tomorrow.

Speech bubble must work properly now. Menus are also located properly.

Most of these issues come from the fact there is now a distinction between Surface size and Surface.Image size, and the code often uses the former instead of the latter.

- [ ] One of such places I identified in OneNoteExporter. That needs to be tested to make sure I got it right.

I think I know where this causes the crop issue. Will fix tomorrow.

Editor now tries to stay completely on screen with bigger zoom levels.

- [ ] But I suspect this also needs to be tested on a multi-monitor configuration. And there might be better logic to make use of available working area there.

Shortcuts are now <kbd>Ctrl</kbd>+<kbd>0</kbd> for Actual Size and <kbd>Ctrl</kbd>+<kbd>9</kbd> for Best Fit.

Lakritzator:
I'm not so enthusiastic about the needed surface.Width -> surface.Image.Width change.
But I just wanted to document it here, the reason for this change is due to the fact that the Width & Height are part of the Control.

We might consider to add new fields, which don't need "surface.Image".Width

I still found some references to surface.Width, which will probably cause issues. For instance the constructor for the FreehandContainer or DrawableContainer.AlignToParent.

Lakritzator:
I found with the current state, after your pushes, still an issue which was already seen by Jens:
If we move the Speechbubble Tail, parts of the capture are changing (different color / sharpness).

I expect there is an GraphicsState state = targetGraphics.Save() & targetGraphics.Restore(state) mismatch.

What we still need to check, is what happens with the zoom when we have something which needs scrolling.

KillyMXI:
Yep, because Surface is a Control and its dimensions are used externally, we can't keep using them for the Image size with the introduction of zoom. There are choices where to draw the boundary between coordinate spaces, but it has to be somewhere in any case.

> FreehandContainer 

Just pushed the fix

> DrawableContainer.AlignToParent

Didn't stage this change yet.
Currently, it is used only when we paste text from clipboard, as far as I can see.
 I started to wonder whether it's a good idea to put it in the middle of image (might be out of sight), or we should put it in the middle of the currently visible area (some work required).

KillyMXI:
Hmm. Image paste uses cursor location since [Feature-995](https://greenshot.atlassian.net/browse/FEATURE-995) but text paste don't.
I think making text paste consistent with this behavior will be a right thing to do. 

It's weird how AddTextContainer() is defined in terms of alignment rather than X,Y coordinates.

KillyMXI:
That was quite a digression, but I got the better logic for paste location.
There are some arbitrary numbers chosen to be not too different from previously existing behavior.
I also took the liberty to remove DrawableContainer.AlignToParent completely. It will be easier to add something new if ever needed than carry some dead but distracting code around.

Greenshot/Forms/ImageEditorForm.cs
```diff
@@ -1444,28 +1463,129 @@ private void TornEdgesToolStripMenuItemMouseUp(object sender, MouseEventArgs e)
...
+
+		/// <summary>
+		/// Compute a size that the form can take without getting out of working area of the screen.
+		/// </summary>
+		private Size GetAvailableScreenSpace() {
+			var screen = Screen.FromControl(this);
+			var screenBounds = screen.Bounds;
+			var workingArea = screen.WorkingArea;
+			if (Left > screenBounds.Left && Top > screenBounds.Top) {
+				return new Size(workingArea.Width - Left, workingArea.Height - Top);
```
jklingen:
```suggestion
				return new Size(workingArea.Right - Left, workingArea.Bottom - Top);
```
Fixes the [secondary monitor issue](https://github.com/greenshot/greenshot/pull/201#issuecomment-619582140).
For the primary monitor e.g. `Width` and `Right` are the same, but not for the secondary.

jklingen:
Thx for updating the hotkeys, I think `0` and `9` are quite handy :+1: 

With regards to 'Should the window be resized with the screenshot' - good point. I was expecting what I got used to from other software, but that doesn't mean that this is also the best in our case. Same goes for 'should "zoom to best fit" zoom window to fit screen' :heavy_check_mark: 

* [ ] Another question for discussion: should the window also resize to the top/left for "zoom to best fit", or would that be confusing? I often have Greenshot editors on the lower half of my screen, which means that "zoom to best fit" does nothing. What do you think :thinking:

jklingen:
> Editor now tries to stay completely on screen with bigger zoom levels. But I suspect this also needs to be tested on a multi-monitor configuration. 

Works fine on multi monitor setup :heavy_check_mark:

Greenshot/Forms/ImageEditorForm.cs
KillyMXI: Ah right. I messed up concrete screen and continuous coordinate space where it also lies itself. Thanks

KillyMXI:
> Another question for discussion: should the window also resize to the top/left for "zoom to best fit", or would that be confusing?

Tough question. Honestly, at this point I would like to avoid dumping a lot of time to make something clever and prefer to stop at something that does simple things well.

After Lakritzator's comment I started to think should Best Fit give bigger zoom and allow scroll to appear in certain conditions. Or should it be split into "Fit horizontally" and "Fit vertically"...

Maybe we should revise the thread and decide what are the blocking issues for this feature (I still have a few) and what can be left for later.

Lakritzator:
I still see some drawing artifacts when using < 100% and the freehand comes into play.
For my tests I drew a highlight, textbox and some lines, than I changed to 75%, when I started to draw with the freehand, weird things happened.

My best guess currently would be that it's due to not having a graphics.Save / restore in the draw.

I had a quick fix, but I'm not sure if it solves all issues.

In DrawableContainer in the method DrawContent there is a call to draw, place the store & restore around it.

```
GraphicsState state = graphics.Save();
Draw(graphics, renderMode);
graphics.Restore(state);
```

KillyMXI:
This can be reproduced with any element, line for example.
For me it currently looks like this:
For small enough objects something forces Nearest Neighbor rendering, no matter what is set before.
If you drag one adorner of a line - it will switch back and forth between smooth and pixelated look.
Transition happens at around 400 pixels distance (in image coordinates).

I tried to save and restore GraphicsState at different places with no success (I've also tried your suggestion).

So far I can't see what in Greenshot code can cause this. I'm starting to suspect it might be something with .Net.

KillyMXI:
I figured what was the issue. It is all about background under elements, so it is mostly in the SurfacePaint() method.
Parts of image were redrawn without giving an exact instruction how draw them, so apparently GDI+ has different fallback settings for different cases...

Now I need to tidy it up and make sure it works with filters too.

Lakritzator:
Great to hear!!! That must have been a pain to find, and to fix... I've had this experience way to often.
This is why we are hoping to finally get really started with a WPF based editor, where we can get different problems 😁

KillyMXI:
Rendering issues fixed, I think.

Now I realized that switching between zoom levels causes visible part of image to jump.
Working to insure center of visible area stays static when possible.

KillyMXI:
Now it feels pretty nice.

Anything still missing?

Greenshot/Drawing/DrawableContainerList.cs
```diff
@@ -523,9 +523,9 @@ public void PullElementsToTop(IDrawableContainerList elements)
 			}
 		}
 
-		public virtual void ShowContextMenu(MouseEventArgs e, ISurface surface)
+		public virtual void ShowContextMenu(MouseEventArgs e, ISurface iSurface)
 		{
-			if (!(surface is Surface))
+			if (!(iSurface is Surface surface))
```
Lakritzator, line 552: When this was written, we didn't have this language feature yet :)

Lakritzator:
Wow, that was quite a lot and this PR is most likely the biggest coming from an contributor yet!

I reviewed the code, but only online without VS, and it looks great.

Again I can only do nit-picking, and if I must I could say that I don't like the "magic" numbers being passed to the ZoomClipRectangle method.... If you wouldn't use the value inside, I would have suggested an extension method for the inflating.

I had a long day at work, and had to do a lot of work in my house, I will try running it tomorrow.
I lost track of all the things we noticed and currently I wouldn't know what could be missing.

jklingen:
I think I'll manage to do some re-testing later that day. But from what I already saw (plus what I read from you) I am pretty confident :) Again, great Job @KillyMXI , thanks a lot for your work on this.

I also lost track of all the todos and ideas floating around in this discussion, will have a look at them later, but I think the rest was rather nice-to-have ideas which don't need to be implemented now.

Note to ourselves @Lakritzator : maybe next time we should use issues to keep track of todos and ideas coming up in pull requests. The "Tasks" feature in Github doesn't seem to be nearly as powerful as  "Todos" in Gitlab.

Lakritzator:
I've done some quick testing, in general it seems to be working.
But I see some of the pixels jerk, as if parts are drawn with an pixel offset... weird.

jklingen:
I can see the same as @Lakritzator e.g. while moving a speech bubble tail, but personally, I could live with that.

One more question - I noticed that the font size in text boxes and speech bubbles is not scaled - is that on purpose? While it might be helpful in specific situations to have text at the original size (to make sure it is legible), it also can partially hide your text in other situations (see screenshot) and you cannot tell whether the edited text will fit into its container :thinking: 

not editing text
editing text

jklingen:
Apart from that, I tried to extract a list of items mentioned above which have not been cleared yet, I hope I didn't miss any. (Which does not mean that all of them need to be done now. Just to keep track and for us to see whether there are still show stoppers; we might want to keep some ideas for the backlog, or discard them altogether - let's see)

---

> Maybe add Ctrl + Mousewheel support, too?

I would have said "definitely yes" if we were zooming the screenshot within the window, but as we're zooming the complete window, I tend to "no" - this might be unexpected.

> Should the editor open with 100% zoom (exact capture, but might not fit on screen completely) or with Best Fit (full overview, but less details). If latter, I most likely need to make a provision for a stop value (of 100%), so it won't zoom in before requested explicitly.

I'd say, keep it as it is (for now) - idea for later.

> One of such places I identified in OneNoteExporter. That needs to be tested to make sure I got it right.

Anyone tested this yet?

> What we still need to check, is what happens with the zoom when we have something which needs scrolling.

@Lakritzator not sure if that's what you meant - but I actually found a new issue with a very very tall screenshot (scrolling capture of a newspaper website, several screen heights (abt 30.000px)) - "best fit" scaled to 0%, which then caused a crash ("System.OverflowException: "Arithmetic operation resulted in an overflow.").... I think we need to have min/max zoom values. And I think this is a show stopper, because scrolling capture is a commonly used feature.

> Another question for discussion: should the window also resize to the top/left for "zoom to best fit", or would that be confusing? I often have Greenshot editors on the lower half of my screen, which means that "zoom to best fit" does nothing.

Even though it feels a bit weird when "zoom to best fit" doesn't do anything even though there's space available above, I think this is something we could fix later.

---

Last but not least: there are some unresolved code review comments, seem to be minor changes - are we going to apply them?

KillyMXI:
> pixels jerk

It's due to rounding errors when top or left edge of a fragment changed. Something somewhere is computed based on clip rectangle rather that on whole image coordinates.
Extreme case for of this (or similar) issue can actually be observed with Pixelize filter when forcing it's partial refresh by moving something on top.
I don't know what to do about it without reworking the whole thing. It's a minor issue that appears to be really expensive to prevent. Unless there is some clever trick when scaling clip rectangles - I'll think about it.

> Font size in text boxes

I didn't touch it. Now thinking about it - it probably makes sense to adjust it. Will do.

> "best fit" scaled to 0%

Good catch. It should've stop at 25%. Will fix.

> there are some unresolved code review comments, seem to be minor changes - are we going to apply them?

I should've clicked the "Resolve" button. Applied in 79742d1

One more thing I keep wondering about:

Should Best Fit feature be split into "Fit width" and "Fit height"? Or should they be added as extra options?
Pros: For super tall images it will be handy.
Cons: For the main workflow it seems insignificant, and it increases the amount of menu items.

KillyMXI:
> The fix will be not pretty but quite simple.

I think I did it pretty but not quite simple in the end, lol.

KillyMXI:
I think I'm done.

Lakritzator:
🎉

I'll check out the funny jitter, which definitely was already there before.
 
Today was a German holiday and we planned to renovate a room for my kid, which I started in the evenings of this week and will finish tomorrow.

I'll check the code as soon as I have some time behind my pc. I believe it's awesome!!

jklingen:
Wow, you're quick :astonished: thanks for the update. I gave it another test, and from my perspective it looks great. Thanks for fixing the very-tall-screenshot issue; and I can no longer see jumping pixels :tada: 

Also:

> Smarter zoom - keep selected elements in sight

great idea - that's an awesome addition to the feature :+1:

mischlrebl:
I am eager for a test release :-)

It is not clearly visible from the initial screenshot, but shouldn't the four most important zoom features - which have an own icon - in, out, best fit and actual size be available on the tool menu to? as an own section after the redo.

thanks to all for the efforts!

KillyMXI:
I would argue against adding zoom icons to the top toolbar at this time.

- fugue icons are not well readable in this case, and nearly identical icons without text will be confusing;
- proper implementation is a whole lot of new design questions, it will still add clutter the toolbar, and the gain is rather small.

I thought about adding extra icons to the status bar, but I decided that current design is very straightforward and sufficient for practical use.
At least until enough evidence is collected to see where to move from here. (And then it will be better to apply the new knowledge to hopefully coming new editor instead of spending too much time perfecting the legacy one...)

KillyMXI:
Just to avoid any time waste due to miscommunication, I would love to know:

What exactly is "In progress" now?

I'm not aware of any blocking issues or features left. So in case you expect something - be sure to state it.

Lakritzator:
Sorry, I was really dead tired yesterday... having a 4yo is not without penalty in this time. I wanted to start my work at 07:00, due to him it was 09:45, so now at ~22 hours I finally finished work and I can review your changes... long days.

About your final question: I just added this issue to the things I want to have in 1.3 and while I didn't finish the review yet, it is automatically on "in progress". I just finished reviewing all the code, amazing job!

During my test I found more weird drawing artifacts, I am sure that they were already there before this PR, at least 99%. Just amazed about how tough it is to test such things. 

I do wonder where the Fraction struct comes from, not that we receive code which is under a conflicting license. Without having this cleared I'm afraid I cannot merge yet, do you remember?

Due to that question, and the size of the PR, we are finally activating the CLA bot, to protect us both. Not that you wonder, we had this on our roadmap for quite some time now.

KillyMXI:
Fraction struct is completely mine.
I did a quick look for Nuget packages that might offer Fractional/Rational numbers but found all of them not quite fitting the basic task.
It was easier for me just to write it from scratch.
Thinking about retrospectively, maybe I should've called it "Ratio" instead of "Fraction" - this will make the purpose and distinction cleaner.

I also briefly though about making a more fleshed out analogue as a Nuget package, but I will probably find no motivation for that unless I run into another use for it.

KillyMXI:
Any time estimates?

jklingen:
@KillyMXI we have finally set up the CLA Assistent today, but we want to give it another test. If everything is fine, from my point of view we're ready to merge.

Lakritzator:
I tested the CLA bot with another PR, so far it looks quite good.

By writing the following the bot should check the CLA:
/check-cla

KillyMXI:
Signed

Lakritzator:
(This was started before your signed, I was busy putting my kid to bed, he still doesn't sleep and it'S 23:00 hours)

@KillyMXI so we got it working, your PR is really cool but we found we need to protect you and us from contributions which cause discussions.

You only need to confirm the CLA once, no matter how many PRs you made, it's just for our cooperation.

Why we stated with this now? Nobody made us a PR as big as you did!!!
You must also let us know how you would like to be mentioned in the blog entry we are going to write.

KillyMXI:
> how you would like to be mentioned in the blog entry we are going to write.

"KillyMXI", linked to my GitHub profile, should be good enough.

nicorac:
@KillyMXI your PR is awesome. I really missed the Zoom feature, thank you so much.
I usually work with very large screenshots and I'm now finally able to edit them more accurately.

> > Maybe add Ctrl + Mousewheel support, too?
> 
> I'm not a fan of it. It annoys me more often than it helps, when I accidentally trigger zoom with scroll in any apps. And the way different mice emit scroll events - is a pain on it's own.

Well, let me put my 2 cents:
I was waiting zoom so much that, just after upgrading to latest (unstable) version, I took a screenshot and tried CTRL+Zoom... nothing 😢. 
My first thought was "there's something wrong, this is still the old version", then I noticed the Zoom menu and the bottom-right and said "ok, not yet implemented".

IMHO CTRL+Wheel is really natural (and widespread), and all the softwares I use day-by-day have it: Firefox, Chrome, paint.NET, GIMP, Notepad++, Office, LibreOffice, VisualStudio, VSCode to name a few.

I respect your opinion and work, but please reconsider adding it, maybe with an option to enable/disable it.

KillyMXI:
I suggest to open a separate issue to track mouse scroll zoom..
I see some design challenges there. I would've added it already if it was as straightforward as keyboard hotkeys.
I also have a few other tasks piled up while working on this - non-sponsored work will have to wait.

mischlrebl:
Thank you @KillyMXI - great work!

Scrolling on the windows up and down as well left and right - nice implemenation with the shift + scroll - isn't so handy.
Probably shift + mouse pressing to get a hand tool like in Acrobat to move the whole canvas would be useful.

mischlrebl:
appologize, dragging the canvas in Acrobat Reader is SPACE + mouse (not shift)

PR #212, opened 2020-05-29
Fixes a minor typo in the English language files

sirboffalot:
Fixes a minor typo on the term 'Windows 7' in the English language files.


Lakritzator:
Great find!

Actually the text should be changed a bit, as it also works for Windows 8 & 10, but those were not created before we wrote that text. 🤷

PR #216, opened 2020-06-27
Add placeholder for random alphanumerics in filename

peterfab9845:
This pull request resolves [FEATURE-762](https://greenshot.atlassian.net/browse/FEATURE-762) by adding the `${RR...}` filename placeholder.
The en-US language file for pattern help is updated, but I'm not sure how to request translations for the others. Improvements to the wording are welcome.


Lakritzator:
That is an interesting idea! I would consider this a minor use-case, meaning that not many people will use it.
But the great part of it, is that it's not in the way of anyone, like overloading the UI with another option, and therefore is something I don't mind merging this.

Lakritzator:
One task is left for us though, we might need to get translations for this feature as the other help files are not updated.
As it's a minor feature, that is okay for now.

PR #228, opened 2020-08-08
Update setup.iss

bovirus:
@Lakritzator 

Add Italian translation for additional lanuages
Modify script to use "Uninstall" string localization
Add "Uninstall" string localization for English and Italian.

Please check and merge.
If it's ok please build a new beta. 

Thanks.

### DIFF
--- a/Greenshot/releases/innosetup/setup.iss
+++ b/Greenshot/releases/innosetup/setup.iss
@@ -189,7 +189,7 @@ Root: HKLM; Subkey: Software\Classes\Greenshot\shell\open\command; ValueType: st
 
 [Icons]
 Name: {group}\{#ExeName}; Filename: {app}\{#ExeName}.exe; WorkingDir: {app}; AppUserModelID: "{#ExeName}"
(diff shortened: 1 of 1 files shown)


Lakritzator:
Nice!

I think having the language names in the language of the installer is a very nice premium touch, but if you speak a language I assume you also know the native name.

Just something we, the maintainers, need to think about:
If we add a new language, we need to translate this into all the languages that have this premium feature.

Btw. if you have Visual Studio 2019 community (or better) installed, you could build an installer yourself.
This is how:
* Change the solution to use release
* Rebuild all
* The installer .exe will be placed into Greenshot\releases

bovirus:
I think that If you select Italian and know other language you can want to install other language.

I believe that as translator I prefer to have as much as possible translated (also language name).

Please mrege my proposal for installer changes.
In Italian sound better to have all language names translated.

Lakritzator:
Oh don't get me wrong, I will merge it. I was underway add seeing a diff on a mobile phone is not a good experience :)

PR #239, opened 2020-08-13
Update setup.iss

bovirus:
@Lakritzator 

Update Inno Setup script
Add to default language name the prefix "en." to enhance for English base language
Add "ShowFile" string to use in Icon description" like "Show Readme.txt" and "Show License.txt" (now is only "Readme.txt" and "License.txt"
Add translation for "ShowFile" string

### DIFF
--- a/Greenshot/releases/innosetup/setup.iss
+++ b/Greenshot/releases/innosetup/setup.iss
@@ -190,8 +190,8 @@ Root: HKLM; Subkey: Software\Classes\Greenshot\shell\open\command; ValueType: st
 [Icons]
 Name: {group}\{#ExeName}; Filename: {app}\{#ExeName}.exe; WorkingDir: {app}; AppUserModelID: "{#ExeName}"
 Name: {group}\{cm:UninstallIconDescription} {#ExeName}; Filename: {uninstallexe}; WorkingDir: {app};
-Name: {group}\Readme.txt; Filename: {app}\readme.txt; WorkingDir: {app}
-Name: {group}\License.txt; Filename: {app}\license.txt; WorkingDir: {app}
+Name: {group}\{cm:ShowFile} Readme.txt; Filename: {app}\readme.txt; WorkingDir: {app}
+Name: {group}\{cm:ShowFile} License.txt; Filename: {app}\license.txt; WorkingDir: {app}
 
 [Languages]
 Name: en; MessagesFile: compiler:Default.isl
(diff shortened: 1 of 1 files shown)


Greenshot/releases/innosetup/setup.iss
```diff
@@ -214,42 +214,42 @@ Name: startup; Description: {cm:startup}
...
+en.ruRU=Pусский
+en.skSK=Slovenčina
+en.slSI=Slovenščina
+en.srRS=Српски
+en.svSE=Svenska
+en.trTR=Türk
+en.ukUA=Українська
+en.viVN=Việt
+en.zhCN=简体中文
+en.zhTW=繁體中文
```
Lakritzator, line 252: These translations have nothing to do with the english language, this is why it's not prefixed.
All names of the languages are how they are written in their own language!
I will not apply this part of the change.

bovirus: @Lakritzator 

I'm, so sorry but the rules of Inno Setup script are different.

For the rules of Inno Setup these strings are specific for English (master/base) language.
And you decide to leave when you select English the language names in original language.
Than for mutuation these strings (language names) are used when is not available strings for other languages (as it happen in this case for Italian language that has translation for these language names).

Please take care that a standard in "Inno Setup" to use "en." to higlight that are strings used when you select "English language" in the installer.

For who doesn't known well Inno Setup if you leave without "en." seems that these strings about language names cannot be translated. 
**_And is not true. Can be translated._**

If you use "en.xxx" for other English (base) strings in .iss script you should use "en.xxx" also for language names.
because the rules are the same.

Please use a coerent schematic approach at setup.iss translation strings.

And it help the transltor to identify the strings that can be translated.

Thanks.

jklingen: @bovirus thanks a lot for your contributons :smiley: 

I understand your point. These strings _could_ be translated. IMO it does not add much value, though. It is a wide-spread pattern to list each language in their own name, and I think that most users are absolutely fine with it.

But then, why not do it anyway? Because of maintenance. Maintaining translations is a piece of extra-work for us. (Note that not every translator is able/willing to prepare pull requests as carefully as you do.) In theory, adding a new language would eventually result in multiple small pull requests to have its name translated to every other language. In practice, translating a list of language names is a tedious task for translators and keeping them up to date is awful, so we would most likely end up with a semi-translated language lists in most languages.

So I agree with @Lakritzator and would rather rather prefer to stick with the current pattern for language names.

And yeah... we definitely need to do something about our translation process, I know :confused:

bovirus: @Lakritzator 
@jklingen 

I have a different vision about translation.

I think that a translation should translate as much as possible teh strinmgs and not leave if you can strings untranslated.

About extra work you can do one time and it was done.

bovirus: @Lakritzator 
@jklingen 

Please conisder to modify the installer script to remove 

- Manual selection of GUI languages
- Manual selection of plugins

and modify the installer script to

- Install all languages available
- Install al plugins available

In this way you 

- make simple the installer and relative maintenance
- are you sure that all languages/plugins are installed (not required additional setup to add others).

Thanks.

PR #242, opened 2020-08-16
Update English language

bovirus:
@Lakritzator 

Add strings for "about_translation" (useful for translator).
This string available in other languages.

Please merge.

### DIFF
--- a/Greenshot/Languages/language-en-US.xml
+++ b/Greenshot/Languages/language-en-US.xml
@@ -9,6 +9,7 @@
 Greenshot comes with ABSOLUTELY NO WARRANTY. This is free software, and you are welcome to redistribute it under certain conditions.
 Details about the GNU General Public License:</resource>
 		<resource name="about_title">About Greenshot</resource>
+		<resource name="about_translation">Info about translation</resource>
 		<resource name="application_title">Greenshot - the revolutionary screenshot utility</resource>
 		<resource name="bugreport_cancel">Close</resource>
 		<resource name="bugreport_info">Sorry, an unexpected error occured.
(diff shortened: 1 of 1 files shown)


Greenshot/Languages/language-en-US.xml
```diff
@@ -9,6 +9,7 @@
 Greenshot comes with ABSOLUTELY NO WARRANTY. This is free software, and you are welcome to redistribute it under certain conditions.
 Details about the GNU General Public License:</resource>
 		<resource name="about_title">About Greenshot</resource>
+		<resource name="about_translation">Info about translation</resource>
```
jklingen:
```suggestion -1+1
		<resource name="about_translation"><!-- diplayed in about dialog, use to add credits or other translation-related info, e.g. "English translation by Jane Doe" --></resource>
```

I think we had a comment in here a while ago explaining how to add translator credits, I don't know when and why it got lost.

I prefer this variant, as it does not show up in the english about dialog :smile:  what do you think?

bovirus: @Lakritzator 

Your proposal it's a good idea,.

Please take care that I I used for english language the same identifier of translated language file because otherwise if you make a compare with xml tool (like xmltreNav" you will find one orphan string (in translated language).

For that I propose my solution that use the same identifier of translated language file.
And align English/translated identifier list.

I beleieve that you can leave my text string with "about_translation" identifier and add before this identifier the comment that you posted to help the translator to fill in the string "about_translation".

bovirus:
@jklingen 
@Lakritzator 

Could you please apply this pR then I can close it?

Greenshot/Languages/language-en-US.xml
jklingen: Thanks for getting back. I just changed my proposal so that the comment is now included within the "about_translation" entry, so the XML node for that `name` exists (but is empty). Having an acutal text node in there would make it necessary to have language-specific logic in our about dialog.

I know, translating Greenshot is currently a bit cumbersome. Some time in the future we'll replace this file-based pull request workflow with some platform approach.

Thanks for your contribution.

jklingen (approved):

PR #255, opened 2020-10-23
Get rid of embedded browser

Lakritzator:
This change will make it possible to get rid of the embedded browser, as it was causing issues due to compatibility.

The most important change here, is that instead of opening an embedded browser, we open the default system browser and have imgur redirect to https://getgreenshot.org/oauth/<service>. Here we have a page with makes sure the running Greenshot instance gets the information.
The current solution uses title changes, which are picked up by Greenshot, to communicate. This is a one-way communication, we need to add better error handling / timeouts and see if it works in all cases.

A different solution would be to use a custom URL handler, this might cause the browser to ask security questions though. Also it might be possible to have Greenshot start a HTTP listener, but this might cause issues with mixing http/https (need to test).


Lakritzator:
I just wanted to elaborate a bit on this change:
By using the default browser, we solve the issue that the .NET Embedded browser is not compatible with that what certain cloud service providers need. Also it solves some security issues.

Due to this change, when you need to authenticate (should not happen often), the browser is opened with a link to the cloud provider, after the authentication the browser is redirected to https://getgreenshot.org/oauth/<service> and behind the scenes Greenshot is receiving the information for the authentication, so it can upload the capture, and you can close the browser shortly after.

FYI: The page which is displayed will show ads, as always on our site (but we clearly mark them), we cannot provide Greenshot for free without doing something like that.

Lakritzator:
In the mean time the current state is that the following work:

- Box
- DropBox
- Imgur

Which are the most important ones, Picasa is no longer supported (moved to Google Photos) and Photobucket needs to be fixed.

For all three we need new client id & secrets as they need a different redirect URL to make the new functionality possible.

Lakritzator:
Oops, I forgot Flickr... Need to check the current state of the code.

Lakritzator:
Flickr is still using the, now quite old, OAuth 1.0a standard. Meaning that I can't use the work I did for the other services, I need to check what this means.

Lakritzator:
❗ Before we release this branch, we need to modify the build in such a way that we can use newer client id & secrets for these services, we can't modify the redirect URL for all of them as it will impact current users of 1.2.x

PR #257, opened 2020-11-24
Update de-de.markdown

Mr-Update:
Missing settings added and formulations adjusted.


jklingen:
Sorry for the late reponse :see_no_evil: looks great, thanks for your help improving that document. :+1:

Lakritzator:
@jklingen Sourceforge is still mentioned, we should remove that.

jklingen:
@Lakritzator good point - I removed the translation section from all help files (as the complete process is outdated anyway), outside of this pull request

PR #276, opened 2021-02-04
Select freehand tool as default

ChrisNeedSpace:


jklingen:
Thanks for your suggestion.

I am afraid there's now way we could merge this pull request as it currently is. The solution you provided is very specific to your own use case, but would be quite annoying for many other users who don't use the freehand tool frequently.

In order to work well for everyone, this IMO should rather be an option to remember the last used editor tool, or an option for the user to select a default tool in the settings.

The former has been requested a few times in our JIRA, but it was not implemented yet: https://greenshot.atlassian.net/browse/FEATURE-662

For Greenshot 1.3, we're not planning to add more options, because it means a lot of work (due to the underlying tech stack) and requires new translations.

jklingen:
Closing this, as there was no more feedback.

Issue #268, opened 2021-01-24
Feature request: close Greenshot image editor, after copying the image by Ctrl+Shift+C ?


ChrisNeedSpace:
Is it possible?
Can you please implement it (e.g. as an option)?

ChrisNeedSpace:
Is there anybody out there?  :)

jklingen:
see here: https://greenshot.atlassian.net/browse/FEATURE-359

PR #277, opened 2021-02-04
#268 coded - proposition

ChrisNeedSpace:
Added CloseForm bool parameter.
Hardcoded to true for ClipboardDestination. It might be added to options (but I did not invest time to learn how options are coded...).


jklingen:
Thanks again.

Please see my comment in [your other pull request](https://github.com/greenshot/greenshot/pull/276#issuecomment-782860886). This one is also very specific to your use-case, while it should rather be an option.

This has also been requested in our JIRA, but not implemented yet: https://greenshot.atlassian.net/browse/FEATURE-359

PR #292, opened 2021-03-20
Update URLs to HTTPS

TotalCaesar659:


Lakritzator:
Yeah, that makes sense! Thanks for finding that, and providing a PR.

Issue #116, opened 2019-03-08
Capture Window not working anymore


Rouzax:
**Describe the bug**
I believe this started to happen since my update to Windows 10 version 1803.
When ever I try to capture just window Greenshot will only select my entire screen, or if I move the capture to the Windows Taskbar it will capture that as a window. All other windows are not recognized and selected

**Expected behavior**
Select and capture only the window

**Screenshots**
![image](https://user-images.githubusercontent.com/4103090/54016206-7a9a5780-4182-11e9-9dfd-196e5b3507e3.png)

**Desktop (please complete the following information):**
 - OS: Windows 10 Version 1803
 - Greenshot: 1.2.10 Build 6

Rouzax:
Hmm,
After a reboot it started to work again but it seems to stop working after some time.

TD-er:
I have issues too since a couple of days (3 weeks at most I guess)
It may take minutes for Greenshot to appear when pressing the PrntScr button.
Opening the context menu from the icon bar then also takes minutes to appear.
Killing Greenshot and starting it again will make it perform fast again.

Also if it has been used recently it reacts fast, but after some time (few hours) it will take minutes again to get some reaction from Greenshot.
I should be running the last released version:
![image](https://user-images.githubusercontent.com/3751318/54925847-735ba380-4f0f-11e9-8826-6039cd373752.png)

Shockah:
Also having the same issue. Killing and starting explorer back again seems to fix it. Didn't figure out the cause yet though.

Lakritzator:
When creating an issue here, there is an template which is clearly not read... we have a separate system for support & bugs from a users perspective.

That said... 

Can you make a list of tools / utilities you use? They might have influence on the way Greenshot works.

Rouzax:
@Lakritzator, I'm used that all the tools I use do their issues on GitHub as well and I indeed missed that section in the Create Issue template, sorry.

I migrated to ShareX so can't really be of any help but this is my installes Sofwtare list

DisplayName | DisplayVersion | Publisher
-- | -- | --
Adobe Acrobat DC | 1.901.220.040 | Adobe Systems Incorporated
Adobe Creative Cloud | 4.9.0.504 | Adobe Systems Incorporated
Adobe Photoshop CC 2019 | 20.0.5 | Adobe Systems Incorporated
Adobe Refresh Manager | 1.8.0 | Adobe Systems Incorporated
[...]

PR #304, opened 2021-04-02
Build fixes

SamB:
Just a few build improvements I came up with due to debugger misconfiguration.

In particular, I had "Break when exceptions cross AppDomain or managed/native boundaries" enabled,
so the debugger broke in on a failed call somewhere inside WasCurrentProcessToastActivated(),
at which point I noticed that the debugger wasn't finding the symbols.

This turned out to be because the PDB files from NuGet dependencies weren't being propagated along with the DLLs.

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -107,16 +107,4 @@
   <ItemGroup>
     <PackageReference Include="MSBuildTasks" Version="1.5.0.235" GeneratePathProperty="true" DevelopmentDependency="true" />
   </ItemGroup>
-
-  <Target Name="PostBuild" BeforeTargets="PostBuildEvent" Condition="$(MSBuildProjectName.Contains('Plugin')) And !$(MSBuildProjectName.Contains('Test')) And !$(MSBuildProjectName.Contains('Demo'))">
-    <Exec Command="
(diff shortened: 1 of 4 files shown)


Lakritzator:
Cool, very thoughtful.
But I cannot merge this.
It might look like your changes make sense, but I do not want to have direct connection between Greenshot and the plugins.
Especially there should not be any leaking of dependencies, and this will happend if you have a direct project reference.
It should be possible to install Greenshot without having any plugins.

About the toast error, I moved from my own toast implementation to the "Microsoft.Toolkit.Uwp.Notifications" which seems to have an issue which I fixed myself in my code. I did have a quick check, but couldn't find it yet. I just upgraded the nuget locally, but don't know yet if this solves the issue... Honestly, I just added a try catch for now...

SamB:
Well, there wasn't anything wrong with the code as far as I could tell,
actually, I just had enabled a setting where any time COM code would return
an error code, the debugger would kick in, and it just so happened to land
there.

Re: not wanting to actually import the plugins into the main app, that
makes sense. Perhaps a separate project to pull in all the components and
build the installer would be a good idea, though? The XCOPY thing seems not
to "clean" properly. (There are files left behind after you run the clean
target for the solution -- note that I had to do this as `msbuild -t:clean
-restore` for some reason.)

SamB:
By the way, the reason I was trying to run in the debugger in the first place was that I was seeing something like Issue #116 (desktop was really eager to be clicked in "Capture Window" mode ...), but naturally it wasn't happening in my freshly launched build.

PR #330, opened 2021-09-07
Corrected czech translation

Masv-MiR:
Changed from Ceština into Čeština.


Lakritzator:
A small change, but I can only imagine how irritated Czech people would feel. Thanks!

PR #332, opened 2021-09-30
Update of Czech language file

svatas:
I just reviewed that language file and updated with latest strings from en_us file, also corrected and reviewed used terminology in application.


Lakritzator:
I can't really review this in details, but it looks sensible. Thank you for your effort.

PR #334, opened 2021-10-05
FEATURE-1291: Image Editor - delete file and close

lupinitylabs:
This PR adds functionality as requested in https://greenshot.atlassian.net/browse/FEATURE-1291 (and by a friend of mine).

The Issue
===

A friend described the underlying issue as follows:

He uses the tool with the following destination settings checked:
- Save directly (using preferred file output setting)
- Open in image editor

Every time a screenshot is created, the screenshot is saved to disk and the editor is opened. 

He uses the tool mainly for documentation and annotation, saves some of the screenshots, and discards others. However, even when not saving changes, the initial screenshot stays in the destination path with no way of easily deleting it, so the directory clutters up very quickly.

The requested solution was to have an easy way to quickly delete an image from disk and close the editor.

What it does
===

The PR adds a new "discard" button to the Image Editor ToolStrip, that asks the user if they want to delete the screenshot on disk and close the editor. 

The same function is called if the user presses the `DELETE` key. It felt natural to choose this key - I understand that it is also used to remove the currently selected alteration applied to a screenshot, and that's why the action always requires the user to confirm a dialog so that they don't accidently delete the whole screenshot they are working on.

Another key could be chosen, but we felt that the delete key would be the most intuitive way.

Errors when deleting the File are handled in a try..catch block, but just ignored. The correct handling of errors was verified by exclusively locking the screenshot before discarding.

Further Implications
===

A new button and a new dialog text and title were added and provided in `en_US` and `de_DE`, but other languages are missing. I suppose Hacktoberfest is a good time to try to get those translated ;-)


lupinitylabs:
Robin, just wanted to let you know you can also feel free to close this if you think it's not the right way to go.
I know how hard it is for maintainers to say no sometimes 😁 Just saying don't feel forced to leave this dangling. 👍

PR #338, opened 2021-10-08
add shortcuts for foreground color, background color, line thickness, bold and shadow

meschi89:
this increases the user-experience a lot, and saves time.

### DIFF
--- a/src/Greenshot.Base/Interfaces/Drawing/Container.cs
+++ b/src/Greenshot.Base/Interfaces/Drawing/Container.cs
@@ -111,6 +111,10 @@ namespace Greenshot.Base.Interfaces.Drawing
         bool HasIntersectingFilters(Rectangle clipRectangle);
         bool IntersectsWith(Rectangle clipRectangle);
         void Draw(Graphics g, Bitmap bitmap, RenderMode renderMode, Rectangle clipRectangle);
+        void SetForegroundColor(Color color);
+        void SetBackgroundColor(Color color);
+        int IncreaseLineThickness(int increaseBy);
+        bool FlipShadow();
         void Invalidate();
         void PullElementsToTop(IDrawableContainerList elements);
         bool CanPushDown(IDrawableContainerList elements);
--- a/src/Greenshot.Base/Interfaces/ISurface.cs
+++ b/src/Greenshot.Base/Interfaces/ISurface.cs
@@ -38,6 +38,10 @@ namespace Greenshot.Base.Interfaces
         event SurfaceMessageEventHandler SurfaceMessage;
         event SurfaceDrawingModeEventHandler DrawingModeChanged;
         event SurfaceElementEventHandler MovingElementChanged;
+        event SurfaceForegroundColorEventHandler ForegroundColorChanged;
+        event SurfaceBackgroundColorEventHandler BackgroundColorChanged;
+        event SurfaceLineThicknessEventHandler LineThicknessChanged;
+        event SurfaceShadowEventHandler ShadowChanged;
 
         /// <summary>
         /// Start value of the step-labels (counts)
--- a/src/Greenshot.Base/Interfaces/SurfaceBackgroundColorEventArgs.cs
+++ b/src/Greenshot.Base/Interfaces/SurfaceBackgroundColorEventArgs.cs
@@ -1,0 +1,31 @@
+/*
+ * Greenshot - a free and open source screenshot tool
+ * Copyright (C) 2007-2021 Thomas Braun, Jens Klingen, Robin Krom
+ *
+ * For more information see: https://getgreenshot.org/
+ * The Greenshot project is hosted on GitHub https://github.com/greenshot/greenshot
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 1 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+using System;
+using System.Drawing;
+
+namespace Greenshot.Base.Interfaces
+{
+    public class SurfaceBackgroundColorEventArgs : EventArgs
+    {
+        public Color Color { get; set; }
+    }
+}
--- a/src/Greenshot.Base/Interfaces/SurfaceBackgroundColorEventHandler.cs
+++ b/src/Greenshot.Base/Interfaces/SurfaceBackgroundColorEventHandler.cs
@@ -1,0 +1,25 @@
+/*
+ * Greenshot - a free and open source screenshot tool
+ * Copyright (C) 2007-2021 Thomas Braun, Jens Klingen, Robin Krom
+ *
+ * For more information see: https://getgreenshot.org/
+ * The Greenshot project is hosted on GitHub https://github.com/greenshot/greenshot
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 1 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+namespace Greenshot.Base.Interfaces
+{
+    public delegate void SurfaceBackgroundColorEventHandler (object sender, SurfaceBackgroundColorEventArgs e);
+}
--- a/src/Greenshot.Base/Interfaces/SurfaceForegroundColorEventArgs.cs
+++ b/src/Greenshot.Base/Interfaces/SurfaceForegroundColorEventArgs.cs
@@ -1,0 +1,31 @@
+/*
+ * Greenshot - a free and open source screenshot tool
+ * Copyright (C) 2007-2021 Thomas Braun, Jens Klingen, Robin Krom
+ *
+ * For more information see: https://getgreenshot.org/
+ * The Greenshot project is hosted on GitHub https://github.com/greenshot/greenshot
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 1 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+using System;
+using System.Drawing;
+
+namespace Greenshot.Base.Interfaces
+{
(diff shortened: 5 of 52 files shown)


mischlrebl:
this PR fulfills the requirements of the request https://greenshot.atlassian.net/browse/FEATURE-1125

there is an inconsistency atm where the expertise from the maintainers is needed with their know how.
the attributes of an object like line thickness and shadow are persisted while the colors are not. they all should behave equal, most likely stored. thanks a lot for your efforts!

Lakritzator:
So, I managed to take a look at this and I can see that this is really cool.
I do not understand what you mean with that the colors are not persisted. What do you mean?

Some issues I have, which make it impossible to merge like this:
This PR adds new features to the editor which use keys, in this case keys which were already used for the new zoom functionality.
image
We need to discuss in the team what keys are used for what.

Also you made the key handling inside the surface, which is more or less the drawing canvas. As these have effects on the editor, you need to propagate the information to the editor, by adding new events to the surface. I think it doesn't make sense for the surface to handle keys, especially not for a functionality which should be limited to the editor context (the surface doesn't care about it, just wants to draw). This would reduce a lot of code and complexity. (I know, it does key handling already, and I am not quite sure why we made this poor choice for that functionality.)

You also changed the language files, for the tool-tips, this is the thing I am not so sure about. Especially as "numpad" is very English 😆

Some future improvements would be to make the presets completely configurable.
I am actually wondering if it makes sense to have a preset which is more than just the color of the line or the background, wouldn't it make sense to have a whole set like both the colors, line width, shadow, font etc. so one can apply these quickly?

Lakritzator:
Btw. are you from the same company, or work together? "we find very cool as well"

meschi89:
@Lakritzator I had again a look at it.
I check the place for the code. In my opinion the Surface::ProcessCmdKey is the correct place for the following reasons:

1. As the already existing handling of Left and Right key, this also something similar.
2. The ImageEditor does not have the selectedElements, which is needed.
3. The Surface.cs already contains the logic of the shift modifier.

I reverted all shortcut hints except german, english and dutch.

I left two todo's, which can cause problems in some views. Adding spacing .xml is not a good idea! If you like I can also fix them.

mischlrebl:
@Lakritzator would be great to make a next step in this topic, thank you

Lakritzator:
Although I strongly believe we wrote the way the editor and surface work wrong in the first place, the surface is for the drawing itself and the editor for changing the values, I see from a code standpoint absolutely no reason to enforce that right now. So I am fine.

I tried the feature in the current state, and it works great, so there is that.

@jklingen I'd like your opinion on this, the feature is described in https://greenshot.atlassian.net/browse/FEATURE-1125
I'm okay to merge this, but maybe you see some important facts I overlooked.
I think we might get into discussions about the keys some time later when other features are added, but to be honest... I do not have a crystal orb to look into the future and I also like more community driven features. They did such a great job, it would a let down not to merge this. My standpoint on the feature from a translation and documentation, it doesn't hurt when there is none and we deliver that some other time. We have more features which are not documented 🤷

jklingen:
Hey there, I think color shortcuts are a great addition. Thanks a lot for your contribution 🙌 

In a future version of Greenshot, I'd like to have this feature more discoverable, but doing this now (with the old ugly color picker and us not wanting to make big changes in the UI) would be too much effort for my taste. So, I am basically fine with adding this (undocumented and untranslated) as a power user feature.

The only thing I'm not happy with is using the numbers (without modifiers) as shortcuts.  Chances are that a lot of users will change colors unintentionally (e.g. if they want to enter numbers into a text element, but the element is not in editing mode), which would be an absolutely confusing experience.

So, what do you think about changing these to use e.g. `Ctrl`+`Shift`+numbers instead of only-numbers?

mischlrebl:
thanks a lot both for your positive feedbacks!

the guidance was on the existing shortcuts like r, t, … without modifiers and the approach for a fast and productive user experience in the editor and to keep that further.

if a user hits the color change by random he is probably thankful to explore a new, useful feature so easy.

for sure I am going to extend the (English) documentation with the behaviour as soon as it is merged an released.

Lakritzator:
Thanks @jklingen for looking at it, and I think it was good to discuss it.
I'm not against the numbers without the modifiers, there is always undo when it was by mistake.
With modifiers it would complicate the feature too much.

I'm merging it as it is right now, and we see how we go from there if we get issues on it.

Lakritzator:
Build is now available as version 1.3.231 here https://github.com/greenshot/greenshot/releases

Lakritzator:
Thanks for the great work, sorry that it took us a while to merge.

mischlrebl:
thanks a lot again for taking it in now and committing it too - you made my weekend! 👍 

still, three points are to consider:

"_...there is always undo when it was by mistake_"
seems not be completely true. works for line thickness, but not for color changes.

from the discussion above in October, this small inconsistency exists:
_persistance: bold and shadow remember the current state when adding a new object of the same type. the colors fall back to what the preset is from the tool itself._

furthermore, the behaviour in the end was, that ctrl - +- (on numpad) changed the text size instead of border line by 5px and * (numpad) toggled the boldness of the text. it seems that these changes were not committed at the end.
I will have a look on it with @meschi89 but he is preparing right now to departure for a six week trip overseas.... 😎 🌴

jklingen:
I _strongly_ disagree about the shortcuts.

Not every Greenshot user is an everyday user and interested in maximizing their screenshotting productivity. From my experience, the majority of users does not like surprises.

Switching tools with a single key press is one thing, but modifying the user's work is another thing and AFAIK the Greenshot editor does not do that so far (except for functional keys like Del and arrows). If I am wrong and it does, I absolutely think it better shouldn't.

In the best case, the user notices the color change, thinks "ok that was weird" and fixes it. Other users get frustrated about unexpected things happening. Or maybe they just don't notice or don't care and send the screenshot to their co-worker, and when asked about the colorful annotations they say "dunno. Greenshot is doing this sometime" - "oh. ok. I'll stick to copy-pasting into Powerpoint then ha ha".

I really think we should add modifiers, even though this has already been merged.

Lakritzator:
Although I think that a user who presses a key "by accident" is not very likely, I do agree that it's not directly clear why something happens. I value the opinion of Jens a lot, and he is right that we switch tools with a single key but not change anything directly. This will also cause some issues later if we do not change it, I will take care of it right away... sorry about my enthusiasm of merging this, but I am always very excited about community support.

About the undo, I already knew this and was looking at it. It should work consistent throughout the editor.

Lakritzator:
@jklingen the numpad behaves completely different from the normal number keys, the shift modifier is not really usable here.
Any suggestions here?

jklingen:
Thanks a lot @Lakritzator, I really appreciate that  👍  
Well, I have witnessed a lot of UX sessions in different projects, and it always surprises me what kind of unexpected things people do while using applications - and these "wow, what did just happen?!" moments are often really awkward.

With regards to the numpad. Does it actually add value to have the numpad hotkeys if the same functionality can be achieved with the number keys? 🤔 For my taste, we could leave them out. Or do I overlook some special feature here?

mischlrebl:
this is the current state of the implementation (as mentioned earlier, something was not committed as it seems around Shift +- and *, see lower table. but will not get access to the laptop before Wednesday):

image

the numpad definitively adds much value for the experienced users - which exists out there as well. working only with the character keys has to be a possibility for the laptop users without a full keyboard. and some options are available only on the numpad at the moment.

to work with single key strokes is a very fast way as only one hand is needed all the time while the other resists on the mouse. that was the basic idea.

possible solutions i see at the moment:
- implement ctrl-z fully, a user has then all the time the possibility to come back to the previous state
- make the whole option enableable
- add ctrl-shift on every single stroke key, whereby the shift options resists as they are (has to be checked if that works with the keypad numbers as well as the shift itself isn‘t applicable)

Lakritzator:
I made some changes to the way the values are applied, I do not know why we made it so complicated but that is how it's done so all events & code paths are correctly called so things land on the undo-stack.

Lakritzator:
The current state, as mentioned in my previous comment, is available in 1.3.234 together with a fix for an annoying Slack issue (they change our window size, the change counters it.).

mischlrebl:
the current state in 1.3.234 seems to be:
image


details:
- the attributs are now stored, thank you
- the undo stack works particularly, but not with all changes
- ctrl-0 and -9 (keypad: fill color ) collides with zoom shortcuts, both are executed
- ctrl-- and -+ (numpad: line thickness) collides with zoom shortcuts, both are executed
- line color orange is no longer applicable (would be important to have a third state beside red and green like a traffic light)
- help text's are wrong
- green should be #00FF00 instead of  #008000 (RGB base colors should be used)

it is a pity that the numpad-numbers are no longer used.

Lakritzator:
Thank you for that feedback!!

I will look further into the undo, I believe it's a general thing and one of the reasons I would love to rewrite the editor.
About the green I am sure I kept it as was in the PR, but I can change it.
About orange, the key 0, there was some weird logic which probably blew up. So when changing the line color it should be orange, but when changing the fill color is should be transparent... when I consider myself using the feature, orange is a bit of the odd ones out...

The keys for the zoom in the editor were already mentioned in my comment on Oct the 18th of 2021, but I will place them here again:
image
Due to the fact that we split the key handling over the Editor and the surface, I overlooked the double assignments.
I had them in the back of my mind, but forgot about why we used the 0-9 in the first place.

So we cannot use those key with the specified modifiers, or would need to change them.
@jklingen do you want to come up with a keyboard scheme, which works with both features?

I will not modify the help text, until we have a consensus.

jklingen:
Oh sorry... just noticed that I was mentioned here. I absolutely fail to keep up with the flood of Github notifications 🙈 

Phew, thanks for pointing that out, I totally forgot about the zoom shortcuts.

I'd definitely stick to `Ctrl+0` for the zoom, because people are used to that from many apps, especially browsers. Same for `Ctrl++` and `Ctrl+-`... these are just ubiquitous.

So what about

- `Ctrl+Shift`+numbers for line colors
- `Ctrl+Shift+Alt`+numbers for fill colors
- and maybe something like `Ctrl+Shift` + comma, dot and/or minus for line thickness and shadow?

@mischlrebl I know that these wouldn't be as beautiful for you to operate like the ones you suggested. I just still think that this is an absolute power user feature and that this shouldn't "reserve" ~22 of the most accessible shortcuts... and I feel that the fact that some of them are already in use by more common features supports my point.

FWIW, for Future Greenshot we'll definititely need come up with a more elaborated shortcut concept. I imagine shortcuts with 2nd strokes for special things (e.g. first hit `Ctrl+Alt+F` (for fill color) and then a number to actually pick a color). But I really don't think this fits into Greenshot 1.3.

mischlrebl:
no discussion about the zoom shortcuts, they are there and standard everywhere

don't see a better approach atm as Ctrl+Shift and Ctrl+Shift+Alt. The big question is if the numpad keys would be useable too - which are not with Shift at least - to have them in parallel for keyboard users? It would also bring more possibilities for line thickness and shadow (see below). 

- _Ctrl+Shift+numbers for line colors_
- _Ctrl+Shift+Alt+numbers for fill colors_ and Ctrl+Shift+Numpad-numbers for the same
- _Ctrl+Shift + comma, dot for line thickness_
- Ctrl+Shift + minus is already used by some sort of Auto Crop! For me personally the shadow has no meaning and could be left.
- Ctrl+Shift +Alt+ comma, dot for text size
- Ctrl+Shift +Alt+ minus for text bold

_and I feel that the fact that some of them are already in use by more common features supports my point._ just to be precise, none of the initially suggested shortcuts where ever or are in use, they have been considering the current state very well.

jklingen:
I didn't mean to say that the original suggestions were conflicting. What I was trying to say is: when we reserve a lot of prominent shortcuts for for all of these different properties and colors, chances are that we stumble into conflicts. And even if we currently don't, the shortcuts probably will be missing for future features. Features are obviously already competing about precious shortcuts, sort of ;-) 

For the same reason, I am not a fan of increasing the number of shortcuts by duplicating them on the number pad for convenience. But I can live with that, I guess Greenshot will never have dedicated shortcuts on the numpad, since many devices don't have one.

So I think we can just go like you suggested. 👍 

But please don't be mad at us if we ever change the shortcuts at some point in the future. The more we're talking about this, the more I am sure we will need to thoroughly re-think our shortcut concept when we're planning about a new editor 🤔  

Thanks again for providing this feature 🙌

mischlrebl:
would be happy if you could round up the shortcuts as agreed now as the original submitter will enjoy his holidays some more weeks.

it is ok for me to have it without the numpad, but would be happy if Ctrl+Shift+0 could be orange (in opposite to Ctrl+Shift+Alt+0 where it is transparent). and if the later two for text options would find their way into the code that would be great.

thanks a lot for your efforts and the ongoig work 💯

PR #339, opened 2021-10-09
FEATURE-1196

EricCogen:
It would be great if when choosing color for an object, doucle clicking on the colour square chooses the color and closes the color window

### DIFF
--- a/src/Greenshot.Base/Controls/GreenshotDoubleClickButton.cs
+++ b/src/Greenshot.Base/Controls/GreenshotDoubleClickButton.cs
@@ -1,0 +1,12 @@
+﻿using System.Windows.Forms;
+
+namespace Greenshot.Base.Controls
+{
+    public class GreenshotDoubleClickButton : Button
+    {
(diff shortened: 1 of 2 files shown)


EricCogen:
Hi @Lakritzator,

The ResX change is not needed, probably from previewing the form in the editor, I will re-commit without this change.

As for your suggestion, that should work as well, I attempted this in place but the event wasn't firing so I abandoned that idea and did some research and happened upon the SystemInformation.DoubleClickTime and realized that it was just a matter of tracking the time between clicks.

Another solution here is to just not use buttons, perhaps labels so we don't have to fiddle with the setstyle, etc., is there a reason why it has to be buttons?

mischlrebl:
@Lakritzator thank you

Lakritzator:
@EricCogen I hope you are okay with me "guiding" your PR, I'm not sure what feels comfortable for you and what not. As I said in the beginning, your PR was spot on for the result. Thanks for helping us out, maybe you find something else which you are not happy with and think needs improvement. ❤️

PR #340, opened 2021-10-14
Fix missing paragraph tag

miltonhowe:
tag was missing nesting was off


Lakritzator:
Thank you for this, I have been looking at it for a couple of times, and am still confused why it shows up correct on github but not on the final page. As I have cleaned up the unstable versions, which were visible on the website, I needed a rebuild anyway.

Lakritzator:
FYI: due to caching on cloudflare, it takes a while for the website change to propagate.

PR #345, opened 2021-10-17
Update language-zh-TW.xml

SiderealArt:


Lakritzator:
Thank you for this, the only issue I have in the mean time is knowing how to validate translations.
It looks good from a quick check but @jklingen you okay with this?

src/Greenshot/Languages/language-zh-TW.xml
```diff
@@ -1,14 +1,15 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
-<language description="正體中文" ietf="zh-TW" version="1.1.4" languagegroup="9">
+<language description="正體中文" ietf="zh-TW" version="1.0.4" languagegroup="9">
```
jklingen:
```suggestion
<language description="正體中文" ietf="zh-TW" version="1.2.10" languagegroup="9">
```
We have been using the `version` attribute to keep track of which version the translation is based upon. (That said, I assume it has not been set consistently in all translation files, so I am not sure whether we can actually benefit from that anyway. But it does not hurt to update it.)

jklingen (commented):
@SiderealArt Thanks for updating this translation and sorry for not looking into this earlier. Looks fine for me, some indentations are off, but that does not hurt ;-)

PR #346, opened 2021-10-19
Toolbar item and menu item for copy and close in one click.

bholliger:
Solution for Feature Request #268 
https://github.com/greenshot/greenshot/issues/268

Signed-off-by: Bruno Holliger <bholliger@gmx.net>

### DIFF
--- a/Greenshot/Languages/help-de-DE.html
+++ b/Greenshot/Languages/help-de-DE.html
@@ -76,7 +76,7 @@
 		Im Bereichsmodus können Sie einen Bildschirmbereich auswählen, der abfotografiert
 		werden soll.<br>
 		Nach dem Starten des Bereichsmodus sehen Sie ein Fadenkreuz, das die Position
-		des Mousezeigers auf dem Bildschirm anzeigt. Klicken und halten Sie die linke
+		des Mauszeigers auf dem Bildschirm anzeigt. Klicken und halten Sie die linke
 		Maustaste an der Stelle, an der eine der Ecken Ihres Screenshots liegen soll.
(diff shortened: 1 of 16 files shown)


fabiochelly:
The best would be to use ESC for that: copy the current image then quit. It would be a huge improvement

Lakritzator:
@fabiochelly one of the challenges of building a tool, is to have as little "special" cases as possible. That sometimes means it has trade-offs, someone might really "badly" want one thing, others wouldn't use it.

ESC is already used to cancel a drawing mode, so we cannot just add this because a lot of people would press one time to many and the editor closes. It's also not common practice to close an editor window with ESC. This doesn't mean that this is not possible, it just needs a lot of thought.

You might say, add it as an option, which complicates the application with so many settings. We in fact have no UI for quite a lot of possible settings... just because they are not always worth confusing the users (and we also need to have "space" on the settings screen, and provide translations in many languages)

You already said "Greenshot is a lot lighter", that is exactly the fact.

I currently have more pressing matters, so sorry that I cannot come to an end on this topic, but I do this all in my limited free time.

fabiochelly:
I perfectly understand. For my use case, I just copy a region, add an
ellipse and paste the resulting image on WhatsApp 30 times a day (to help
my customers with my CRM).

You really should transfer the expert part of the options in a simple
ini/yml/json file for advanced users. No UI, no translation and the ability
to respond easily to many simple queries like this one. It works perfectly
on a lot of problems. I'm not a CS developer otherwise I would have added
myself. I just made a donation because I'm using a lot your app.

Keep up the good work,

Fabio Chelly


Le mer. 27 juil. 2022 à 12:56, Robin Krom ***@***.***> a
écrit :

Issue #283, opened 2021-02-22
Unicode icons in textboxes are not rendered correctly


scullhead:
Greenshot V1.2.10 Build 6 (64 Bit)
Windows 10 20H2

- It is possible to insert unicode emojis and icons into text boxes (e.G.: 😁✔ 🠆, ...) by copy & paste or by Pressing the "Win + . " Keys

- The unicode icons are shown corectly when the text box is in "edit mode":<br />![Unicode_Icons_Text_Edit_Mode](https://user-images.githubusercontent.com/10958573/108708622-f75ea180-7511-11eb-8e4d-07e4927b26c1.png)

- When you leave the edit mode, some icons are shown (but differently than in edit mode ( e.g. the check mark ✔) and some are not shown at all (just rectangles):<br />![Unicode_Icons_Preview_Mode](https://user-images.githubusercontent.com/10958573/108708778-2c6af400-7512-11eb-991e-b2ae8e34bb74.png)

Lakritzator:
I've added the fix to the most recent 1.3 build (1.3.151 or later), which is not officially released yet. 
You can find it here: https://github.com/greenshot/greenshot/releases

Lakritzator:
I actually have to roll-back my changes, the fix caused more harm than good.
It seems that the routine which allows me to use unicode, has other and much worse issues.

Lakritzator:
Just rolled back the change, yeah it made Unicode icons visible, but it caused a lot of issues.
I currently don't know how to solve it, without having more other issues.

Marking it as "known issue"

scullhead:
Ok, thanks for the info. Maybe it will be fixed another time. Currently it is more like a "nice to have" feature.

Lakritzator:
I fully agree this should be fixed, but if Windows doesn't offer a good API for the technology we are currently using for the editor (Windows Forms), it has to wait until we finally move to use WPF (or something else).

As a workaround, you should be able to drag & drop images from websites on the drawing canvas, this could be emoticons. Doesn't work with EVERY website, but a lot should work.

jairbubbles:
Hi @Lakritzator, have you considered using `TextRenderer` only when the text contains emojis / icons?

As an alternative, what would you think of adding a new object type "emoji" so that normal text blocks wouldn't be affected? It would help annotate and give impressions on a screenshot.

jairbubbles:
I dug a bit more and it looks like using a font that embeds directly the unicode emoji fixes the issue.
[...]

PR #357, opened 2022-01-08
Use "Segoe UI Emoji" as default font

jairbubbles:
Workaround for #283


Lakritzator:
From my point of view, the new emoji functionality on the other PR makes this obsolete. Changing the default font might have a bigger impact. What do you think?

jairbubbles:
Yes indeed!

PR #359, opened 2022-01-16
.NET 4.7.2 -> .NET 4.8; refactoring

unquietwiki:
Hey there. I know this is a large patch; it's kinda hard to avoid. I've compiled this & using it on my machine right now.

- .NET 4.8 is installed by default on Win10 since 2019, and is also in Win11. It also is supported back to Win7, which is out-of-support & not targeted by this anyway.
- You already had an .editorconfig; I made a few changes to follow the code guidelines.
- I used [Roslynator](https://marketplace.visualstudio.com/items?itemName=josefpihrt.Roslynator2019) to refactor the code. One thing that stood out: there were quite a number of "var" uses; the linting can do all var, or no var, so I chose no var based on the code guidelines.
- I deleted a bunch of unused ISS files from the installer.

Thanks for maintaining this software. I think I've even sent some PayPal over the years, and it's been incredible for documentation.


Lakritzator:
First, a big thanks for going through the work on this and supporting Greenshot.

I'll need to think about if I really want to force people to move to 4.8. There is a big difference between that something is supported or that it's the default. And from my perspective, Greenshot is used on millions of systems and some are very old, the step to 4.7.2 will already break some peoples installation. I'm not sure if there is a NEED for Greenshot to upgrade, now if it would simplify the development I'm for it, but between 4.7.2 and 4.8 there are not that many changes which really bring much to the table. Do you know of any changes which are really important, and make it worth going towards 4.8? (it's a Windows Forms app, only uses WPF in a few use-cases)

What I would like is to go to is the latest .NET (current v6) for Greenshot 2, but providing this with the installations makes it more than 15x bigger. It would simplify the development, and provide new features / performance, there is a great benefit, but I still need to wage the pros against the cons.


About the code formatting, some really make sense. But based upon what did you set the .editorconfig? e.g. I like vars, and I do not like huge empty space on my screen (which a indention of 4 more or less does). I can adjust, but I want to know where the .editorconfig settings came from, if they are some defacto-standard which I didn't know of, we can certainly use them.
Also what is the added value of an empty documentation element, before I at least knew there was something missing.

That was all cosmetics, but what is more important: some of the code cleanups breaks the code. This might not be directly visible, but e.g. here I had some issues: https://github.com/greenshot/greenshot/pull/359/files#diff-23da9a5d03567ded83fbabe79aa2cfcfda5509abd85c74393fee061fbabc80b8L41
The new LogSource() seems not to be the same as new(), I think it has to do with the callermember attribute.

Also for some of our COM code, especially the Office Integration, Visual Studio suggests a code change, but doing so breaks the code. I didn't see the files of which I know they should be untouched, as there were that many changes, but if they are between the changes it most likely doesn't work. (yeah, such stuff is annoying)

btw. I have ReSharper (from JetBrains) and it can do much the same, I usually only do stuff in steps where I try not to bundle to many changes.

unquietwiki:
Thanks for the feedback on this @Lakritzator. Addressing your questions.

- With 4.8, you do get performance improvements & user-friendly changes in Forms & WPF. 4.8 code base is also tied to .NET Core 2.1, so that'll help with the .NET 6 migration.
- _We avoid `this.` unless absolutely necessary_ was in the contributor guidelines. I had another guy not happy with me for converting code to var; and I saw that more code, than not, was typed. Updating the .editorconfig to prefer var can reset this.
- https://github.com/kzu/roslynator/blob/master/.editorconfig has been a template for me on having a proper .editorconfig . Microsoft has a [bunch of documentation](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/code-style-rule-options) on the options, I found useful as a reference.
- You had me curious about the logging error. I did find https://docs.microsoft.com/en-us/dotnet/core/extensions/logging & https://badecho.com/index.php/2021/01/26/net-library-log/ ; I guess there's a thing in .NET 5.0 that creates a new logging infrastructure; it appears to be installable as an NPKG for 4.6 & up. log4net apparently doesn't support newer .NET frameworks; changelog says 4.7.2 at best, but their other matrix is out-of-date. https://logging.apache.org/log4net/release/framework-support.html

Maybe the more prudent thing here would be to take what I have, make another branch, and play around with it?

Lakritzator:
The log issue is not that what you mean, I have added some of my Dapplo branded code in Greenshot and this uses "LogSource". Using LogSource logger = new LogSource() would use the constructor here: https://github.com/dapplo/Dapplo.Log/blob/master/src/Dapplo.Log/LogSource.cs#L68
In that case the CallerFilePath will make sure that the logsource knows what is the source of the log entries.
Somehow having a "new()" (C# 9 Type Inference for the New Keyword) **seems** to break this, I am not sure yet why I can't find anyone else having the issue.

I do not like to force people to 4.8 in the current time frame for 1.3, so I will not merge this.
What is important to know, I was very hesitant to say no to this PR, as I really enjoy work coming from the community and I do not take it lightly.

I would advice, in fact anyone creating a PR which costs work, to check with maintainers if this is a think they would like. It's hard to say no, and than it just lies around 🤷 

I do like the unused iss changes, and I do think there is more room for improvement for the .editorconfig and the formatting!!! So yeah, maybe you can take some of it, into a new PR. No moving to 4.8 yet please! 😉

savornicesei:
@unquietwiki 
In regards to log4net, the latest NuGet package supports netstandard20 which makes it available to .NET framework projects, .NET core and .NET 5/6 projects.
image

unquietwiki:
@savornicesei Acknowledged. I can backtrack & retarget per earlier in the discussion, and see if that meets your needs. I'm still dogfooding the build I made two weeks ago; no issues, but I'm not using logging or Office integration, as you were concerned about.

You mentioned before an interest in going to .NET 6 at some point in the future. I found [this article](https://dotnetcoretutorials.com/2021/11/10/single-file-apps-in-net-6/); I heard this was doable in the past, and it may be something you want to explore.

unquietwiki:
@savornicesei Hey, sorry for the delay on this. I re-implemented the linter set & tried to do the easy fixes. I gather from the source code that you use ReSharper, and that might do some stuff VS Studio doesn't. There's a **lot** more to do; I can tackle some more as able, else you should be able to carry on as able. Thanks for your patience.

Issue #348, opened 2021-11-29
Desktop refreshes every second after Screenshot to Clipboard was taken, Win11 (22504) and Greenshot 1.3.219


schuettecarsten:
On Windows 11 Insider (latest release, 22504.1010), Greenshot 1.3.219 causes endless desktop refreshes after "Print" key was used to capture a screenshot of a selected area and copying this to the clipboard. This onyl happens when the capture is copied to the clipboard, using file output works as expected. After taking some screenshots, the tray icons start to flicker and refresh every second, the same happens to the desktop and open windows. Computer gets slower and slower on each screenshot you take...

badabing2005:
This happens to me too, but not always, I can't pin to what conditions actually trigger it (at times it is very frequent, other times it doesn't happen for days), but I know that the following process 
<img width="638" alt="image" src="https://user-images.githubusercontent.com/22603312/144617863-f32c0bd4-faa4-4eac-9baa-44de3b19f9b4.png">
starts spawning endless child processes that end up dying.
If you kill the parent wslhost process, the flickering stops.

I also experience another issue, although am not sure if it is related to windows 11 issues or 4K monitor issues.
Only sometimes, when I take a region screenshot, and select top left to bottom right, the captured region is not accurate, the top right is clipped, the only way around it when this happens is to select bottom left to top right.


Hopefully this will help identifying the root issue.

schuettecarsten:
@badabing2005 Please use Feedback Hub in Windows 11 to report this issue there. Maybe you want to add your feedback to https://aka.ms/AAd4bd8.

badabing2005:
@schuettecarsten 
Sorry Facebook is totally blocked at our residence.
Here's a screenshot of the started / killed processes when the problem happens.
![image](https://user-images.githubusercontent.com/22603312/144931699-1cc20fbf-b22f-48cb-8a8d-cf7bfbdf45b8.png)

Perhaps someone can put a link-back Facebook Hub to this thread.

schuettecarsten:
@badabing2005 Facebook? I mean the "Feedback" tool that is integrated in Windows 11.

But this here is not a Greenshot issue, as it also occurs when using Paint or other tools to copy data into the clipboard. It seems to be caused by a combination of WSL 2, Windows 11 and Docker Desktop.

pfeigl:
[...]

PR #371, opened 2022-02-03
Optimize DibHelper a bit, now only needs one byte array instead of two (#348)

schuettecarsten:
I have optimized the new DibHelper implementation a bit to get rid of one of the two byte array that are created during the generation of the DIB. Also, dassed try/finally to make sure that source bitmap is always unlocked.

### DIFF
--- a/src/Greenshot.Base/Core/DibHelper.cs
+++ b/src/Greenshot.Base/Core/DibHelper.cs
@@ -23,9 +23,12 @@ namespace Greenshot.Base.Core
         /// <returns>The image converted to DIB, in bytes.</returns>
         public static byte[] ConvertToDib(Image image)
         {
-            byte[] bm32bData;
+            byte[] fullImage;
+            int dataLength;
             int width = image.Width;
             int height = image.Height;
+            const int hdrSize = 0x28;
+
             // Ensure image is 32bppARGB by painting it on a new 32bppARGB image.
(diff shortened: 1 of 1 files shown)


schuettecarsten:
@Lakritzator - Ah, just saw that you've also optimized the DibHelper yesterday.

Lakritzator:
Ah yes, "sorry", I waited with the optimization until it was proven that it solved the issues.
Thanks a lot for this, I guess you understand I will not merge this, so also closing this.

My optimization will not make a new bitmap if this is not needed (although Greenshot would probably use 24bpp 90% of the time), and was also done with the future in mind, I actually would have liked to directly write to the MemoryStream that is needed but unfortunately this is first available with .NET 5 and upwards (I do not count .NET Core 3.1 a valid target for Greenshot)

I also update this just now with a bit better documentation, it's already hard enough to understand why / what when talking about DIBs etc.

P.S.
Another "optimization" would be the next step where I introduce a new IFileFormatHandler throughout Greenshot, making it easier to implement different formats in a modular way. What I mean with optimization in this case, is that having the support of different formats (for the clipboard) is very spread and hard to re-use. Dib is just another file format, so it will make sense to move the DibHelper to a DibFileFormatHandler, As soon as I do that, the logic to create the clipboard is very standardized for different formats. It does not make it faster though, but at least better to maintain.

Lakritzator:
Closing this, as I already did the work shortly before this PR.

PR #380, opened 2022-02-16
Fix/379 preferences window

Christian-Schulz:
I enlarged the SettingsForm a bit and rearranged the labels and controls on every tab. 

I tested it on Windows 21H1 with different languages and different DPI. Now it looks a bit better and all texts are readable.

Plus a little fix that toggle the expert setting labels like the other controls on this tab.

Examples


Lakritzator:
I keep having issues with most windows, not only this, on my multi-screen test setup.
It's driving me crazy, I managed to have some very good code which resized the window, but the font... it just doesn't want to scale.

My setup is one screen on 100% and one on 150%, do you have any issues with that?

PR #382, opened 2022-02-18
prevent negative fontsize

Christian-Schulz:
Fix for Bug while rotating screenshot with counters.

Old:
Rotate_bug_1
Rotate_bug_2

New:
Rotate_bug_3


Lakritzator:
Nice, good find. Thanks for helping us!

Issue #249, opened 2020-09-02
"Crop out" feature request.


Alacritas1:
Cropping out can be used to extract table rows with the heading line included. Right now i have to take 2 screenshots, one of the row and one of the heading, and merge them at destination. If I could crop out the rows above the target row with a tool I could export a single file with the heading and the target row. Total image size could get smaller automatically i imagine.
But I don't know how helpful this would be to others. Thanks in advance.

kirkilj:
In the past, I used a tool that let you control this crop-out feature by supporting options to apply torn edges to the edges that were cropped out and to adjust the space injected between the cropped regions.

tomgwilliams:
Agreed! This is one of the biggest features I'm missing from Greenshot compared to SnagIt. SnagIt has the Vertical or Horizontal cut out tool, which I use very very often. You basically set this up so that a user drags a full width/Ht Horizonal or Vertical region from start to end, and you splice the image into three different parts (if they run to the end of the image one of the 3 parts is just empty), deleting the center area, and shifting the two remaining splices together. 

As a follow on, it would be great to add a blurred, torn, etc. to the splice seams, but just a straight line will meet 95% of this ask. The other stuff is absolutely optimization.

mischlrebl:
just a small side note: if the source is Excel, you could also just select the cells, copy them and paste it into GS.
would need two steps again, but should be a little bit faster.

uwbwsvd:
I often need parts of the screen. The top and the bottom for example. I have to cut out the rest and shrink so the picture.

Christian-Schulz:
> I often need parts of the screen. The top and the bottom for example. I have to cut out the rest and shrink so the picture.

This feature was already integrated with #388 in v1.3.244 , only this issue has not been closed.

<img width="854" height="1100" alt="Image" src="https://github.com/user-attachments/assets/552f476c-94ae-47b8-bfce-c797fb7fd96f" />

uwbwsvd:
Ah. Thanks!
Is it possible to create a cutout with frayed edges?
So that you can tell something has been cut out there?

Christian-Schulz:
> Is it possible to create a cutout with frayed edges?

Currently not. There is an open feature request #935 for that.

Issue #376, opened 2022-02-08
Being able to choose the number of digits for ${NUM} Filename pattern


jupiter104:
**Describe the bug**
If you use **${NUM}** as the **Filename pattern** it automatically outputs an incrementing 6 digit number but what if you want an incrementing 4 digit number or a 7 digit number?

**To Reproduce**
Example steps to reproduce the behavior:
1. In the **Output** tab, make sure your **Filename pattern** is **${NUM}** 
2. In the **Destination** tab, make sure you have checked **Save directly**
3. Take multiple screenshots.
4. You will see the file names of the screenshots are six digit numbers.

**Expected behavior**
It would be better if Greenshot had a filename pattern option to choose the number of digits. For example, ${NUM3} would output incrementing 3 digit numbers starting from 001, then 002 and so on. {$NUM7} would output 7 digit numbers starting from 0000001.

**Versions (please complete the following information):**
 - Greenshot 1.2.10 Build 6 (64 Bit)
 - Windows 10 Pro (Version: 21H2, OS Build: 19044.1503)

**Additional context**
I've also noticed if you use ${title} as the filename pattern, it can output very long filenames if the title of the window was very long. It would be better if we could limit the number of characters like ${title10} which would only use the first 10 characters to name the screenshot file.

Lakritzator:
We've got you covered, wait for my answer later... can't write right now, maybe in ~12 hours.

Lakritzator:
So, for the formatting you have the following options:

`${x:yz}` where:


1.  "x" the placeholder, we support
  - domain
  - user
  - hostname
  - YYYY - current year
  - MM - current month
  - DD - current day
  - hh - current hour
  - mm - current minutes
  - ss - current seconds
  - now - datetime of now
  - capturetime - time of the capture
  - NUM - incrementing number
  - title - title of the window which was captured
  - MyPictures - special folder for your pictures
  - MyMusic - special folder for your music (don't ask...)
  - MyDocuments - special folder for your documents
  - Personal - special folder for your personal files
  - ApplicationData - special folder for your application data (roaming)
  - Desktop - special folder for your desktop
  - LocalApplicationData - special folder for your local application data
 
2. y is the formatter, and z the parameters for that formatter we accept
 - p for padding, with width,padchar (e.g. ${NUM:p10,0} for a 10 digit NUM with padding 0s)
[...]

PR #388, opened 2022-02-24
Enhanced ability to crop an image vertically and horizontally. #249

Christian-Schulz:
I extended the cropping mode.
Now you can crop the image horizontally and vertically.

The CropContainer has three Styles `DefaultCropStyle`, `HorizontalCropOutStyle` and `VerticalCropOutStyle`.

You can change the current style by clicking the crop button (C) again. 

The first click on the crop button (C) activates the last style the second click activates the next.

HorizontalVerticalCrop

`DefaultCropStyle` --> `HorizontalCropOutStyle` --> `VerticalCropOutStyle` --> `DefaultCropStyle` and so on.

I chose this mechanism so that I don't have to add a toolStripItem for the crop style. 
That would have required a translation. 

The PR/MR covers the following features:

[FEATURE-539](https://greenshot.atlassian.net/browse/FEATURE-539) "torn image in the middle"
[FEATURE-805](https://greenshot.atlassian.net/browse/FEATURE-805) "Increase Crop functionality"
[FEATURE-873](https://greenshot.atlassian.net/browse/FEATURE-873) "Bearbeitungsfunktion: Streifen ausschneiden"
[FEATURE-900](https://greenshot.atlassian.net/browse/FEATURE-900) "Cut-Out cropping"

Kown issue: sometimes I get a NullReferenceException when undoing (I am testing it some more)


mischlrebl:
very impressing and cool - congratulations!

would it not be easier for the handling to only have one style (with the 8 handles)? this way all the possibilities are available and the tool works all the time the same way so everyone would understand it immedeately.

another point: the last missing option I miss in GS is the ability to crop out a section like in paint. means the selection gets blank on the image whereby the selection would be available as new own image in a separate layer. some of the linked features describes exactly that need. probably a third icon on the toolbar could adress it: Confirm - Cancel - Crop out, or right button click (menu). 

thanks a lot again for the great work - and to take these input into consideration eventually.

Lakritzator:
This is great work, and a huge improvement for Greenshot.
I only had a shallow look at the code, and didn't notice any real discrepancies on our coding style, and the part with the NullReferenceException I will have a look at when we I have a bit more time for this.

The only issue I have with it, is the usability. If we keep it like it is right now, people will only stumble on the feature by accident.
I'm not 100% sure how @mischlrebl means it with keeping the the 8 "handles" (crop adorners). But I agree that we need to find a solution for this. Maybe it would make sense to have more buttons where the ok (check) and cancel (X) are to select the mode? I would also like to see the "auto crop", which is currently a bit hidden in the menu, be more present.

@mischlrebl I agree that a crop out is also nice, and in general it's the region inverse of what is selected but what I am somehow missing: what is left at the place you crop out? Greenshot doesn't have a concept like a background color... For transparent images this is understandable, and we COULD "upgrade" the image to have transparency when this is used, but if that is what we are looking for?

mischlrebl:
sorry about the confusion about the 8 handles. I made a wrong assumption without trying it. 
at the end a fourth mode would be nice, the selection of the first one with the ability to cut it out and paste in at once.

@Lakritzator paint uses always white as background color as it looks. GS uses transparent when expanding the area with "Crtl Shift +"
btw, it would be nice to describe the Auto Crop feature in the menu with its shortcuth "Ctrl Shift -"

Christian-Schulz:
I added options for crop mode.
Now next step looks like this.

HorizontalVerticalCrop_v2

I still have ideas for two more options. Crop fixed size and fixed ratio.

@mischlrebl GS has a feature for expanding the image!?  I always took a screenshot twice as workaround for this.

mischlrebl:
@Christian-Schulz yes, Ctrl-Shift-+ increases the image on each side by 25px

Lakritzator:
It's a weird feeling when contributors can answer quicker than me 😉 
I was putting my kid into bed, and saw the question about the expanding the image...

I believe we are now reaching the phase in an application where we have more features than are visible at a simple glance, and we have no good way of promoting them. I know quite a few which not everybody knows off, and some are even hidden in the .ini configuration. If I look at ShareX, which is also a great screenshot tool, I do not want to go into the direction of that many options and make it harder for the user.

Example of features some people do not know of:

- Drag and drop of image files from the file system, OR even from some websites. These are placed upon the surface, where you can move and resize them. This also works with copy & paste.
- You can also capture a part of the screen and insert it directly into an open editor
- You can insert the contents of a window, directly from the editor
- There is a lot of things you can tweak for the filename generation, as is documented only in #376

Back on topic:
That looks amazing, where did you get the icons?

While looking at your demonstration I realized that:
We might need an auto-crop which scans away from the current crop region to the inside, removing a step which you were doing. Solving the "I need to have only this area inside my capture. This might also be something for the region selection.

Continuing on the other mentioned features, fixed size and fixed ratio, this sounds good and is also something which the region capture is missing. In fact, some people were requesting a capture mode similar to that of the crop, with that I mean the way you move the selection rectangle. About the fixed ratio, we already should have some code for that, this was also needed in #358 (which is waiting for a stable version of ImageSharp.Drawing)

I'll try to have a look at it, but I like to finish some stuff I started. Maybe I first do a quick code review before I try it out...

P.S.
Your feature, at least the way it is, does cause one challenge... that is translations of the tooltips of the buttons. Maybe we can skip those for now, I do not see a huge loss if we do not provide a translation, compared to how it is now.

Lakritzator:
FYI, I'll quickly do a code review, sometimes I might sound blunt, if I do, I do not mean it like that! 😉 
I prefer to quickly point out something, usually it's something the contributor overlooked / likes to do differently or do not see the big picture, sometimes contributors are right and I don't see the light at the moment.

src/Greenshot.Editor/Drawing/CropContainer.cs
```diff
@@ -45,13 +70,70 @@ protected override void OnDeserialized(StreamingContext streamingContext)
...
+                        break;
+                    }
+                case CropMode.Vertical:
+                    {
+                        InitVerticalCropOutStyle();
+                        break;
+                    }
+                default:
+                    {
+                        InitCropStyle();
```
Lakritzator: Might it make sense to call the CreateDefaultAdorners directly, without InitCropStyle wrapping it? I couldn't find a second using of it.

src/Greenshot.Editor/Drawing/CropContainer.cs
```diff
@@ -45,13 +70,70 @@ protected override void OnDeserialized(StreamingContext streamingContext)
...
+        }
+
+        private void InitCropStyle()
         {
             CreateDefaultAdorners();
         }
 
+        private void InitHorizontalCropOutStyle()
+        {
+            var defaultHeight = 25;
```
Lakritzator: I'd prefer "const" here, as that is what it is... I do wonder if there is a need to make such values scalable like depending on the DPI.

src/Greenshot.Editor/Drawing/CropContainer.cs
```diff
@@ -45,13 +70,70 @@ protected override void OnDeserialized(StreamingContext streamingContext)
...
+            if (_parent?.Image is { } image)
+            {
+                Size = new Size(image.Width, defaultHeight);
+            }
+            CreateTopBottomAdorners();
+        }
+
+        private void InitVerticalCropOutStyle()
+        {
+            var defaultWidth = 25;
```
Lakritzator: Same here, const.

src/Greenshot.Editor/Drawing/CropContainer.cs
```diff
@@ -83,27 +165,108 @@ public override void Draw(Graphics g, RenderMode rm)
                 return;
             }
 
+
             using Brush cropBrush = new SolidBrush(Color.FromArgb(100, 150, 150, 100));
```
Lakritzator, line 165: Just a general comment: That WE never taken the color into a const, is actually weird... it feels like a magic number.

src/Greenshot.Editor/Drawing/Fields/FieldType.cs
```diff
@@ -31,31 +31,33 @@ namespace Greenshot.Editor.Drawing.Fields
...
-        public static readonly IFieldType LINE_COLOR = new FieldType("LINE_COLOR");
-        public static readonly IFieldType LINE_THICKNESS = new FieldType("LINE_THICKNESS");
-        public static readonly IFieldType MAGNIFICATION_FACTOR = new FieldType("MAGNIFICATION_FACTOR");
-        public static readonly IFieldType PIXEL_SIZE = new FieldType("PIXEL_SIZE");
-        public static readonly IFieldType PREVIEW_QUALITY = new FieldType("PREVIEW_QUALITY");
-        public static readonly IFieldType SHADOW = new FieldType("SHADOW");
-        public static readonly IFieldType PREPARED_FILTER_OBFUSCATE = new FieldType("PREPARED_FILTER_OBFUSCATE");
-        public static readonly IFieldType PREPARED_FILTER_HIGHLIGHT = new FieldType("PREPARED_FILTER_HIGHLIGHT");
-        public static readonly IFieldType FLAGS = new FieldType("FLAGS");
+        public static readonly IFieldType ARROWHEADS = new FieldType(nameof(ARROWHEADS));
```
Lakritzator, line 34: Nice! 👍 
I first thought I was blind and didn't see why the next likes were changes, until I noticed the nameof... that makes sense, when this code was written there was no nameof.

src/Greenshot.Editor/Drawing/Surface.cs
```diff
@@ -1226,6 +1226,112 @@ public bool ApplyCrop(Rectangle cropRectangle)
...
 
+        /// <summary>
+        /// Crop out the surface
+        /// Splits the image in 3 parts(top, middle, bottom). Crop out the middle and joins top and bottom. 
+        /// </summary>
+        /// <param name="cropRectangle">rectangle of the middle part</param>
+        /// <returns></returns>
+        public bool ApplyHorizontalCrop(Rectangle cropRectangle)
+        {
+            if (IsCropPossible(ref cropRectangle))
```
Lakritzator: Can you please change this to an invert of the if, so using a fail fast and thus have less indention in the rest of the method? (if !IsCropPossible -> return false)

src/Greenshot.Editor/Drawing/Surface.cs
```diff
@@ -1226,6 +1226,112 @@ public bool ApplyCrop(Rectangle cropRectangle)
...
+            {
+                Rectangle imageRectangle = new Rectangle(Point.Empty, Image.Size);
+                Bitmap tmpNewimage, tmpImageTop, tmpImageBottom;
+                // Make sure we have information, this this fails
+                try
+                {
+                    tmpNewimage = new Bitmap(Image.Size.Width, Image.Size.Height - cropRectangle.Height);
+                    tmpImageTop = ImageHelper.CloneArea(Image, new Rectangle(0, 0, Image.Size.Width, cropRectangle.Top), PixelFormat.DontCare);
+                    tmpImageBottom = ImageHelper.CloneArea(Image, new Rectangle(0, cropRectangle.Top + cropRectangle.Height, Image.Size.Width, Image.Size.Height - cropRectangle.Top - cropRectangle.Height), PixelFormat.DontCare);
```
Lakritzator: please make sure these 2 bitmaps (tmpImageTop & tmpImageBottom) are all disposed, otherwise we have a memory leak and also get GDI errors after a while.

src/Greenshot.Editor/Drawing/Surface.cs
```diff
@@ -1226,6 +1226,112 @@ public bool ApplyCrop(Rectangle cropRectangle)
...
+            {
+                Rectangle imageRectangle = new Rectangle(Point.Empty, Image.Size);
+                Bitmap tmpNewimage, tmpImageLeft, tmpImageRight;
+                // Make sure we have information, this this fails
+                try
+                {
+                    tmpNewimage = new Bitmap(Image.Size.Width - cropRectangle.Width, Image.Size.Height);
+
+                    tmpImageLeft = ImageHelper.CloneArea(Image, new Rectangle(0, 0, cropRectangle.Left, Image.Size.Height), PixelFormat.DontCare);
+                    tmpImageRight = ImageHelper.CloneArea(Image, new Rectangle(cropRectangle.Left + cropRectangle.Width, 0, Image.Size.Width - cropRectangle.Width - cropRectangle.Left, Image.Size.Height), PixelFormat.DontCare);
```
Lakritzator: Same as the other method: please make sure these 2 bitmaps (tmpImageTop & tmpImageBottom) are all disposed, otherwise we have a memory leak and also get GDI errors after a while.

src/Greenshot.Editor/Drawing/Surface.cs
```diff
@@ -1226,6 +1226,112 @@ public bool ApplyCrop(Rectangle cropRectangle)
...
+
+        /// <summary>
+        /// Crop out the surface
+        /// Splits the image in 3 parts(left, middle, right). Crop out the middle and joins top and bottom.
+        /// </summary>
+        /// <param name="cropRectangle">rectangle of the middle part</param>
+        /// <returns></returns>
+        public bool ApplyVerticalCrop(Rectangle cropRectangle)
+        {
+            if (IsCropPossible(ref cropRectangle))
```
Lakritzator: Same as the other method, invert the if.

src/Greenshot.Editor/Drawing/Surface.cs
```diff
@@ -1971,6 +2101,15 @@ public void ConfirmSelectedConfirmableElements(bool confirm)
             }
         }
 
+        public void RemoveCropContainer()
+        {
+            if (_cropContainer != null)
```
Lakritzator: I prefer an `if (_cropContainer == null) return;` and have the other code without indention.

src/Greenshot.Editor/Forms/ImageEditorForm.Designer.cs
```diff
@@ -143,10 +141,15 @@ protected override void Dispose(bool disposing)
 			this.toolStripSeparator16 = new System.Windows.Forms.ToolStripSeparator();
 			this.btnHelp = new GreenshotToolStripButton();
 			this.propertiesToolStrip = new ToolStripEx();
-			this.obfuscateModeButton = new BindableToolStripDropDownButton();
-			this.pixelizeToolStripMenuItem = new GreenshotToolStripMenuItem();
+            this.obfuscateModeButton = new BindableToolStripDropDownButton();
+            this.cropModeButton = new BindableToolStripDropDownButton();
```
Lakritzator: There is a lot of indention issues with this file, I can see that this is not your fault, but maybe you can clean it up?
What I do like, and what we unfortunately do not communicate, we do NOT use the designer, as the code might change so much that we can no longer track what changed. You also didn't do so, at least that is what it looks like, and only committed what was changed.

src/Greenshot.Editor/Forms/ImageEditorForm.cs
```diff
@@ -727,8 +728,12 @@ private void BtnArrowClick(object sender, EventArgs e)
 
         private void BtnCropClick(object sender, EventArgs e)
         {
-            _surface.DrawingMode = DrawingModes.Crop;
-            RefreshFieldControls();
+            if (_surface.DrawingMode != DrawingModes.Crop)
```
Lakritzator: also here, invert if.

src/Greenshot.Editor/Forms/ImageEditorForm.cs
```diff
@@ -1292,6 +1297,7 @@ private void BindFieldControls()
             new BidirectionalBinding(previewQualityUpDown, "Value", _surface.FieldAggregator.GetField(FieldType.PREVIEW_QUALITY), "Value",
                 DecimalDoublePercentageConverter.GetInstance(), NotNullValidator.GetInstance());
             new BidirectionalBinding(obfuscateModeButton, "SelectedTag", _surface.FieldAggregator.GetField(FieldType.PREPARED_FILTER_OBFUSCATE), "Value");
+            new BidirectionalBinding(cropModeButton, "SelectedTag", _surface.FieldAggregator.GetField(FieldType.CROPMODE), "Value");
```
Lakritzator, line 1294: Note: just saying, as it's not your addition I can express my horror: the BidirectionalBinding thingies are horrible, I cannot stand a new without assigning them... but I didn't write that part, it's grown hysterically. (historically)

Christian-Schulz:
I have taken all your hints.  Tabs and spaces are mixed in some files, especially in designer files and translation files. I have rewritten this so that only my changes are committed.

The icons are drawn by me. I reused parts form existing icons from editor resources.

I would say the feature is complete, but i still have the NullReferenceException when undoing.
I can reproduce that in release/1.3. 
Steps:
* activate crop 
* select a 0x0 Pixel rectangle
* confirm
* click undo

selecting a 0x0 pixel rectangle is difficult, but in the new horizontal/vertical mode it is easier. I get the same exception when I select the whole image.

Lakritzator:
Great to hear the progress!

The NRE itself seems to come from the fact that the crop container itself is placed on the undo stack, in this case the size changes to it, which do not make sense. Now somehow, besides that a CropContainer change shouldn't be on the undo stack, the crop container doesn't seem to be on a surface (parent is null). I need to check why this happens, and if this generally the case or only in the odd case where you crop with 0x0.

Back to how you reproduced the NPE, this is something which should be prevented:
Both a crop of 0x0 or a crop of the whole image, would not make sense to apply, so there should be a check of the crop size before it's applied. This means that in this case, do not apply a change when the user presses enter. Also the button should be disabled so it's visible that it's not possible.

Lakritzator:
I've pushed some code changes with logic which makes sure that the CropContainer never ends on on the undo/redo stack, it is in fact a miss-use of the DrawableContainer, as in the end it's only there to support the crop action, and not something which ends on the end result. The undo/redo is managed by the SurfaceBackgroundChangedMemento, which takes care of changes in the size of the capture.

This should solve the issue, but I didn't have enough time yet to test everything through...

Lakritzator:
The previous commit is not needed, but if I already touch files I rather remove some code smells.

What I actually came here for, was to mention that the creating of the resulting image in both ApplyHorizontalCrop and  ApplyVerticalCrop could be a lot more efficient by just using the [DrawImage](https://docs.microsoft.com/en-us/dotnet/api/system.drawing.graphics.drawimage?view=dotnet-plat-ext-6.0#system-drawing-graphics-drawimage(system-drawing-image-system-drawing-rectangle-system-drawing-rectangle-system-drawing-graphicsunit)) copying from the source to the new to be formed image, without actually cloning the areas first. That saves a huge amount of copying of pixels:

* Create new image for the result
* Clone area 1 (pixels are copied)
* Clone area 2 (pixels are copied)
* Draw area 1 to new image (same pixels copied again)
* Dispose area 1
* Draw area 2 to new image (same pixels copied again)
* Dispose area 2

Could be:
* Create new image for the result
* DrawImage area 1 to new image
* DrawImage area 2 to new image

This will also reduce the (GDI) memory usage.

Lakritzator:
I already made the change I suggested, I hope you are ok with that.

Lakritzator:
I found some small inconsistencies while starting some tests, and I'm not sure if they come from this PR:

- After crop I sometimes do not see an update on the undo button, so it looks like it's not possible to undo
- If I press the crop button, and get the checkbox / cancel button, and press the cancel button, all buttons seem to be disabled.

What I didn't look at yet, is trying to use the feature when using the zoom...

mischlrebl:
If there would be a build, I could offer some testing

Christian-Schulz:
I think a rework for confirm/cancel is needed.
The check/cancel buttons are displayed because the CropContainer exists. The confirm/cancel action only works when the CropContainer is selected.  So you cannot cancel the crop mode before you have selected something.

Christian-Schulz:
I would say its ready for review/merge.
Tested with toolbar buttons and keyboard shortcuts.
Tested with and without zoom.
Tested all crop modes with pixel perfect reference image.
Confirm/cancel works.
Undo/redo works.

Lakritzator:
Looks very nice!

Here are some small things I want to discuss:

- I'm not sure if we can provide the translations for the new crop modes, there are many people to contact for that and we do not have dedicated translators. I wonder if we can live without any texts on the buttons...
- If I press the "check" button before selecting anything, it cancels like I pressed ESC. I would expect the check button to be disabled until something is selected.
- Did you ever think about having a conversion from "normal" crop selection to horizontal or vertical? Changing the crop mode, feels a little wasteful... "oh no, I already selected it, now it's gone". So going from a rectangle to only the horizontal or vertical selection. Maybe we can even keep the crop information hidden, so switching back from horizontal or vertical to normal might bring back the selection which was hidden. It might be a bit hard to explain, but before I go into details, maybe you already know what I mean ;)
- What do you think of improving autocrop? e.g. if we already selected a crop region, moving towards autocrop can scan inwards from that selection (versus going inwards from the edges), and provide a new automated selection. Can we do a vertical or horizontal auto crop, might that make sense?

Christian-Schulz:
* I have only translated English and German completely. In all other languages I used the existing translations from the individual files, i.e. in `editor_cropmode_default` i only cut off the letter "C" for the shortcut.

* Correct, as long as you have not selected anything, Cancel and check are the same thing. Currently, confirm/cancel are independent of the container type. They are dependent on an existing FieldFlag `CONFIRMABLE`. An additional dependency on `Surface.HasSelectedElements` is reasonable and should be possible. 

* Do you mean that the rectangles of one crop mode are used to initialise (preselect) the next mode. I think the result is not usable in every case. Changing from vertical to horizontal results in a selection of the whole image cause they are bind to the edge.
Or do you mean that the last rectangle for each crop mode should be saved? If you change the crop mode and go back the last rectangle is used. Where is this information stored? In the current editor instance (Surface), or in a FieldValue (ini) ? 

* I personally never used the auto crop function. So i don't know if incremental auto crop is a use case. To implement this, we have to add an extra "automated selection" button. I think it would be possible, also for vertical and horizontal, but is it really a use case?

Lakritzator:
- I really need to think a bit about what we do with the translations... @jklingen we have an awesome feature here, but some buttons need a label, meaning it would need translations. I'm not sure if we could do without, the images on the button might be enough to understand the feature. My suggestion would be to merge the feature, and leave out the translations (unless we expect that we can managed adding them). Even if it would not be intuitive enough, people will only miss out on an additional feature.

- I would say, it makes sense to disable the confirm button when there is nothing to confirm. Having a button which does the same as the button next to it, just confuses the users.

- I could imagine that we keep a real rectangle, like with the normal crop, and if the user uses vertical crop-mode we only display the vertical lines and apply the vertical selection when cropping, if the users uses the horizontal crop-mode we only display the horizontal lines and apple the horizontal selection when cropping. This we will only do when we move from a normal to one of the other modes, and we can go back. Anyway, this is a bit more complex and I do not need it in this PR, it's just an idea. 

- I noticed in your screen recording that you cropped something, to autocrop it afterwards, which pretty much showed what I see more often and also do myself... I see a need here. I'll have a look if I can make an adjustment to support this use-case.

P.S.
While thinking more about it, due to this PR, the crop-container is more of a functionality of the editor as the surface... so we really are abusing the surface.

Christian-Schulz:
Now the confirm button is only enabled when there is a selection.
I have added a hatch pattern to the selection rectangle.

Now it looks like this.
Crop_Hatch_Pattern

Lakritzator:
Yeah, that with the button is great!

I'm not sure about the pattern though, I think I need to see it for real.

P.S.
I thought I was late in bed, but you seem to have won last night...

Lakritzator:
I've added the auto crop change I wanted to see. So before autocrop is called, I take the bounds of the current crop container and pass this along so the routine can look inside that rectangle to a place to auto crop.

This also means I saw the pattern, and I am not sure if I like it.

Christian-Schulz:
Yes, I don't like the hatch pattern. The normal user will not know what it is supposed to mean.

What do you think of this idea with arrows?
Selection_with_arrows
I have tested a bit. It is not ready yet.

Lakritzator:
Just leave it as it was (without pattern) for now. We can always look at that later, but I see many challenges when you start bringing arrows in (what happens if the area is very small, etc).

jschof:
Christian, this is amazing! Your contribution will make GS truly great and reduce the need to use other tools to next to nil. Thank you (and Robin) for all the time you've dedicated to this, know that it's greatly appreciated.

Christian-Schulz:
I have removed the hatch pattern and tested the recursive auto crop.

repeat_auto_crop

There was a little bug. Every DropDownItemClicked Event was subscribed more than once, so the auto crop was triggered too often.

src/Greenshot.Editor/Forms/ImageEditorForm.cs
```diff
@@ -280,11 +284,6 @@ private void UpdateUi()
             // a smaller size than the initial panel size (as set by the forms designer)
             panel1.Height = 10;
 
-            fontFamilyComboBox.PropertyChanged += FontPropertyChanged;
```
Lakritzator, line 283: I must be blind, but where did these 3 lines go?? Did I remove them, don't recall? I do understand that this is bad, register events in an update ui doesn't sound good, but where did they go?

Lakritzator: I found that you moved these to the Designer file, that makes sense.

Lakritzator:
I will merge this, although we still have some open todos with the translations the feature itself is fully finished and awesome ❤️ 
@Christian-Schulz Thanks for the GREAT work, while I am working on issues and documenting I unfortunately do not have time to work on many features (which is sad, I love that the most).

If needed we can blend out the feature, or we see if the community steps up to translate this soon. But I do not want to risk that I make changes which cause a merge conflict and thus need me to go through the code fixing that...

Lakritzator:
First build with this feature is: https://github.com/greenshot/greenshot/releases/tag/v1.3.244

PR #391, opened 2022-03-21
Create language-website-ru-RU.xml

ViktorOn:


ViktorOn:
In fact, you are the first project that asks me for this. Sorry, but I don’t want to go anywhere else through my Github account.
If it helps, then I hereby give my consent to the use of this contribution in this project.

PR #398, opened 2022-04-08
Refactoring to use Dapplo.Windows

Lakritzator:
I rewrote a lot of the interop code for Greenshot in separate libraries under the project Dapplo.Windows.
The biggest advantage of this is that it's easier to test, and continue development.
A disadvantage is that a fix needs to be made there first, and this costs a bit more time.

This PR is quite big, a lot of changes, so there is a risk that something works differently and might cause bugs.


Lakritzator:
Some background about this PR:

I've been working on this for was longer than a year, to make usable and more modern APIs for most of the "Win32" code and also made most of them testable. In fact I already used this in a branch which was supposedly going to be Greenshot 2, but stopped due to some .NET issues.

More on the MIT licensed open source project can be found here: https://github.com/dapplo/Dapplo.Windows

Some planned (and thus available) parts have not been integrated yet:

* The whole window information system
  * Currently the information on what windows are where is still created by logic inside Greenshot using some of the APIs in Dapplo.Windows 
* Clipboard code
  * including the possibility to have delayed "rendering" and thus reducing memory)
  * Automatic notification of information on the clipboard, making additional features possible (like reacting to a Firefox capture)
* .NET 6+ compatible & better performance

Issue #397, opened 2022-04-07
Pressing Escape after Rectangle tool throwing Unexpected error 


easyrepro:
**Describe the bug**
Pressing Escape after Rectangle tool throwing Unexpected error 

**To Reproduce**
Example steps to reproduce the behavior:
1. Capture '....'
2. Open Editor
3. Add rectangle
4. Hit Escape in keyboard

Error: 
```
Software version: 1.3.249 - 9ae5598ab8 (64 bit)
.NET runtime version: 4.0.30319.42000+
Time: 2022-04-07 15:59:54 -07:00
OS: Windows 10 2009 x64 build 19044
GDI object count: 136
User object count: 71

Exception: System.NotImplementedException
Message: No confirm/cancel defined for Container type Greenshot.Editor.Drawing.RectangleContainer

Stack:
   at Greenshot.Editor.Drawing.Surface.ConfirmSelectedConfirmableElements(Boolean confirm) in D:\a\1\s\src\Greenshot.Editor\Drawing\Surface.cs:line 2109
   at Greenshot.Editor.Drawing.Surface.ProcessCmdKey(Keys k) in D:\a\1\s\src\Greenshot.Editor\Drawing\Surface.cs:line 2523
   at Greenshot.Editor.Forms.ImageEditorForm.ProcessCmdKey(Message& msg, Keys keys) in D:\a\1\s\src\Greenshot.Editor\Forms\ImageEditorForm.cs:line 1167
   at System.Windows.Forms.Control.ProcessCmdKey(Message& msg, Keys keyData)
   at System.Windows.Forms.ContainerControl.ProcessCmdKey(Message& msg, Keys keyData)
   at System.Windows.Forms.Control.PreProcessMessage(Message& msg)
   at System.Windows.Forms.Control.PreProcessControlMessageInternal(Control target, Message& msg)
   at System.Windows.Forms.Application.ThreadContext.PreTranslateMessage(MSG& msg)
```

Configuration dump:

Lakritzator:
Should be fixed with the latest unstables.

PR #399, opened 2022-04-11
IsConfirmable for IDrawableContainer

Christian-Schulz:
Property IsConfirmable added for IDrawableContainer to indicate whether confirm/cancel is available.
To avoid System.NotImplementedException when pressing esc/enter on a container that does not support it.

Solve #397


Lakritzator:
Thanks, yesterday I had a look at the other issue, and was very pleased to see that you already had this on the radar 👍

PR #406, opened 2022-05-16
Update language-sv-SE.xml

dotar:


Lakritzator:
Thanks for the update, please do sign the CLA, otherwise we cannot merge this.

If you do not understand the CLA, which is OK... Summary:

1. With the CLA you give us the copyright, thus allow us to take your change(s) and use/modify/publish it. You also get the copyright for that change back, so you can still do whatever you want with the "knowledge" and code of that change (and only that change).
2. There is a two way disclaimer, so you provide the change "as is" and give no warranty, which means we can not come back to  you and "sue" you if something doesn't work. This means you are not liable if something goes wrong, this also goes both ways.
3. You are also automatically a contributor, but can if you want, place your name into the translation comments.

This prevents some legal discussions in both directions.
You should read it though...

jklingen:
Thanks a lot 👍

Issue #404, opened 2022-05-09
Crop does not work anymore as before


mischlrebl:
**Describe the bug**
the crop tool in the editor does not show the selection rectangle immedeately, only after

**To Reproduce**
Example steps to reproduce the behavior:
1. Capture '....'
2. Open Editor
3. Press 'C'
4. Select an area
5. The rectangle is not there anymore, only after releasing the mouse key

**Expected behavior**
while selecting the area, the selection rectangle should be shown

**Additional context**
btw, wasn't there an auto-crop item before in den edit menu?

Ishmaeel:
Can confirm. Area selection by dragging no longer works. A fixed-size selection rectangle is created upon mouse release which can then be resized.

@mischlrebl, Auto-crop item is now hidden inside the crop type selection dropdown in the toolbar button.

mischlrebl:
you are right with the crop tool, thank you
btw, wasn't there a shortcut too? I can't remember...

Ishmaeel:
No, there wasn't a shortcut for auto crop.

Lakritzator:
Please check the build (1.3.258) with the fix from @Christian-Schulz from here: https://github.com/greenshot/greenshot/releases
I'm currently still down with COVID...

Ishmaeel:
@Lakritzator nicely fixed, thank you, and get well soon.

Lakritzator:
> @Lakritzator nicely fixed, thank you, and get well soon.

@Christian-Schulz did the work 🙇‍♂️

PR #407, opened 2022-05-16
Fix initial crop selection

Christian-Schulz:
Bugfix crop selection scaling #404

Disable scaling for horizontal and vertical crop modes, as it makes no sense in these cases.

Btw, but I think this is known as I read the last commit. The centered scaling does nothing. It inflates the rectangle all times by 0.


Lakritzator:
I'm currently not at my best, combatting COVID with the whole family.
Might not have the energy to do much in the next few days.

PR #418, opened 2022-06-23
Added a "Color Picker to Clipboard" Button

mdiller:
This pull request adds a button on the side of the image editor that allows the user to click anywhere on the image to extract the color of that pixel, which it then copies to the clipboard in the usual #RRGGBB hex format. 

https://user-images.githubusercontent.com/3231343/175240326-b1a37d9d-2396-4cc4-9c05-4277bb0ea415.mp4

This is a feature/tool I've been wanting to have for a while, and have never found anything that works quite the way I want. I've been using greenshot for a while now and have gotten used to using it enough that I feel like its a perfect platform for something like this. I'd understand if this feature can't be added because people might not like it and theres no way to add a way to disable it because of a lack of localization strings, but I figured I'd try and see anyway in case ya'll decide it can be added. I'm happy to make any adjustments requested. I'd like to add some buttons at the top of the screen when active that can do things like disable/enable the "#" character, and toggle between lowercase/uppercase, but I'm guessing that'd probably be a no-go because we'd need localized strings for the tooltips. Speaking of tooltips, the added button is using the same string as the existing color picker form. Hopefully thats ok?

Also, I'll acknowledge that a similar functionality already exists inside the "Color Picker" form/dialog, but the idea of this feature is that it should be easy to use and should heavily reduce the amount of clicks needed to copy a color.

Let me know what you think!

### DIFF
--- a/src/Greenshot.Base/Interfaces/DrawingModes.cs
+++ b/src/Greenshot.Base/Interfaces/DrawingModes.cs
@@ -24,6 +24,7 @@ namespace Greenshot.Base.Interfaces
     public enum DrawingModes
     {
         None,
+        ColorPicker,
         Rect,
         Ellipse,
         Text,
--- a/src/Greenshot.Editor/Controls/Pipette.cs
+++ b/src/Greenshot.Editor/Controls/Pipette.cs
@@ -65,7 +65,7 @@ namespace Greenshot.Editor.Controls
         /// <param name="hotspotX">Hotspot X coordinate</param>
         /// <param name="hotspotY">Hotspot Y coordinate</param>
         /// <returns>Cursor</returns>
-        private static Cursor CreateCursor(Bitmap bitmap, int hotspotX, int hotspotY)
+        public static Cursor CreateCursor(Bitmap bitmap, int hotspotX, int hotspotY)
         {
             using SafeIconHandle iconHandle = new SafeIconHandle(bitmap.GetHicon());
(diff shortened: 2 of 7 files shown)


Lakritzator:
Hi Malcolm, thank your for contributing to Greenshot!!!

This is exactly why open source exists, you wanted to improve a product and took action ❤️ 

Your feature is definitively something I understand, and can be useful for others too.

I myself am wondering if the editor is the correct place for this, doesn't it make more sense to have a color "capture" mode directly in the region capture, so one does not need to make a screenshot first and open it in the editor?

mdiller:
Great, I'm glad you think this could be a good thing!

Personally I feel it's better in the editor, because I often like to grab a picture of something and then either grab different colors from several parts of the image, or make it easy to click a couple different pixels  in the area I'm interested in to make sure I'm getting the right color. it's also better the way if you want to grab an image from you screen that might be dynamic (from a video game etc) and then poke around in that for a bit.

We could also potentially implement both if you think people would be interested in that?

What would you suggest next steps are for this pull request?

albyrock87:
I honestly think that replacing clipboard content inside the editor with the color code is a bad idea. The user is not aware of it, because it's not the main purpose of that button.

IMHO having a color picker as a "Capture color" action would be definitely better from UX standpoint, though [PowerToys already has this feature done well](https://docs.microsoft.com/en-us/windows/powertoys/color-picker) except for the magnifier.

Lakritzator:

PR #428, opened 2022-08-04
Improve clipboard destination UX

albyrock87:
- On clipboard notification click, open the editor
  Like the Windows' Snipping Tool, when the use clicks on the notification, the screenshot editor will open
- When the editor was opened from a clipboard destination, prompt to copy to clipboard on close
- Add hero image to Windows 10 notifications with the screenshot preview
  image


albyrock87:
@Lakritzator what do you think about this? Thanks

Lakritzator:
I was on holidays, and will look at this in the coming days. (I have a lot of catching up to do)

Lakritzator:
So I had a look at this, and having the screenshot in the notification is a nice thing.
It would be important to keep the aspect ratio for the image though!

But I am not sure, you say "on clipboard notification click, open the editor".
I get notifications whenever I export a screenshot, and when I export I do not want to open in in the editor again.

Did you want to make it in such a way, that when someone copies something to the clipboard, outside of Greenshot, there would be a notification which you can click?
Maybe you can describe what exactly is what you want to happen?

albyrock87:
@Lakritzator the Windows 10 Snipping tool works this way:
1. User hits Win+Shift+S 
2. The capture tool appears
3. User selects the area
4. The area gets copied to the clipboard automatically
5. A notification with the image appears
6. If the user clicks on the notification, the snipping tool editor appears and the user can edit the screenshot 
7. Inside the tool the user can edit the screenshot and copy again the changed content to the clipboard or the user can save it to a file

I just wanted to implement the number 6 when the user chooses the clipboard as a destination.
In all the other use cases it makes no sense to open the editor given that the file has already been saved somewhere.

I may agree with you on the aspect ratio of the preview, but should we resize to fit and cut what cannot be inside or do we want to add black bars?
I prefer the first one.

Thanks

DraakjeYoblama:
This was exactly the feature I was looking for, I hope it gets implemented soon

albyrock87:
I received no answer in more than one year, so I'm going to close this PR.
If the maintainers want to add this, feel free to copy paste the code.

PR #431, opened 2022-08-16
Fix of BUG2951

jdavila71:
Changes to FIX BUG2951. Some newer Office installations use a different registry keys and some corporate GPOs block Office installation to write onto the Apps registry key. Changed the code so that the program looks in both the Apps path and the more standard OfficeInstall registry keys. This version was tested in Windows 8.1, Windows 10 and Windows 11 with no issues.

### DIFF
--- a/src/Greenshot.Base/Core/PluginUtils.cs
+++ b/src/Greenshot.Base/Core/PluginUtils.cs
@@ -40,13 +40,87 @@ namespace Greenshot.Base.Core
         private static readonly ILog Log = LogManager.GetLogger(typeof(PluginUtils));
         private static readonly CoreConfiguration CoreConfig = IniConfig.GetIniSection<CoreConfiguration>();
         private const string PathKey = @"SOFTWARE\Microsoft\Windows\CurrentVersion\App Paths\";
+        private static string[] strRootKeys = { @"SOFTWARE\Microsoft\Office", @"SOFTWARE\WOW6432Node\Microsoft\Office" };
         private static readonly IDictionary<string, Image> ExeIconCache = new Dictionary<string, Image>();
 
         static PluginUtils()
         {
             CoreConfig.PropertyChanged += OnIconSizeChanged;
         }
+        /// <summary>
+        /// Clear icon cache
+        /// </summary>
+        /// <param name="sender"></param>
+        /// <param name="e"></param>
+        public static string GetOfficeExePath(string keyname, string returnValue)
+        {
+            string strKeyName = "";
 
+            switch (keyname)
+            {
+                case "WINWORD.EXE":
+                    strKeyName = "Word";
+                    break;
+                case "EXCEL.EXE":
+                    strKeyName = "Excel";
+                    break;
+                case "MSACCESS.EXE":
+                    strKeyName = "Access";
+                    break;
+                case "POWERPNT.EXE":
+                    strKeyName = "PowerPoint";
+                    break;
+                case "OUTLOOK.EXE":
+                    strKeyName = "Outlook";
+                    break;
+                case "ONENOTE.EXE":
+                    strKeyName = "OneNote";
+                    break;
+                case "MSPUB.EXE":
+                    strKeyName = "Publisher";
+                    break;
+                default:
+                    strKeyName = "";
+                    break;
+            }
+            RegistryKey rootKey = null;
+            RegistryKey officeKey;
+            RegistryKey programKey;
+            RegistryKey installRootKey;
+
+            foreach (string strRootKey in strRootKeys)
+            {
+                rootKey = Registry.LocalMachine.OpenSubKey(strRootKey);
+
+                if (rootKey != null)
+                {
+                    string[] officeVersions = rootKey.GetSubKeyNames();
+                    if (officeVersions != null)
+                    {
+                        officeVersions = Array.FindAll(officeVersions, r => r.Contains("."));
+                        Array.Reverse(officeVersions);
+                        // string latestOfficeVersion = officeVersions.Where(r => r.Contains(".")).Max();
+
+                        foreach (string officeVersion in officeVersions)
+                        {
+                            officeKey = Registry.LocalMachine.OpenSubKey(strRootKey + "\\" + officeVersion);
+
+                            if (officeKey != null)
+                            {
+                                programKey = officeKey.OpenSubKey(strKeyName);
+                                if (programKey != null)
+                                {
+                                    installRootKey = programKey.OpenSubKey("InstallRoot");
+                                    if (installRootKey != null)
+                                        return installRootKey.GetValue(returnValue).ToString() + "\\" + keyname;
+                                }
+                            }
+                        }
+                    }
+                }
+            }
+            return string.Empty;
+        }
         /// <summary>
         /// Clear icon cache
         /// </summary>
@@ -84,7 +158,7 @@ namespace Greenshot.Base.Core
                 if (key != null)
                 {
                     // "" is the default key, which should point to the requested location
-                    return (string) key.GetValue(string.Empty);
+                    return (string)key.GetValue(string.Empty);
                 }
             }
 
--- a/src/Greenshot.Base/Core/PluginUtils.cs
+++ b/src/Greenshot.Base/Core/PluginUtils.cs
@@ -52,41 +52,23 @@ namespace Greenshot.Base.Core
         /// </summary>
         /// <param name="sender"></param>
         /// <param name="e"></param>
-        public static string GetOfficeExePath(string keyname, string returnValue)
+        public static string GetOfficeExePath(string keyname)
         {
-            string strKeyName = "";
-
-            switch (keyname)
+            string strKeyName = keyname switch
             {
-                case "WINWORD.EXE":
-                    strKeyName = "Word";
-                    break;
-                case "EXCEL.EXE":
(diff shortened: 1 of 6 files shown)


Lakritzator:
Very interesting, thank you for looking into that.
I'll make a code review now.

src/Greenshot.Base/Core/PluginUtils.cs
```diff
@@ -40,13 +40,87 @@ public static class PluginUtils
...
         static PluginUtils()
         {
             CoreConfig.PropertyChanged += OnIconSizeChanged;
         }
+        /// <summary>
+        /// Clear icon cache
+        /// </summary>
+        /// <param name="sender"></param>
+        /// <param name="e"></param>
+        public static string GetOfficeExePath(string keyname, string returnValue)
```
Lakritzator: Is there a reason for the returnValue param, this always seems to be "Path" and the method is already called GetOFficeExe**Path**, so might it make sense to remove the param?

src/Greenshot.Base/Core/PluginUtils.cs
```diff
@@ -40,13 +40,87 @@ public static class PluginUtils
...
+                case "POWERPNT.EXE":
+                    strKeyName = "PowerPoint";
+                    break;
+                case "OUTLOOK.EXE":
+                    strKeyName = "Outlook";
+                    break;
+                case "ONENOTE.EXE":
+                    strKeyName = "OneNote";
+                    break;
+                case "MSPUB.EXE":
```
Lakritzator: We do not use MSPUB (Publisher) so I do not see a reason for this.

src/Greenshot.Base/Core/PluginUtils.cs
```diff
@@ -40,13 +40,87 @@ public static class PluginUtils
...
 
+            switch (keyname)
+            {
+                case "WINWORD.EXE":
+                    strKeyName = "Word";
+                    break;
+                case "EXCEL.EXE":
+                    strKeyName = "Excel";
+                    break;
+                case "MSACCESS.EXE":
```
Lakritzator: We do not use MSAccess (Acces) so I do not see a reason for this.

src/Greenshot.Base/Core/PluginUtils.cs
```diff
@@ -40,13 +40,87 @@ public static class PluginUtils
...
+        /// <summary>
+        /// Clear icon cache
+        /// </summary>
+        /// <param name="sender"></param>
+        /// <param name="e"></param>
+        public static string GetOfficeExePath(string keyname, string returnValue)
+        {
+            string strKeyName = "";
 
+            switch (keyname)
```
Lakritzator: This is actually a perfect case for the [C# Switch-Expression](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/switch-expression)

src/Greenshot.Base/Core/PluginUtils.cs
```diff
@@ -40,13 +40,87 @@ public static class PluginUtils
...
+            RegistryKey rootKey = null;
+            RegistryKey officeKey;
+            RegistryKey programKey;
+            RegistryKey installRootKey;
+
+            foreach (string strRootKey in strRootKeys)
+            {
+                rootKey = Registry.LocalMachine.OpenSubKey(strRootKey);
+
+                if (rootKey != null)
```
Lakritzator, line 77: I like fail fasts, so `if (rootKey is null) continue;`
This would prevent deep nesting of ifs, reducing the read complexity.

src/Greenshot.Base/Core/PluginUtils.cs
```diff
@@ -40,13 +40,87 @@ public static class PluginUtils
...
+                    break;
+            }
+            RegistryKey rootKey = null;
+            RegistryKey officeKey;
+            RegistryKey programKey;
+            RegistryKey installRootKey;
+
+            foreach (string strRootKey in strRootKeys)
+            {
+                rootKey = Registry.LocalMachine.OpenSubKey(strRootKey);
```
Lakritzator, line 75: The return value of [OpenSubKey ](https://docs.microsoft.com/en-us/dotnet/api/microsoft.win32.registrykey.opensubkey?view=net-6.0#microsoft-win32-registrykey-opensubkey(system-string))is a [RegistryKey](https://docs.microsoft.com/en-us/dotnet/api/microsoft.win32.registrykey), which is an IDisposable, and needs to be disposed.

src/Greenshot.Base/Core/PluginUtils.cs
```diff
@@ -40,13 +40,87 @@ public static class PluginUtils
...
+
+                if (rootKey != null)
+                {
+                    string[] officeVersions = rootKey.GetSubKeyNames();
+                    if (officeVersions != null)
```
Lakritzator: `if (officeVersions is null) continue;`

src/Greenshot.Base/Core/PluginUtils.cs
```diff
@@ -40,13 +40,87 @@ public static class PluginUtils
...
+                    {
+                        officeVersions = Array.FindAll(officeVersions, r => r.Contains("."));
```
Lakritzator: You are allowed to use Linq, I do not think this is an extremely time critical loop or function.
officeVersions = officeVersions.Where(r => r.Contains(".").Reverse();

src/Greenshot.Base/Core/PluginUtils.cs
```diff
@@ -40,13 +40,87 @@ public static class PluginUtils
...
+                        Array.Reverse(officeVersions);
+                        // string latestOfficeVersion = officeVersions.Where(r => r.Contains(".")).Max();
+
+                        foreach (string officeVersion in officeVersions)
+                        {
+                            officeKey = Registry.LocalMachine.OpenSubKey(strRootKey + "\\" + officeVersion);
```
Lakritzator: Dispose the officeKey

src/Greenshot.Base/Core/PluginUtils.cs
```diff
@@ -40,13 +40,87 @@ public static class PluginUtils
...
+
+                            if (officeKey != null)
```
Lakritzator: `if (officeKey is null) continue;`

src/Greenshot.Base/Core/PluginUtils.cs
```diff
@@ -40,13 +40,87 @@ public static class PluginUtils
...
+                            {
+                                programKey = officeKey.OpenSubKey(strKeyName);
```
Lakritzator: Dispose programKey (you can just write using in front of it).

src/Greenshot.Base/Core/PluginUtils.cs
```diff
@@ -40,13 +40,87 @@ public static class PluginUtils
...
+                                if (programKey != null)
```
Lakritzator: if (programKey is null) continue;

src/Greenshot.Base/Core/PluginUtils.cs
```diff
@@ -40,13 +40,87 @@ public static class PluginUtils
...
+                                {
+                                    installRootKey = programKey.OpenSubKey("InstallRoot");
```
Lakritzator: Dispose installRootKey

src/Greenshot.Base/Core/PluginUtils.cs
```diff
@@ -40,13 +40,87 @@ public static class PluginUtils
...
+                                    if (installRootKey != null)
```
Lakritzator: if (installRootKey is null) continue;

jdavila71:
I implemented the suggestions that Lakritzator mentioned and tested again in Windows 8.1, Windows 10 and Windows 11 and it works with no issues. I have pushed the changes to my fork for your review.

src/Greenshot.Base/Core/PluginUtils.cs
jdavila71: I tried implementing this but I get error messages when I test it. This is why I use the Reverse from the Array class...

Lakritzator (approved):

Lakritzator (approved):

Lakritzator:
Thanks, looks already a lot better.  I'll check the the LINQ part, and I wonder if the disposing can be done with a using statement, this would prevent also that something isn't disposed when an exception happens. But that is just details 😄

Lakritzator:
FYI, I went through the code of the method you created, and used "using" and LINQ to optimize it a bit more, this is what I ended with:
https://github.com/greenshot/greenshot/blob/release/1.3/src/Greenshot.Plugin.Office/OfficeUtils.cs#L17

Thanks for the support!! I think it will fix the issue, so let's hope it doesn't have side effect.

I'll comment with the version that has your fix here, so we can test it in the open.

Best wishes,
Robin

Lakritzator:
P.S. one thing that was an error, was returning a string.Empty when nothing was found, this should have been null.

Lakritzator:
See: https://github.com/greenshot/greenshot/releases/tag/v1.3.270

jdavila71:
Tested version 1.3.270 on Windows 8.1, Windows 10 and Windows 11 using Office 2013, Office 365 32bit and Office 2021 x64 Enterprise and PRO versions with no issues... Cheers... If you need any assistance let me know.

PR #433, opened 2022-08-30
Push out, fit, and snap

FF-Brown:
Adds three related features, all accessed by right-clicking an object: 
1. Push any object outside the current surface, expanding the editing area with white space to accommodate it.
2. Stretch any object to fit the width or height of the current surface
3. Snap (move) any object(s) to the edge (any direction) of the current surface

As this is my first time contributing here, I'm not 100% clear on what's expected. For example, not sure if there is documentation I should update or where to add unit tests. Please let me know what changes need to be made. Also, I used a temporary icon for these menu items since I wasn't sure what y'all would want.

Again, I'm happy to make any changes required to get these features approved.


Lakritzator:
I just came back from my holidays, and wanted to have a look at this.
Any reason why you closed this pull-request?

FF-Brown:
Yeah, I just found a couple things that needed some more work. I'll open another PR once I clean it up a bit.

Lakritzator:
Ok, thanks for the quick feed-back. I really appreciate that you interact with this repo.

Although I didn't see the features live yet, and I am not 100% sure for what use-cases I would use these features, but they sound useful. Maybe you can describe some cases where one uses this?

I'll give you some hints on th ecode:

- Make sure you never use hard-coded strings for the UI, get them via our language system. e.g. `Language.GetString(LangKey.editor_downonelevel)` This also means you need to add translations to our language files.
- Try to group the items in the context menu, via a sub menu, so we do not get a huge list of things.
- Make sure you use the code guidelines, e.g. .editorconfig, I noticed you are having if/else without curly braces. Actually, use a switch statement for that:
```
if (direction == Direction.LEFT)
    left = targetElement.Width;
else if (direction == Direction.RIGHT)
    right = targetElement.Width;
else if (direction == Direction.TOP)
    top = targetElement.Height;
else if (direction == Direction.BOTTOM)
    bottom = targetElement.Height;

```

We cannot roll out features without translations, so I might enable them for when the use selected English but I won't like to have those entries in English in the context menu when the language is set to French. Translations are a lot of work, to contact all the people, so we usually do most changes in one go, and not every feature.

FF-Brown:
Definitely! I love Greenshot so I'm excited to be able to contribute.
I'll add those changes in with the cleanup I'm doing.

PR #449, opened 2022-10-12
Added localizations for zoom menu items

damiansikora85:


Lakritzator:
Thanks for helping us out, but as these are just the English texts, I'm not going to merge this.

I rather fix the resource loading that it falls back to English, or some other solution, otherwise we will overlook the fact that these languages do not have translations yet.

Issue #164, opened 2020-02-03
Can you change the style of the numbered counters?


cjohnsonuk:
**Is your feature request related to a problem? Please describe.**
Its a question to see if this option already exists and a feature request if not

**Describe the solution you'd like**
I'd like to be able to choose the background colour, text colour and  shape of the counters to fit our colour scheme.

**Describe alternatives you've considered**
I've considered having a folder of transparent .pngs to drag into my screen shot but its just not as convenient, especially when on a laptop with limited screen estate.  I've looked through the source as much as I can to see if there's a config somewhere or a bunch of images used to create the counters but can't find anything.

ryantheleach:
I'd like to see it be able to be changed from numbers to letters as well. e.g. A,B,C

lkurasinski:
@cjohnsonuk you can change  background colour and text as well, but nothing more than solid/transparent.

I'd love to see also an option to change label of 'number' manually.

stamilselvan:
Would be nice to reset the counter and start again. Flexibility of negative numbers is also thinkable.

Lakritzator:
Sorry about the late response, well it's a free product done in my free time... bla bla, many reasons.

I usually try to answer easy questions as quick as possible, whenever I see them. But with this I had some questions.
If you use the numbered counter, you can change the color of the bubble and the number, so that part should be covered, right?

The shape is a different topic, there is currently only a bubble. What kind of shapes can you imagine?
Will this feature also help: #358

@ryantheleach that would be something like a counting schema, e.g. counting up/down, alphabet, roman numbers?

@lkurasinski I see an issue with that, that if you change something it doesn't fit into the schema again. You currently can up/down which also changes the "attached" bubbles.

@stamilselvan Do you mean that you have multiple sets of bubbles?

In general I want to be clear, we could in theory build many things, but the concept of Greenshot is "keeping it simple". Having many UI settings, which are hard to explain, is not our current way of working. If we find a solution which fits into that thinking and covering many use-cases, than it's good.

ryantheleach:
@Lakritzator basically. At times there's instances where I need to markup 2 different categories of counts. E.g. 1,2,3 and A,B,C in the same image, It's rare and far between, and easy enough to do by hand, but figured it'd be worth mentioning.

stamilselvan:
@Lakritzator: Many thanks for your effort and keeping the community sprint! 
[...]

PR #458, opened 2022-11-10
Allow user to add an outline and a drop shadow to the step label.

wskellenger-intrepid:
Allow an outline and a drop shadow around the StepLabelContainer.

New options appear:

image

And the result:

image

### DIFF
--- a/src/Greenshot.Editor/Drawing/StepLabelContainer.cs
+++ b/src/Greenshot.Editor/Drawing/StepLabelContainer.cs
@@ -147,6 +147,8 @@ namespace Greenshot.Editor.Drawing
         {
             AddField(GetType(), FieldType.FILL_COLOR, Color.DarkRed);
             AddField(GetType(), FieldType.LINE_COLOR, Color.White);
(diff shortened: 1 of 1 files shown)


wskellenger-intrepid:
Related to #164 and https://greenshot.atlassian.net/browse/SUPPORT-115?oldIssueView=true

Lakritzator:
I'm not 100% sure who is going to use this, but there is no direct reason not to allow it.
Thank you for your contribution.

wskellenger-intrepid:
> I'm not 100% sure who is going to use this, but there is no direct reason not to allow it. Thank you for your contribution.

Hi @Lakritzator, documentation for our application used exactly this type of notation, and they were all created by hand, and I'm not sure which program was used for this.  (for example see: https://github.com/intrepidcs/VSpy3-Docs/blob/main/application-notes/application-note-vehiclescape-standalone-logging/application-note-vehiclescape-standalone-logging-part-1-message-collection-options.md) 

Now when I update the documentation I just use Greenshot with the outline option, I've been using my branched build since I made this pull request.

PR #459, opened 2022-11-11
add outline thickness, dropshadow options to counter feature for 1.2.10 version

wskellenger-intrepid:
This is the same as #458, but targets the 1.2.10 release branch


Lakritzator:
I'm going to close this, #458 was merged for 1.4 and we do not backport this.

Issue #457, opened 2022-11-04
Counter numbers above 9 are all displayed as 1


Doublemyst:
When adding counter numbers, after 9 I only see 1. In order to see the whole number, the ball has to be resized.

Greenshot version used is 1.3.270.

P.S. In "about greenshot" there is an unneeded ")" symbol either after "64", or after "bit".

FF-Brown:
Yeah, I've noticed this as well. I can take a look at this if you'd like @Lakritzator

Lakritzator:
Oh I see, either the font is too big or the circle to small.
@FF-Brown sure that would be nice, I'm busy working on a huge PR in which there still are some issues.

I'm not sure where to look either, but the StepLabelContainer would be a good start.

There is a line which calculates the font size, maybe there is something with it:
float fontSize = Math.Min(Math.Abs(Width), Math.Abs(Height)) / 1.4f;

PR #460, opened 2022-11-12
Calculate optimal font size for StepLabel  #457

Christian-Schulz:
Set a smaler fontsize if the counter has more then one digit.

font_size_steplabel


mischlrebl:
very cool!

jklingen:
Thanks a lot for this fix, so much better than the quick fix we added last week 😆

PR #461, opened 2022-11-14
Jg/221114 testbuild arm64

jglathe:
Enable target Windows 11 ARM64.

Minimal changes required to get it to compile for ARM64, with corresponding changes in setup.iss to cater for the "no, I won't set all subprojects to ARM64, some stay on Any CPU" behaviour. ARM64 builds in bin\ARM64\release instead of bin\release, and this requires some changes.
There are other artifacts in the .csproj files I have no explanation for, only that VS2022 ARM Preview may do this automatically.

The installation package works nicely here.


siggipop:
Ok sooo is this (or will it) be part of the release? or do i have to build it locally as for now?

Lakritzator:
I'll have a look at this, I know that .NET Framework 4.8.1 should even work better with ARM...
But I need some time, especially with the amount of conflicts, that I probably caused most of myself... 😱

jglathe:
Hmm interesting. Shall I rebase and try to ensure that its mergeable?

Issue #482, opened 2023-04-13
Print Screen Hotkey does not work anymore with Windows 11 KB502531


fratzi:
Hi,

with Windows 11 KB5025310 (Insider Build) windows is using print-screen button for opening snipping tool.
With this, Greenshot can't map and overwrite anymore the printscreen button. There is also a error message at the first start of greenshot. 

My Greenshot Build is 1.2.10 build 6

fratzi:
This was just for information - maybe GS could inform about the possibility to disable the new windows built in screenshot feature. Currently it's only faulting with an error message, that it can't bind print-screen button.

jglathe:
Agree. However I thought about doing just this, on install? https://www.elevenforum.com/t/enable-or-disable-use-print-screen-key-to-open-screen-snipping-in-windows-11.520/

The registry key would look like this :

```
Windows Registry Editor Version 5.00

[HKEY_CURRENT_USER\Control Panel\Keyboard]
"PrintScreenKeyForSnippingEnabled"=dword:00000000
```

I tried it out on my wdk2023 (arm64 😁) and it worked nicely. If the default PrtScr feature is disabled, first start after install doesn't have an issue either. 

![2023-04-15 20_22_17-Print Screen Hotkey does not work anymore with Windows 11 KB502531 · Issue #482 ](https://user-images.githubusercontent.com/33185912/232247359-bcd7417f-3c1e-4255-8b58-1e6d24f7195e.png)
![2023-04-15 20_23_34-Info über Windows](https://user-images.githubusercontent.com/33185912/232247361-1abd7e83-3dfc-424a-b249-788b12c5c5c3.png)

It may be interesting to note that the registry key actually isn't there first, but it will be changed according to the setting of the feature in windows settings. I re-confirmed it by setting and un-setting it via regedit, the setting changed in the dialog (after reopening it), too.

![2023-04-15 20_34_17-Einstellungen](https://user-images.githubusercontent.com/33185912/232247539-4d205be8-e98e-46c0-a649-f6f5b7e4c249.png)

jglathe:
[Binaries for x86 and ARM64](https://github.com/jglathe/greenshot/releases/tag/1.3.280-arm64)

Slesa:
Well, I have the "Use Printscreen" disabled, but it still does not work. This is so nasty... I wish I could keep Windows 10 :-/

arkadiusz-koziel-nc:
[...]

PR #484, opened 2023-04-16
add install option to disable the default win11 prtscr tool

jglathe:
Hi there, inspired from #482 and with the help of ChatGPT I hacked out the installer to disable the PrtScr default tool on Windows 11 installations. It is shown as an extra option if you're on Win11 >= Build 22000. This can be refined further if required. Built and tested on wdk2023, arm64 - the PR is platform independent, though.


jglathe:
2023-04-16 16_48_37-installer - FreeCommander XE

Lakritzator:
That's pretty cool, I was thinking something along that line, but you beat me to it 😏 

There are a few challenges with that:

- Translations, we have way to many languages and not a good process to update them. (this slows us down dramatically)
- Is it a one-time thing? If someone installs it today and tomorrow they update Windows 11 (it's currently in insiders) I am not sure if this sticks? But what I mean, is that it will not help later.

Let me see how to handle those things, but I can merge it for 1.3, and decide later.

jglathe:
Could be an error handler if the hotkey hook is set. Would be the next thing to complete the handling of this. Re translations: DeepL is pretty okay.

Lakritzator:
> Could be an error handler if the hotkey hook is set. Would be the next thing to complete the handling of this. Re translations: DeepL is pretty okay.

We have a meeting this month, where I will address the topic on how to continue, it is slowing down any UI development. I could imagine we could use DeepL or ChatGPT etc, but we will never be able know it it's a good translation.

Carlos300000:
Hey there @Lakritzator, 

is this also an option in the Inno-Installer-Inf-File, that we can use for silent installations?

Best regards,
Carlos

PR #497, opened 2023-05-31
Corrects designer issues in Visual Studio (MS)

GrandStrateguerre:
Hello,

I correct some designer issues, if you agree , can you merge ?

I did some testing, seems to be good.

Best regards


GrandStrateguerre:
There is better way, I'll redo

GrandStrateguerre:
There is better way, I'll redo

PR #498, opened 2023-06-07
Added ability to hide destinations from the picker context menu.

KieranDevvs:
This pull request adds the concept of hidden destinations. If a destination is hidden then the type is still loaded into the DI container (to be able to manage and unhide the plugin) but it will not be shown as an option for the user to interact with. This functionally lets you hide options you have no intention of using, while still letting the user manage the options via the UI rather than having to find the config file in `%appdata%` and configure the `ini` file themselves.

I initially planned for this feature to use the `excludedDestinations` feature however, when attempting to implement it, I found out that if the destination is excluded then its not loaded into the DI system and thus cannot be managed at runtime (to unexclude it).

Thus the introduction of a 3rd state:
 - Hidden
   - The destination is loaded at runtime but is hidden from most user interactivity
 - Excluded
   - The destination is not loaded at runtime
 - Active
    - The destination is functional

image
image
image

### DIFF
--- a/src/Greenshot.Base/Core/CoreConfiguration.cs
+++ b/src/Greenshot.Base/Core/CoreConfiguration.cs
@@ -212,6 +212,9 @@ namespace Greenshot.Base.Core
         [IniProperty("ExcludeDestinations", Description = "Comma separated list of destinations which should be disabled.")]
         public List<string> ExcludeDestinations { get; set; }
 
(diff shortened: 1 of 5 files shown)


KieranDevvs:
Closing as the repository seems to be abandoned.
I've moved over to [ShareX ](https://github.com/ShareX/ShareX)as it has more features (i.e gif support), based on .NET and is open source.

PR #512, opened 2023-08-22
Multiple highlights

FF-Brown:
Fixes an issue (I'll link the report later) identified with the area highlights. Each highlight darkens the entire area around it, including other highlight areas. This means that if the user places multiple highlight areas, they all get progressively darker and blurrier. 
My fix just passes the locations of other highlight areas in when we draw one of them so each can be excluded from the filters. Not incredibly efficient but I can't see users having any more than 3-4 of these, so the effect should be super minimal, otherwise I'd have done it differently. For some reason this operation is very slow to begin with, which might be worth looking into at some point.

Feedback is welcome!


FF-Brown:
Forgot to include these before/after
Before:
greenshot_before_highlight_changes_high_quality
After:
greenshot_after_highlight_changes_high_quality

FF-Brown:
Performance on this is not great, looking into a fix.

FF-Brown:
Added this commit just to demonstrate that we could apply the same changes to the grayscale filter. Will revert if we decide not to go that route, though.

FF-Brown:
This one I'm not entirely happy with, but it does the job. Significantly improves the performance when adding highlight areas (specifically the ones that blur and darken). It's just not a very clean way of doing it. Required more "special case" code than I like. Would love some input on it. Still looking at other options. 

It does fix the brightness issue as well, though (getting darker with each added highlight area)
image

src/Greenshot.Base/Interfaces/Drawing/IDrawableContainer.cs
```diff
@@ -61,6 +61,8 @@ public interface IDrawableContainer : INotifyPropertyChanged, IDisposable
 
         bool HasFilters { get; }
 
+        bool IsAreaHighlightContainer { get; }
```
Lakritzator, line 64: My first hunch would be to have a different name, or see if we can place this outside the IDrawableContainer.

Lakritzator:
Doing a review now, as this does make sense and solves some issues.
I also need to place this solution against #622

src/Greenshot.Editor/Drawing/DrawableContainer.cs
```diff
@@ -57,6 +57,9 @@ public abstract class DrawableContainer : AbstractFieldHolderWithChildren, IDraw
         private const int M11 = 0;
         private const int M22 = 3;
 
+        public static bool IsBlurFilterApplied = false;
+        public static bool IsBrightnessFilterApplied = false;
```
Lakritzator, line 61: static (&public) fields are definitively a no-go from my side, this screams side-effect / threading issues. Especially if someone has 2 editors open. Let's find a solution here, if we go forward with this.

Lakritzator:
I hope you do not mind, but I merged your idea/work with that of @Kissaki into #1350

PR #514, opened 2023-08-26
Fix scaling with fixed ratio

jairbubbles:


jklingen:
Thanks for the bugifx, great job 💪

jairbubbles:
So glad to see some activity here! 🥰

PR #528, opened 2024-01-04
ImeMode = ImeMode.On → ImeMode.NoControl

wsryu91:
When inserting text in Korean Windows, 
there was an error where English letters were entered full-width instead of half-width.

Changes were made to correct errors.


Lakritzator:
I was slightly hesitant, as I have absolutely no experience with Ime, as I have no need for it.
But it seems to make sense, approving this.

Lakritzator:

PR #553, opened 2024-09-23

Felipefpl:
Hi, i updated the translation for my language, i also would like to report a few things:

1 - The settings' GUI doesnt resize itself to make space for the translated words that are bigger than the english ones.

2- It would be nice if all dialog boxes/settings' windows started always in the center of screen.


Felipefpl:
I submitted this TWO DAMN YEARS AGO and only now you care to say something about it and even so it's through IA?? I'm quitting this. This behaviour towards someone who cares about your project enought to send a translation makes this unworthy of all my effort and the others's efforts too.

PR #586, opened 2025-05-24
Japanese language files update

Rukoto:


jklingen:
Thanks a lot for the update 💪

PR #594, opened 2025-05-30
Extend SimpleServiceProvider, include the ability to remove services

Christian-Schulz:
The function is not currently used. 
This allows us to use the `SimpleServiceProvider` later in a test framework.


Lakritzator:
I need to look at bit more in detail on this, as I would expect conflicts while using the dictionary and lists, already in the current implementation but packing more on it enhances the risk...

PR #595, opened 2025-05-30
[Fix] Add lock for IniConfig.SectionMap

Christian-Schulz:
I added a lock for IniConfig.SectionMap.
This fixes an issue with e.g. simultaneous running unit tests.


Lakritzator (approved):
I'm absolutely no fan of locks, but due to the design of this code I would agree with this. Can't see any reasons not to accept this fix.

Issue #546, opened 2024-07-23
Greenshot 1.3.277 (Pre-Release) - Installation Routine Issue


ll4mat:
Hi There

Just tested the pre-release candidate (1.3.277). The following was found:

When the installer is started in user scope, "C:\<username>\AppData\Local\Greenshot" is suggested as the installation directory. If the installer is started as admin (elevated), "C:\Program Files\Greenshot" is suggested as the installation directory. As far as good.

In both cases, however, the corresponding registry key containing the version information, uninstall-strings, installdate and such is created in HKCU hive "HKEY_CURRENT_USER\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\Greenshot_is1", which - for several reasons - is bad.

On a installation in admin context (machine-based), i expect the corresponding registry key to be created either under ‘HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall’ or under ‘HKEY_LOCAL_MACHINE\SOFTWARE\WOW6432Node\Microsoft\Windows\CurrentVersion\Uninstall’ where it belongs to, and not in HKCU. 

For an installation in user-scope HKCU is fine. 
For an installation in machine-scope, it just sucks.

LCI-Kevin-LAURENT:
Hi, 
I find myself in the same situation. 
After deploying greenshot 1.3.277 in admin on workstations, the installation files are located in "C:\Program Files\Greenshot" however the registry key is not in the right place because it is in HKU  : 
[HKEY_USERS\.DEFAULT\Software\Microsoft\Windows\CurrentVersion\Uninstall\Greenshot_is1].
Instead of being in HKLM as suggested in the post above
Which means that Windows does not even see the software as installed because it does not appear in the list of installed programs.

chbwien:
And there is another bug: If you have already installed the program as admin in c:\program and restart the installer as user (because you simply forgot to elevate the installer, which is really unusual), the old settings will be used (destination c:\program) and the installer will simply fail due to lack of write permissions in the destination.

bratensauce:
I just installed the final version 1.3.290 and unfortunately this issue is still present.

Not nice at all. I was looking forward for this release for the Win11 upgrade issue to be fixed but I guess I cannot deploy this version with such an ugly bug.

<img width="1888" height="876" alt="Image" src="https://github.com/user-attachments/assets/1a728eb3-1e61-4424-bb3e-4a92549a06a3" />

[...]

PR #614, opened 2025-07-15
Write Uninstall Information to HKCU or HKLM depending on User/Admin Level #546

jklingen:


jklingen:
Better fixes have been applied with 2121547387e8867ba4f379fc5f8a7f1bcdd45982 and #641. This pull request is no longer needed.

PR #621, opened 2025-07-16

Felipefpl:
Adde the new strings and fixed the old ones.


Christian-Schulz:
I think you have selected the wrong target branch, which is why so many changes are displayed

Felipefpl:
@Christian-Schulz - i fixed my submission but i cant add my PR now. Do you know what coud be happening?

PR #636, opened 2025-07-24

Felipefpl:
Added/fixed the strings as needed.


jklingen:
I noticed this is the same as #636 (just pointed to different branches). Closing this one, we'll work with the other one.

Felipefpl:
Ok, thanks.

PR #574, opened 2025-02-08
#572 Add ArrowHeadCombination to BinaryFormatterHelper

FF-Brown:
(cherry picked from commit 561cbf760754ec14da4bf8d89a86b40a4806cdd6)

Looks like this enum just got missed in the types dictionary in BinaryFormatterHelper. Should resolve #572

Issue #375, opened 2022-02-07
Solution design for a new Greenshot file format


Lakritzator:
**Is your feature request related to a problem? Please describe.**
The current Greenshot uses as it's own file-format a PNG, followed by the serialized surface (the canvas) elements. Unfortunately this is not only very proprietary, it's also not stable over versions. This means that something written in Greenshot 1.2 cannot be loaded in 1.3 etc.

**Describe the solution you'd like**
I'd like a file format which is clearly documented, and even could be used by other applications.

**Describe alternatives you've considered**
Use SVG

**Additional context**
...

Lakritzator:
I've extracted most requirements of a potential file format from the previous information.
So we need something which can store a background image, and draw different annotations on top of it.
These annotations could be different shapes (lines, arrows, circles, rectangles, freehand) in different colors, text and images.

It's important to be able to create this file from the information in the editor, and be able to read the file to build the editor state back again.

This means we need to store:

- The captured image
- An ordered list of elements to draw on the captured image
- The images for the elements
- The text for the elements, with the font information
- The locations of the elements
- The extra positions of the elements (e.g. the tail of the speech bubble)
- The points (path) for the freehand drawing
- For the shape elements the line thickness, line color, fill color, arrows 

Some additional requirements:

- Multiple layers
- Lossless (no quality loss)
- All image elements are stored with their original information, so they can be resized later without additional quality loss.
- Meta data (like timestamp etc)
- Line caps & dash (to enable more ways to draw elements)


Some optional requirements:

- The file can be read by other image viewers, by piggy-bagging on an existing format. e.g. store all the information in a new PNG chunk.

Lakritzator:
I guess it should be possible to use SVG, but the specs are very extensive and complex. Also the goal of SVG is not the same as we need for Greenshot.
I also looked at draw.io, but my first impression was that it's very complex and I'm not sure if it wouldn't cost me more time to use it and maintain it then to make my own.

I wrote a [tweet](https://twitter.com/lakritzator/status/1564312166668402689?s=20&t=vRMv5-0uCopAUFDXtEa8AQ), to see if someone has an idea for this.

Otherwise I could **imagine** to build something like this:

- Build a set of classes to represent a json format (due to the ease of de/serializing with .NET) for the information I need.
- Convert the in memory representation of the editor information to this set of classes, and serialize this to a json format
- Use a .zip archive to store the files like the capture, the serialized json and all other images.
[...]

PR #638, opened 2025-07-26
New Greenshot file format

Christian-Schulz:
First of all, I must apologize. The pull request is very large, but there was no other way.

Everything else that is not directly related to the Greenshot file format i moved to other PR's.

https://github.com/greenshot/greenshot/pull/593 (✅merged in main)
https://github.com/greenshot/greenshot/pull/594 (✅merged in main)
https://github.com/greenshot/greenshot/pull/595 (✅merged in main)

---
### Overview

The main reason for this PR is that we need to move away from using BinaryFormatter and separate the serialization functions from the domain classes. 

* This PR implements a new file format (V2) for Greenshot editor files (.greenshot) and templates (.gst), replacing the BinaryFormatter with MessagePack serialization. 
* The implementation is backward compatible for all files up to Greenshot version 1.2.
* I added a new test project with xUnit.
* I refactored most of the .greenshot file related stuff and moved it to the new namespace `Greenshot.Editor.FileFormat`.
* I refactored all `DrawableContainer` classes and removed all [Serializable] attributes.

### Serialization

I chose MessagePack as the new serializer because it is recommended as a replacement. It supports everything we need so far and also allows us to handle future file format changes and content extensions.
There are Greenshot Editor files (.greenshot) and templates (.gst). These are very similar to each other. Everything I write about .greenshot below also applies to .gst.

The `Greenshot.Editor.FileFormat.GreenshotFile` represent a .greenshot file with domain data.
In `Greenshot.Editor.FileFormat.DTO` are Data Transfer Objects that represent a .greenshot file in a lightweight structure. Main class is `GreenshotFileDTO`. 
These DTO classes are serialized using MessagePack.
`Greenshot.Editor.FileFormat.Dto.ConvertDomainToDto.ToDto()` is used to convert `GreenshotFile`  into DTO.
`Greenshot.Editor.FileFormat.Dto.ConvertDtoToDomain.ToDomain()` performs the same operation in reverse.

These changes together ensure that the serialization logic is largely separated from the domain classes.

I have implemented a new versioning scheme for .greenshot files. Previously, the file version was closely linked to the application version. From now on, the versions are separate.
The old files for Greenshot 1.2 and 1.3 are now handled under the new version V1 in `Greenshot.Editor.FileFormat.V1.GreenshotFileV1`.
The new files in `Greenshot.Editor.FileFormat.V2.GreenshotFileV2`.

The new .greenshot file is just like the old ones a PNG file with binary data at the end. **The main difference is that now the rendered editor image is saved.** This includes all graphical elements in the editor. The raw screenshot from surface is also included in the binary data.

For copy and paste within the Greenshot Editor i used the same serialization mechanism as for the file format. 

More information can be found here: [Greenshot.Editor/FileFormat/readme.md](https://github.com/Christian-Schulz/greenshot/blob/feature/new_greenshot_file_format_v2/src/Greenshot.Editor/FileFormat/readme.md) .

### Backward compatibility

To support all old files there is the new namespace `Greenshot.Editor.FileFormat.V1.Legacy`. 
This is the only place where the BinaryFormatter is used.

Some rare `DrawableContainers` are serialized differently in Greenshot 1.2 and 1.3. I have implemented a workaround here that allows both variants to be read.

The main functionality, which can read and write every known file version, is implemented in the new `Greenshot.Editor.FileFormat.GreenshotFileVersionHandler`.

### Refactoring

I removed all [Serializable] attributes and most of serialization-specific code from `DrawableContainer` classes. The [Serializable] attributes are only needed in legacy-container classes.

All code from `Surface`, `ImageEditorForm` and `ImageIO` that dealt with reading and writing .greenhot files has been simplified and moved to `GreenshotFileFormatHandler`. (and to the new `GreenshotTemplateFormatHandler` for .gst files)

### Testing

I added a new test project.
The test data contains .greenshot files with all individually possible `DrawableContainers` for versions 1.2 (V1), 1.3 (V1), and V2.
The unit tests test the loading of each individual file.
There are unit test for `ConvertDomainToDto` and `ConvertDtoToDomain` for every `DrawableContainer`.
The `ConvertDomainToDto` test also checks container default values.
There are unit test that test the entire process from serialization to deserialization.
And there are a few unit tests to check the bugs I found along the way.

### Special notes

There is a `CursorContainer` that is not being used. I think this was intended for the mouse cursor, but for a capture with mouse cursor we use the `IconContainer`.
So, is the `CursorContainer` obsolete?

For the `MetafileContainer` i implemented some workarounds. For serialization with MessagePack we have to store the original source file, but GDI+ does not support saving .wmf or .emf files, because there is no encoder. So I had to create a replacement. During the deserialisation with BinaryFormatter of old file formats, i convert the internal metafile to a PNG. Visually in the editor, I didn't saw any difference. There is no problem with newly created `MetafileContainers`, as we have the original source file.

The BinaryFormatter is still used, but could easily be removed.

For the future, after switching to .NET9+, we could implement a new file format that is a valid PNG with the help of ImageSharp. That could contains the serialized data as PNG chunks. This would only require a few classes to be adapted.

---

I'm leaving the PR as a draft because I think there will be a few more adjustments based on the comments before it's finally finished.

_Please feel free to mark typos in my code._[](url)


jklingen:
Wow, that's impressive 🤩 I didn't have a closer look into the implementation yet, but the detailed description shows these PRs are carefully crafted. Thanks for your efforts 👍  

> First of all, I must apologize. 
lol, no - not really 😏 this is an enormous task and I totally understand that it cannot be sliced into smaller pieces than that.

I will try to have a look at it as soon as possible - given the large number of changes, this pull request shouldn't be left lying around for too long.

Christian-Schulz:
@FF-Brown, I had a merge conflict with your changes from #574, I'm pretty sure I re-implemented the same fix. Can you double check that?

FF-Brown:
@Christian-Schulz I don't see any changes to BinaryFormatterHelper in this PR, and that's the only file touched by #574. Where's the conflict?

Christian-Schulz:
> Where's the conflict?

I deleted Helpers/BinaryFormatterHelper.cs  and moved it to FileFormat/V1/Legacy/LegacySerializationBinder.cs

FF-Brown:
Ah didn't see that, will take a look

FF-Brown:
> I'm pretty sure I re-implemented the same fix

Yep looks good to me

Neustradamus:
It is needed to change of version, go to 2.0 with this PR.

Christian-Schulz:
> It is needed to change of version, go to 2.0 with this PR.

No, from the user's point of view, these changes are not that significant.

src/Greenshot.Editor/Drawing/ArrowContainer.cs
```diff
@@ -56,7 +55,6 @@ public ArrowContainer(ISurface parent) : base(parent)
         protected override void InitializeFields()
         {
             AddField(GetType(), FieldType.LINE_THICKNESS, 2);
-            AddField(GetType(), FieldType.ARROWHEADS, 2);
```
jklingen: Question: has this been removed intentionally?

src/Greenshot.Editor/FileFormatHandlers/GreenshotFileFormatHandler.cs
```diff
@@ -22,112 +22,96 @@
...
+        }
+
+        return false;
+    }
 
-        private ISurface LoadSurface(Stream surfaceFileStream)
+    /// <summary>
+    /// <inheritdoc />
+    /// </summary>
+    /// <remarks>This implementation loads the <see cref="GreenshotFile"/> from stream. Use this to creates a <see cref="ISurface"/> and uses <see cref="Surface.GetImageForExport"/> wich renders all contained elements into the image.</remarks>
```
jklingen:
```suggestion
    /// <remarks>This implementation loads the <see cref="GreenshotFile"/> from stream. Creates a <see cref="ISurface"/> and uses <see cref="Surface.GetImageForExport"/> wich renders all contained elements into the image.</remarks>
```

src/Greenshot.Editor/FileFormatHandlers/GreenshotFileFormatHandler.cs
```diff
@@ -22,112 +22,96 @@
...
-            {
-                var greenshotMarker = streamReader.ReadToEnd();
-                if (!greenshotMarker.StartsWith("Greenshot"))
-                {
-                    throw new ArgumentException("Stream is not a Greenshot file!");
-                }
+    /// <summary>
+    /// Load a <see cref="ISurface"/> from file path
+    /// </summary>
+    /// <remarks>This implementation loads the <see cref="GreenshotFile"/> from file. Use this to creates a <see cref="ISurface"/>.</remarks>
```
jklingen:
```suggestion
    /// <remarks>This implementation loads the <see cref="GreenshotFile"/> from file. Creates a <see cref="ISurface"/>.</remarks>
```

src/Greenshot.Editor/FileFormat/readme.md
jklingen, line 1: Awesome job with the documentation 🤩

src/Greenshot.Editor/FileFormat/readme.md
```diff
@@ -0,0 +1,216 @@
...
+
+# Greenshot Editor File
+
+- Greenshot editor files are saved with the `.greenshot` file extension.
+- Class `GreenshotFile` represents a .greenshot file as domain object.
+- It contains the PNG image data that came from `Surface` (the editor canvas). More precisely from `Surface.Image`.
+- It also contains a `DrawableContainerList` that came from `Surface.Elements`. A `DrawableContainer` object represents a graphical element (such as text, shapes, and highlights) that are placed on the editor canvas.
+- For serialization/deserialization, the `GreenshotFile` is converted to a Data Transfer Object (DTO) `GreenshotFileDto`, which contains the data to be serialized.
+
+## Supported File Formats
```
jklingen, line 17: Note: the section "Supported File Formats" (and sub sections) are duplicated in here.

jklingen:
@Christian-Schulz JFYI, quick update: had the first deeper dive into the code today - really impressive job so far. I need to review in slices, though 😅

Christian-Schulz:
Hi @jklingen ,

just wanted to kindly check if you’ve had a chance to review this PR yet.
Thanks a lot for your time and effort! 🙏

jklingen:
@Christian-Schulz sorry for the loooong wait 😞 I was actually planning to find the time this week (still am). Longer time slots for Greenshot have been rare in the recent weeks, and the few I had were eaten up by other topics. Thanks again for your patience and ongoing support.

src/Greenshot.Editor/Drawing/CursorContainer.cs
```diff
@@ -34,22 +32,21 @@ namespace Greenshot.Editor.Drawing
 {
     /// <summary>
     /// Description of CursorContainer.
+    /// This Container is not in use. For a capture with mouse cursor the IconContainer is used cctor Surface(ICapture capture) in <see cref="Surface"/>.
```
jklingen: Good catch! Created #698 to clean this up in future.

src/Greenshot.Editor/Drawing/DrawableContainerList.cs
```diff
@@ -780,5 +780,14 @@ public void AdjustToDpi(int dpi)
                 drawableContainer.AdjustToDpi(dpi);
             }
         }
+
+        /// <summary>
+        /// Creates a clone by using serialization
```
jklingen:
```suggestion
        /// Creates a clone by using our own DTO/domain conversion logic.
```

src/Greenshot.Editor/Drawing/MetafileContainer.cs
```diff
@@ -74,6 +102,37 @@ protected override void Dispose(bool disposing)
...
+        {
+            if (image == null) throw new ArgumentNullException(nameof(image));
+
+            using (Bitmap tempBitmap = new Bitmap(1, 1))
+            using (Graphics referenceGraphics = Graphics.FromImage(tempBitmap))
+            {
+                IntPtr hdc = referenceGraphics.GetHdc();
+                try
+                {
+                    // Erstelle ein neues Metafile mit der Größe des Bildes
```
jklingen:
```suggestion
                    // Create a new Metafile with the size of the image
```

src/Greenshot.Editor/FileFormat/Dto/Container/DrawableContainerDto.cs
```diff
@@ -0,0 +1,66 @@
...
+[Union(10, typeof(HighlightContainerDto))]
+[Union(11, typeof(MetafileContainerDto))]
+[Union(12, typeof(ObfuscateContainerDto))]
+[Union(13, typeof(StepLabelContainerDto))]
+[Union(14, typeof(SvgContainerDto))]
+[MessagePackObject]
+public abstract class DrawableContainerDto 
+{
+  
+    [Key(10)]
```
jklingen: Question: did you consider using ``StringKey` instead of `IntKey`?
I know that `IntKey` is much faster than `StringKey`, but instictivly I'd rather choose named keys, because it seems the more flexible and less error-prone variant 🤔 
That said, I have to admit that I never worked with MessagePack before, just read up about the keys a bit.

Assumption: the majority of screenshots have less than 10 elements drawn on them, and very few screenshots have more than 100 elements. (absolutely non-represantive recap of what I saw people doing with Greenshot so far 😁 )

src/Greenshot.Base/IniFile/IniConfig.cs
```diff
@@ -389,7 +397,12 @@ public static IEnumerable<string> IniSectionNames
         {
             get
             {
-                foreach (string sectionName in SectionMap.Keys)
+                List<string> sectionNames;
+                lock (SectionMapLock)
+                {
+                    sectionNames = [.. SectionMap.Keys];
```
jklingen: Question: when running the versino locally, I need to set `<LangVersion>preview</LangVersion>` in order to make the spread operator compile. Does this work for you out of the box?

jklingen:
@Christian-Schulz once again... impressive job 🤯 😆 

The amount of changes didn't fit into the usual slots I can take for Greenshot, and I noticed that doing it in tiny slices doesn't work so well - so it needed to wait until I am on holiday and can afford to grab a larger slot. Thanks again for your patience. I left a few comments, mainly questions and some nit-picks.

Great that you added unit tests as well 💪 

I will need to do some more manual testing though. I gave it a quick try today and ran into a null reference exception, but it wasn't clearly reproducible, so I'll try to isolate and/or debug it.

Thought: given the large amount of changes and endless combinations of elements in a greenshot file, I am a bit concerned that we won't find all issues during testing and produce a lot of broken/incomplete files. Maybe _for a start_ it would be a good idea to have the new format separated  (say, `.greenshot2`), so that using it is an explicit user choice? And then make it default version e.g. in Greenshot 1.5 🤔 

@Lakritzator you have been working on the .greenshot file format a lot more than me. I assume you don't have the time to review all the changes, which is totally okay 😅 but maybe you have an opinion on the general approach, apart from the implementation.

src/Greenshot.Editor/Drawing/ArrowContainer.cs
Christian-Schulz: yes, it was added twice.  Look 4 Lines after this. And field  `ARROWHEADS` is an enum.

src/Greenshot.Editor/FileFormat/Dto/Container/DrawableContainerDto.cs
Christian-Schulz: IntKeys are recommended.  
Info [here](https://github.com/MessagePack-CSharp/MessagePack-CSharp#object-serialization)

> If all keys are indexes, arrays are used for serialization, which offers advantages in performance and binary size. 

Keys must be unique. I thought that would be easier with int. 
And my main reason is that if the class evolves over time, you can't rename the StringKeys, and it could lead to annoying name conflicts with new properties later on.   

What I'm not so sure about is that I have specified rules for index numbering.
([0-10] for versioning, [11-99] for base class, and [100-199] for inherit class)
This creates some null entries in the binary data.

Christian-Schulz:
> Thought: given the large amount of changes and endless combinations of elements in a greenshot file, I am a bit concerned that we won't find all issues during testing and produce a lot of broken/incomplete files.

In the TestData folder i created greenshot files with every possible element/container. For the creation I used Greenshot v1.2.10 and Greenshot v1.3.290. 
The different elements in the Greenshot file interact very little with each other.  I do not see many combinations of elements as problematic. The new version can load all these old files.
_(Remark: Both old versions i had to fix locally for that, because it was not always possible to load some of there own saved element types)_
In the test project I test to save and reload every element. In [DrawableContainerListSerializationTests.cs](https://github.com/greenshot/greenshot/blob/c693c4f4a9e2e88e99bdfbca85d8d094582c29cd/src/Greenshot.Test/Editor/FileFormat/Dto/GreenshotFileV2/DrawableContainerListSerializationTests.cs) I test some special combinations.

src/Greenshot.Base/IniFile/IniConfig.cs
Christian-Schulz: Yes, I guess you have an old SDK-version installed (.NET 7 SDK or older). 
I use the current version .NET 9 SDK.

jklingen:
@Christian-Schulz sorry for the loooong wait 🙈 It was important for me to get @Lakritzator's opinion on this, as I knew he had already put some brainpower into the subject.

In the meantime, we have briefly discussed his ideas, the approach he had in mind is a bit different than yours. For a quick outline, it involves having a zipped file format (e.g. *.gsz)  that contains various JSON files (e.g., for annotation, metadata, etc.) in addition to the binary image files.

What I love about that idea is that it is super-transparent, which would be more open and make it easier to analyze and/or manually fix issues in existing files. Say, for example, a future nasty bug (from the past) created near-inifinite coordinates for an element in the file, causing a crash - a tech-savvy user would even have good chances of fixing it themselves.

What are your thoughts, @Christian-Schulz?

Of course, the solution proposed in here is also very good - also it is nearly finished 🤔 I guess both have their pros and cons. We just need to decide soon, as this pull request won't age very well (regarding merge conflicts).

In case we decide to go the zipped JSON way rather than MessagePack I think there's still a lot of stuff in this PR that we could use, e.g. handling of the legacy format, testing capabilities. But of course, I wouldn't expect you (@Christian-Schulz) to re-write the rest, you have already put in a lot of work.

Christian-Schulz:
As I had started the development, i analysed these options as a replacement for BinaryFormatter. 

- XML-Serializer 
  - I've hated it for almost 20 years. I've always had little strange problems with it, it leads to large file sizes and can also cause security problems when parsing. 😉 _(Maybe I'm biased)_
- JSON-Serializer
  - It serializes to text, but we have a few binary properties that would need to be converted to Base64.
- MessagePack
  -  Supports binary data and a large number of built-in standard types.
- Protobuf
  - Far too complex in my opinion.

Both JSON and MessagePack support options for evolving the DTO over time, polymorphism, Renaming properties, optional properties, deprecated properties, etc.
The main reason for the MessagePack serializer was the binary data, i.e., image in an ImageContainer.   

I was thinking about the approach with a ZIP file, as in the Open Document Format or NuGet packages too.
At the time, I found it too complex, as you had to implement everything yourself. 

What I really liked about the old file format was that it was actually a PNG file. It could be opened with any image viewer or browser.

I wanted to keep this feature. In this new version, even with all visible elements.
Yes, currently they are not 100% PNG-compliant. 
But with a newer version of ImageSharp, it would be possible to save our data in correct PNG-Chunks.

In the end, I see that a decision must be made on these two main arguments:

- human-readable file ?
- PNG file ?

Both solutions would be fine for me.

I implemented all in mind of that the file format could change over time.
To replace PNG/MessagePack with ZIP/JSON, only the attributes in the DTOs and the classes under `Greenshot.Editor.FileFormat.V2` needs to be changed and the test files would have to be recreated.

I noticed that you mentioned a different file extension: `.gsz`. Is there a specific reason for this?

Lakritzator:
Ah yeah, using a custom chunk in a .png was on my plan years ago, you can still see it here: https://github.com/SixLabors/ImageSharp/issues/251
You can see I linked it with a Greenshot issue, which I created to discuss the next format: https://github.com/greenshot/greenshot/issues/375

But the ImageSharp change was never finished, the issue is still open :/

The .gsz is just an idea, .greenshotzip...
It's like "comic book .zip" -> .CBZ

My goal was to keep the original images which were dragged etc onto the canvas, so we can edit them however it would be possible. So also SVG... With quality loss.

jklingen:
> What I really liked about the old file format was that it was actually a PNG file. It could be opened with any image viewer or browser.

Fair point. I had that completely off my radar. (Personally I mostly create one-time screenshots, so I never use .greenshot files.). As developer *and* user, I generelly prefer transparent file formats, but I can imagine that the PNG part is an important feature for quite some users, esp. when using Greenshot for documentation purposes.

Christian-Schulz:
> Personally I mostly create one-time screenshots, so I never use .greenshot files.

It's the same for me. I never use .greenshot files myself. 
I'm sure there are some users who do use .greenshot files.
But how many users actually use .greenshot files directly as images?

There is a preview function for Windows Explorer [see thumbnail-provider](https://learn.microsoft.com/en-us/windows/win32/shell/thumbnail-providers). It would be possible to develop an extension for Windows Explorer for our own file format. But I don't know if the effort involved is worth it.

Christian-Schulz:
> But how many users actually use .greenshot files directly as images?

I've thought about it again. With the current version, there can only be a few. With the current version, only the screenshot can be seen, all elements like lines, text, and effects are not visible on it.

I would now recommend the ZIP/JSON format.

That would be my first thought about the file structure in zip format.

```
my_file.gsa (Greenshot Archive)
├─ content.json        # Contains the actual content data 
├─ meta.json           # Contains metadata about the file/archive (file version, saved by Greenshot version ...)
├─ screenshot/         # Folder with the screenshot itself
│  ├─ screenshot.png    
│  └─ ...  			   # (maybe other formats in future e.g. for HDR content )
├─ images/             # Folder with image files
│  ├─ image1.png
│  ├─ image2.jpg
│  └─ ...
└─ thumbnails/         # Folder with preview/rendered image (maybe for later use)
   └─ preview.png
```

Christian-Schulz:
@Christian-Schulz not for me: 
- Since we have published the pre-release version v1.4, backward compatibility must be checked.
- Support the new Emoji-Container

jklingen:
> only the screenshot can be seen, all elements like lines, text, and effects are not visible on it.

Good point. The value of the preview is limited if it doesnt show what the file content looks like.

> I would now recommend the ZIP/JSON format.
> 
> That would be my first thought about the file structure in zip format.

Sounds like we're all team zipped JSON by now 😁 for me, the suggested file structure looks fine - and I think it's pretty much like what @Lakritzator had in mind?

Christian-Schulz:
Okay, I would start with a first attempt.

Lakritzator:
The weekend I was a bit occupied with arranging a birthday party, yesterday I had a work event, and at the moment I'm traveling to Berlin... so I'm a bit slow responding. Right now I have to read up and prepare some slides in the train, but I will think up some parts that are important to me.

- Zip to make it flexible and in the end produces one file, which doesn't need rocket science to extract
- Flexible, I can leave out parts or later add structured data, e.g. I was thinking that the thing we need to implement for the restart manager (updates which stop & start Greenshot) can write the whole editor state, also the undo-redo log, but in different files.
- The screenshot should be png by default, all the other image should be stored exactly as they were (could this cause issues?)
- Version support, so maybe we have a fixed file name for each version? (undo-redo_v1.json ?)
- The json screenshot information would contain all the information to put together the end result. Every annotation with it's important information. We can probably standardize the style information. I also like to add a layer concept, first we will have one.
- The would be something like OCR data later
- Also meta data, like time of capture, user, what window was active, location on the screen etc.

Some things that I thought of for the annotations, just some brainstorming:

- They have an order, how do we handle that? I was thinking about having the elements in their own "structure", so a list of "arrows!, everything having their ID.
- A layer with the elements order, if needed with an order id (1,2,3,4), and a link to the ID
- Style is standard, fill color, line color, shadow on/off etc.
- Text could be added as separate thing
- Location etc information is for the annotation type, not style.

Now I have to continue work

Lakritzator:
Oh, in the future I want to manage all captures in a document store (e.g. LiteDB), from which one can export. This will be the source for all the information, so saving a Greenshot file is more or less just an export from there. The structure is practically 100% the same, the ZIP just brings everything together.

Christian-Schulz:
FYI: This is still Work-In-Process.

The current version can save and load JSON/ZIP files. If you want, you can try it.

<details><summary>Sample content.json </summary>

```json
{
  "formatVersion": "V3",
  "schemaVersion": 1,
  "imagePath": "Images/image_1.png",
  "renderedImagePath": "Images/image_2.png",
  "containerList": {
    "containerList": [
      {
        "$type": "Line",
        "left": 240,
        "top": 175,
        "width": 37,
        "height": 40,
        "fields": [
          {
            "fieldTypeName": "LINE_THICKNESS",
            "value": {
              "$type": "Int",
              "value": 4
            },
            "scope": "LineContainer"
          },
          {
            "fieldTypeName": "LINE_COLOR",
            "value": {
              "$type": "Color",
              "color": "#FF0000FF"
            },
            "scope": "LineContainer"
          },
          {
            "fieldTypeName": "SHADOW",
            "value": {
              "$type": "Bool",
              "value": false
            },
            "scope": "LineContainer"
          }
        ]
      },
      {
        "$type": "Speechbubble",
        "text": null,
        "storedTargetGripperLocation": {
          "x": 45,
          "y": 312
        },
        "left": 58,
        "top": 281,
        "width": 75,
        "height": -30,
        "fields": [
          {
            "fieldTypeName": "LINE_THICKNESS",
            "value": {
              "$type": "Int",
              "value": 3
            },
            "scope": "SpeechbubbleContainer"
          },
          {
            "fieldTypeName": "LINE_COLOR",
            "value": {
              "$type": "Color",
              "color": "#FF0000FF"
            },
            "scope": "SpeechbubbleContainer"
          },
          {
            "fieldTypeName": "SHADOW",
            "value": {
              "$type": "Bool",
              "value": true
            },
            "scope": "SpeechbubbleContainer"
          },
          {
            "fieldTypeName": "FONT_ITALIC",
            "value": {
              "$type": "Bool",
              "value": false
            },
            "scope": "SpeechbubbleContainer"
          },
          {
            "fieldTypeName": "FONT_BOLD",
            "value": {
              "$type": "Bool",
              "value": false
            },
            "scope": "SpeechbubbleContainer"
          },
          {
            "fieldTypeName": "FILL_COLOR",
            "value": {
              "$type": "Color",
              "color": "#FFFFFFFF"
            },
            "scope": "SpeechbubbleContainer"
          },
          {
            "fieldTypeName": "FONT_FAMILY",
            "value": {
              "$type": "String",
              "value": "Arial"
            },
            "scope": "SpeechbubbleContainer"
          },
          {
            "fieldTypeName": "FONT_SIZE",
            "value": {
              "$type": "Single",
              "value": 20
            },
            "scope": "SpeechbubbleContainer"
          },
          {
            "fieldTypeName": "TEXT_HORIZONTAL_ALIGNMENT",
            "value": {
              "$type": "StringAlignment",
              "value": "Center"
            },
            "scope": "SpeechbubbleContainer"
          },
          {
            "fieldTypeName": "TEXT_VERTICAL_ALIGNMENT",
            "value": {
              "$type": "StringAlignment",
              "value": "Center"
            },
            "scope": "SpeechbubbleContainer"
          }
        ]
      },
      {
        "$type": "StepLbel",
        "number": 1,
        "counterStart": 1,
        "left": 164,
        "top": 248,
        "width": 30,
        "height": 30,
        "fields": [

```
</details>

JimBobSquarePants:
> Ah yeah, using a custom chunk in a .png was on my plan years ago, you can still see it here: [SixLabors/ImageSharp#251](https://github.com/SixLabors/ImageSharp/issues/251) You can see I linked it with a Greenshot issue, which I created to discuss the next format: #375
> 
> But the ImageSharp change was never finished, the issue is still open :/
> 
> The .gsz is just an idea, .greenshotzip... It's like "comic book .zip" -> .CBZ
> 
> My goal was to keep the original images which were dragged etc onto the canvas, so we can edit them however it would be possible. So also SVG... With quality loss.

@Lakritzator What chunks did you want to store this in? ImageSharp already writes custom data as text chunks. 

https://github.com/SixLabors/ImageSharp/blob/da6d56bc62ed90c26a4c37a7dbba55f6d90ddae1/src/ImageSharp/Formats/Png/PngEncoderCore.cs#L1111

I think that actually closes the requirements of the open issue.

Lakritzator:
Wait a bit with this, I have a huge change going on, but after that I will share some ideas...

Christian-Schulz:
🥲 It's mostly done, documatation and merge is left.

Lakritzator:
Monday I need to do an exam, and with all the family stuff going on, I unfortunately do not have time.
Besides I'm still struggling with #1000 but almost got most stuff working, after that I have time.

I have a prototype for this somewhere, maybe we can match experiences and ideas, somewhere next week.

Christian-Schulz:
I would say the PR is ready for the next review. 

Greenshot now serializes all data in JSON and saves it in a zip archive (`.gsa` file) .
The old `.greenshot` files can still be read but can no longer be saved.

Greenshot uses the same serialization mechanism for the clipboard. This uses JSON without a ZIP file.

General information about the new format can be found here: [Greenshot.Editor/FileFormat/readme.md](https://github.com/Christian-Schulz/greenshot/blob/feature/new_greenshot_file_format_v2/src/Greenshot.Editor/FileFormat/readme.md) .

Feel free to ask me questions.

Christian-Schulz:
In the current version 1.4  many changes have been added that conflict with this PR.

Mainly the test framework, CaptureHelper and ImageIO, as well as some minor confilcts in the drawable containers.

I think this will remain open for a while, so I'm changing its status to draft.

PR #644, opened 2025-07-28
Fix: Define plugin name for exclusion in INI

### LINKED ISSUE #642: ExcludePlugins INI setting no longer works
**Describe the bug**
We use the `ExcludePlugins` setting in the `greenshot-fixed.ini` as documented [here](https://getgreenshot.org/faq/how-remove-plugins-or-destinations-from-greenshot/):
`ExcludePlugins=Imgur Plugin` 

The setting is no longer applied with the most current version 1.3.296, while the exact same config works with 1.2.10.6.

**To Reproduce**
1. Add the line mentioned above to your greenshot(-fixed).ini ( I tried both, fixed and user ini, neither work)
2. Start Greenshot
3. Imgur Plugin (or any other) is loaded although excluded in the ini file.

**Expected behavior**
ExcludedPlugins set in INI are not loaded.

**Screenshots**

 - Greenshot version: 1.3.296
 - Windows version: Win10 19045.6093 & Win11 26100.4652 (tested both)

Christian-Schulz:
This fixes #642.

The list of plugin names, as described in the [FAQ](https://getgreenshot.org/faq/how-remove-plugins-or-destinations-from-greenshot/), is no longer available.

I have defined a new way that extracts the name from the plugin's sub-namespace.
This is the new current list:
```
"Box"
"Confluence"
"Dropbox"
"ExternalCommand"
"Flickr"
"GooglePhotos"
"Imgur"
"Jira"
"Office"
"Photobucket"
"Win10"
```
I intentionally omitted the suffix `“ Plugin”` as it was previously included.

### DIFF
--- a/src/Greenshot/Helpers/PluginHelper.cs
+++ b/src/Greenshot/Helpers/PluginHelper.cs
@@ -219,30 +219,14 @@ namespace Greenshot.Helpers
                 {
                     var assembly = Assembly.LoadFrom(pluginFile);
 
+                    if (IsPluginExcludedByConfig(assembly, pluginFile) ) 
+                    {
+                        continue;
+                    }
+                    
                     var assemblyName = assembly.GetName().Name;
-
                     var pluginEntryName = $"{assemblyName}.{assemblyName.Replace("Greenshot.Plugin.", string.Empty)}Plugin";
                     var pluginEntryType = assembly.GetType(pluginEntryName, false, true);
-
-                    // the sub namespace from plugin is used to include/exclude plugins
-                    var excludeIdentifier = assemblyName.Replace("Greenshot.Plugin.", string.Empty);
-
-                    if (CoreConfig.IncludePlugins is {} includePlugins
-                        && includePlugins.Count(p => !string.IsNullOrWhiteSpace(p)) > 0 // ignore empty entries i.e. a whitespace
-                        && !includePlugins.Contains(excludeIdentifier))
-                    {
-                        Log.WarnFormat("Include plugin list: {0}", string.Join(",", includePlugins));
-                        Log.WarnFormat("Skipping the not included plugin '{0}' with version {1} from {2}", excludeIdentifier, assembly.GetName().Version, pluginFile);
-                        continue;
-                    }
-
-                    if (CoreConfig.ExcludePlugins is { } excludePlugins
-                        && excludePlugins.Contains(excludeIdentifier))
-                    {
-                        Log.WarnFormat("Exclude plugin list: {0}", string.Join(",", excludePlugins));
-                        Log.WarnFormat("Skipping the excluded plugin '{0}' with version {1} from {2}", excludeIdentifier, assembly.GetName().Version, pluginFile);
-                        continue;
-                    }
 
                     var plugin = (IGreenshotPlugin) Activator.CreateInstance(pluginEntryType);
                     if (plugin != null)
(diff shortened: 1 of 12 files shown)


chrizzk:
Thanks! Personally I'd prefer to keep the " Plugin" suffix though to not have a breaking change. Current INI configurations will no longer work after this change.

Christian-Schulz:
@chrizzk Yes, I understand that. However, since half of the plugin names were changed/replaced between versions 1.2 and 1.3, there would be breaking changes anyway.
I thought it would make sense to switch to the namespace as an identifier now.

Christian-Schulz:
I added a fix for `IncludePlugins` #648 to this PR.

src/Greenshot/Helpers/PluginHelper.cs
```diff
@@ -224,10 +224,23 @@ public void LoadPlugins()
                     var pluginEntryName = $"{assemblyName}.{assemblyName.Replace("Greenshot.Plugin.", string.Empty)}Plugin";
                     var pluginEntryType = assembly.GetType(pluginEntryName, false, true);
 
-                    if (CoreConfig.ExcludePlugins != null && CoreConfig.ExcludePlugins.Contains(pluginEntryName))
+                    // the sub namespace from plugin is used to include/exclude plugins
+                    var excludeIdentifier = assemblyName.Replace("Greenshot.Plugin.", string.Empty);
```
jklingen: Great you restored the *in*clude functionality, too 👍 

nit-pick: this variable should get a neutral name now, e.g. `pluginIdentifier` or `pluginConfigIdentifier`.

src/Greenshot/Helpers/PluginHelper.cs
```diff
@@ -224,10 +224,23 @@ public void LoadPlugins()
                     var pluginEntryName = $"{assemblyName}.{assemblyName.Replace("Greenshot.Plugin.", string.Empty)}Plugin";
                     var pluginEntryType = assembly.GetType(pluginEntryName, false, true);
 
-                    if (CoreConfig.ExcludePlugins != null && CoreConfig.ExcludePlugins.Contains(pluginEntryName))
+                    // the sub namespace from plugin is used to include/exclude plugins
```
jklingen: As suggested in the comments, I'd prefer to keep the configuration backwards compatible *for now* (even if the config identifiers partially don't match with the product names.... I think we'll need a clean-up for the plugins in the near future anyway, so we could rename then and officially announce the breaking change in the release notes.)

Idea: maybe we should add a custom metadata attribute (e.g. `ConfigIdentifier` or similar) to each plugin and compare against that? This would also bring the advantage of decoupling the allowlist/blocklist functionality from technical identifiers, which could help prevent it from breaking again in the future. 🤔  What do you think?

Christian-Schulz: I think assembly attributes are a good solution. The advantage is that we can still perform the check before the plugins are instantiated.

Christian-Schulz:
I added assembly attribute for plugin identifier. 
A fallback to plugin namespace is still there, but that should never occur.

There are two new plugins: `"GooglePhotos Plugin"` and `"Win10 Plugin"`. Are these really new, or are they just new names for `"Picasa Web Plugin"` and `"OCR Plugin"`?
Should we use the new names or the old names as identifiers?

jklingen:
> I added assembly attribute for plugin identifier. A fallback to plugin namespace is still there, but that should never occur.

Thanks a lot, great idea. Also very that cool you extracted the check into a separate method which improves readability.

> There are two new plugins: `"GooglePhotos Plugin"` and `"Win10 Plugin"`. Are these really new, or are they just new names for `"Picasa Web Plugin"` and `"OCR Plugin"`? Should we use the new names or the old names as identifiers?

Good question. I needed to look this up in the history myself (I have never been involved in the plugins that much). 

* GooglePhotos has indeed been renamed from Picasa, so I'd suggest to keep the old name *for now*, for the sake of backwards compatibility.
* [Win10 was around in Greenshot 1.2](https://github.com/greenshot/greenshot/tree/release/1.2.10/GreenshotWin10Plugin) already, but obviously hasn't been documented in the [FAQ entry](https://getgreenshot.org/faq/how-remove-plugins-or-destinations-from-greenshot/). I suggest we use "Win10 Plugin" for consistency for now. (In future, we should probably revise whether the functionality should still be a plugin at all...)
* OCR Plugin has been removed with 26fe579d3147c228bc3b3786ae2af15bebabb149, probably because Win10Plugin already had its own OCR destination.

So with regards to backwards compatibility, we'd still have two breaking changes (Win10 added, OCR removed), but I think that's okay. My main concern about breaking the configuration (without official announcement) is about the plugins that can do uploads, e.g. Imgur.

jklingen:
Thanks so much for providing that fix ❤️ I tested this locally and it works like charm.

PR #653, opened 2025-08-05
Fix update interval check on app start

Christian-Schulz:
If the `UpdateCheckInterval` setting is set to more than 0 days, Greenshot ignores this value each time it is launched. It checks for new versions every start.
As long as Greenshot is running, the next check would then be correct after the interval.

I have added a calculation that determines the next possible check date. If this is in the future, no check is performed.


Lakritzator:
The general idea is good and I think the configuration is only saved after the update check is actually done, so it shouldn't cause any side effects.

Lakritzator (approved):
We should support to reduce every unneeded request to the outside.

PR #660, opened 2025-08-13
[Feature] Close all open editors.

Christian-Schulz:
I added a feature to close all open editors at once.
_This references [FEATURE-333](https://greenshot.atlassian.net/browse/FEATURE-333) and [FEATURE-880](https://greenshot.atlassian.net/browse/FEATURE-880)_

While pressing `CRTL` the closing action in file menu is changing to `Close all`.

Close_all_2

Every `ImageEditorForm` has a static list with all created editors.  So the close all action calls `close()` on every editor in this list.
The request to save on every editor may still occur if necessary.

I added some translations also.

### DIFF
--- a/src/Greenshot.Editor/Forms/ImageEditorForm.cs
+++ b/src/Greenshot.Editor/Forms/ImageEditorForm.cs
@@ -450,6 +450,10 @@ namespace Greenshot.Editor.Forms
         {
             ClearItems(fileStripMenuItem.DropDownItems);
 
+            // check if CTRL is pressed and toggle CloseMenu
+            bool ctrlActive = (ModifierKeys & Keys.Control) == Keys.Control;
+            ToggleCloseMenuItemAction(ctrlActive);
+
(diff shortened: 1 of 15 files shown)


Lakritzator:
Interesting idea, I'm not against it.

It is very user "unfriendly" in a way, that it is another option people can't find without knowing though. But at least it doesn't clutter up the UI. 🤷‍♂️

Christian-Schulz:
Other possible implementations would be:
 (b) - Two close buttons
 (c) - Change the default action for closing to “Close all” and only close the current window with ALT+F4 and [x].

jklingen:
I share @Lakritzator's thoughts: I'd prefer a second menu item over of a "secret" hotkey, also. (Yes the file menu is getting longer - in future, maybe we should consider moving the destinations to another menu? 🤔 but not now)

Christian-Schulz:
Changed it to the two button version.

image

PR #670, opened 2025-08-25
[Fix] Delay in context menu

### LINKED ISSUE #666: Menus have a delay.
**Describe the bug**
After taking a screenshot, the menu that opens has a lot of delay and lag when switching between options. Please fix this problem.

**Screenshots**

https://github.com/user-attachments/assets/495fe1f8-dad2-44f3-84d9-7f8d3556c6fd

 - Greenshot version : 1.3.300
 - Windows version : 11 24H2

Christian-Schulz:
In some context menus, there was a strange endless loop that was supposed to ensure that the menu was disposed correctly.

I have changed this behavior and now use the Closed-Event.
In addition, `dispose()` is scheduled in the UI thread to wait for the ClickHandler to complete.

We had the same issue in the destination picker menu and in the editor context menu. I correct both.
This fixes #666

### DIFF
--- a/src/Greenshot.Base/Core/AbstractDestination.cs
+++ b/src/Greenshot.Base/Core/AbstractDestination.cs
@@ -22,7 +22,6 @@
 using System;
 using System.Collections.Generic;
 using System.Drawing;
-using System.Threading;
 using System.Windows.Forms;
 using Dapplo.Windows.Common.Extensions;
 using Dapplo.Windows.Common.Structs;
@@ -184,6 +183,8 @@ namespace Greenshot.Base.Core
                 TopLevel = true
             };
 
+            menu.SetupAutoDispose();
+
             menu.Opening += (sender, args) =>
             {
                 // find the DPI settings for the screen where this is going to land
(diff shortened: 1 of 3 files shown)


Lakritzator:
Hey, I'm pretty sure I implemented it like that for a reason, although I'm not perfect.
I **think** this was needed due to the fact that the menu shouldn't close when someone clicks next to it.
The delay also might come from the fact that the menu entries are dynamic, and "submenu" entries are calculated while moving over them.

Christian-Schulz:
In the PickerMenu-Closing-Event” we keep track if closing is permitted or canceled.
During testing, I did not notice any differences in behavior in that way.

The calculation for the submenus is only triggered when you hover over an item with the mouse for x ms, but not when you quickly move over it.

I am pretty sure that only the `Sleep(100);` in the loop causes the delay.

The loop was added in this commit: e654c38eb808f87d0a593f28f358ce4122e74c5f

Lakritzator:
Well that WAS ages ago, wow.
Yeah, sounds sane, let's get this into the next build.

I will start doing more and more PR / issues management again, after being overwhelmed with work and private things...
Before I merge something, I need to get my bearings on our branches etc... been away way to long.
Also it's way past midnight, so good night.

ClubSpeedBogdan (approved):

PR #686, opened 2025-09-23
[Fix] Missing types in SafeSerializationBinder 

### LINKED ISSUE #685: Greenshot File Association and Launch with File Arguments Not Working When Already Running
**Describe the bug**
Greenshot can no longer be launched with file arguments (.greenshot, .png, or other formats) via file association or direct command line, which is problematic for users who have configured .png files to open with Greenshot by default.

**Current Behavior**
When Greenshot is already running: File association and command line file arguments fail to work
When Greenshot is not running: It is possible to open Greenshot with a file argument
Workaround available: Files can still be opened directly from within Greenshot using "Open With" functionality

**To Reproduce**
For .greenshot files :
Ensure Greenshot is already running in the system tray
Double-click on any .greenshot file or use command line : "C:\Program Files\Greenshot\Greenshot.EXE" --openfile "[FullPathToFile].greenshot"
Observe that the file does not open in Greenshot

**Expected behavior**
Greenshot should accept file arguments consistently regardless of whether the application is already running, maintaining backward compatibility with file associations and direct command execution.

 - Greenshot version 1.3.301
 - Windows 11 version 24H2

**Additional context**
The file associations in the Windows registry are correctly set for Greenshot.
This behavior change appears as a regression related to the recent security vulnerability fixes. Despite correct registry settings, launching Greenshot with file arguments fails if the application is already running. The functionality to open files directly from within Greenshot ("Open With") remains unaffected.

Christian-Schulz:
Added all missing types to `SafeSerializationBinder` that are allowed to be sent with WM_COPYDATA.

This fixes #685

### DIFF
--- a/src/Greenshot/Helpers/CopyData.cs
+++ b/src/Greenshot/Helpers/CopyData.cs
@@ -608,6 +608,10 @@ internal class SafeSerializationBinder : SerializationBinder
     private static readonly Type[] AllowedTypes = new Type[]
     {
         typeof(CopyDataObjectData),
+        typeof(CopyDataTransport),
+        typeof(KeyValuePair<CommandEnum, string>),
+        typeof(List<KeyValuePair<CommandEnum, string>>),
+        typeof(CommandEnum),
     };
 
     public override Type BindToType(string assemblyName, string typeName)


jklingen:
Ooops, I missed that `object` type 🙈 

Once again, thanks for analyzing this and providing a fix so fast. 🚀

Lakritzator:
I guess there is still something missing, see #872

`[SafeSerializationBinder] Unexpected type received via WM_COPYDATA, a malicious process might be trying to attack. Cancelling deserialization. Suspicious type: Greenshot, Version=1.3.0.0, Culture=neutral, PublicKeyToken=null - Greenshot.Helpers.CopyDataTransport
`

Lakritzator:
It looks like it's a version mismatch, the type is registered but it's just a different version.... which might mean that the user has 2 versions on his system...

Lakritzator:
I fixed this by removing the version information before comparing, the deeper issue of having multiple versions is another thing.

PR #688, opened 2025-09-26
open Preferences page instead of showing error when running second in…

JustinCredibuilt:
…stance

instead of displaying an error "an instance of Greenshot is already running", let's show the Preferences page instead. I hear many questions about that error and I think this may be a better user experience. It's a big enough issue you've even got an entry for it in the FAQ. Open to discussion of course!

### DIFF
--- a/src/Greenshot/Forms/MainForm.cs
+++ b/src/Greenshot/Forms/MainForm.cs
@@ -273,19 +273,10 @@ namespace Greenshot.Forms
                             using Process currentProcess = Process.GetCurrentProcess();
                             instanceInfo.Append(index + ": ").AppendLine(Kernel32Api.GetProcessPath(currentProcess.Id));
                         }
-
-                        // A dirty fix to make sure the message box is visible as a Greenshot window on the taskbar
-                        using Form dummyForm = new Form
+                        using (var settingsForm = new SettingsForm())
                         {
-                            Icon = GreenshotResources.GetGreenshotIcon(),
-                            ShowInTaskbar = true,
-                            FormBorderStyle = FormBorderStyle.None,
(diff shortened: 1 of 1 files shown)


jklingen:
Thanks for your suggestion 👍 We can discuss about alternative options, I just won't merge it as it currently is, for two reasons: 

* you'd need to sign the CLA - in case you wonder what this is about, you might want to have a look at [our FAQ entry about Contributor License Agreement](https://getgreenshot.org/faq/what-is-a-cla-and-who-has-to-sign-it/)
* I am not convinced that the suggested implementation is good for most users. What would the user want to do in the settings dialog in that situation? 

Some thoughts:
 
I can think of two scenarios how users run into this situation:
  * The user wants to start Greenshot and is unaware that the same Greenshot executable is already running in the background.
  * The user wants to start Greenshot and is unaware that another Greenshot executable is running in the background (e.g. trying to run the ZIP release from Downloads directory, while the installation from Programs directory is already running.)

The current solution targets the 2nd scenario (it helps them to find out that the other running instance is another executable), it can also help in the 1st (if the user happens to understand what it is about), but I agree that it's probably confusing for most users running into it.

I think a more user friendly variant for the 1st scenario would be to display a notification at the systray icon, which says something like "Greenshot is already running in the background, right-click the icon or use it."

Lakritzator:
I do not like it when we keep PRs open for to long, especially if we can say this is not the way we want to go.
This does hurt me a bit though, we appreciate every PR that comes in, even small ones...
So I do want to thank you for your investment in our community.

But I agree with Jens, and this PR doesn't work like the described way.
I also have to add that the issue of "an instance of Greenshot is already running" has been addressed and should hardly happen anymore.

PR #689, opened 2025-09-28
Update language-tr-TR.xml

xqtp:
updated with latest strings from en_us file, also translation errors fixed


jklingen:
Thanks a lot 👍

PR #691, opened 2025-09-30
Update installer/additional_files/readme.txt

jonadem:


jklingen:
Thanks for submitting this, we have updated this change log entry in a subsequent commit: https://github.com/greenshot/greenshot/tree/main/installer/additional_files

The issue is that we don't know the build number until we actually build the release, that's why we have xxx in there for the upcoming release. We should probably use a placeholder and have it automatically replaced during the build process.

PR #712, opened 2025-11-13
[Fix] compiler error CS9273 in C#14

Christian-Schulz:
Building Greenshot with current .NET SDK and C# 14 causes compiler error [CS9273](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/partial-declarations?f1url=%3FappId%3Droslyn%26k%3Dk(CS9273)#field-backed-properties).

`field` is now a reserved keyword.

To fix this, the variable is renamed to `fileInfo`.


Lakritzator:
I actually made the same change today while I tripped over this error 😆

Christian-Schulz:
> I actually made the same change today while I tripped over this error 😆

Yeah, I saw it in the other issue. You choose the same variable name as a replacement.

PR #730, opened 2025-12-01
Add setting for Effect > Border in Greenshot.ini

weihongji:
Add setting for Effect > Border in Greenshot.ini. Details:
```ini
[Editor]
BorderEffectSettings=Key:BorderEffect|Color:203,203,203|Width:1
```

Add setting "UseBestFit" to control the auto zoom-in behavior when "Open image from clipboard"
```ini
[Editor]
UseBestFit=False
```


Lakritzator:
Nice work.
Btw: I usually prefer to have 1 change in an external PR, but at least you documented it.

Issue #735, opened 2025-12-05
After editing on save re copy to the clipboard automatically option instead of extra click


danrhodes:
This GitHub repository is for developers, if you want to request a feature for Greenshot as a user please do so in our JIRA issue system here: https://greenshot.atlassian.net/projects/FEATURE/issues/filter=allopenissues

**Is your feature request related to a problem? Please describe.**
I use greenshot daily hindered of times a day to take screenshots for our service desk tickets and I often draw boxes of annotate also. When saving I also have to click copy to clipboard. I know it's only 1 extra click but when I might do 500 screen shots a day it adds up.

**Describe the solution you'd like**
I would like an options to auto copy back to the clipboard on save

**Describe alternatives you've considered**
Keep clicking the extra button

**Additional context**
Greenshot rules!

danrhodes:
This is implemented via https://github.com/greenshot/greenshot/pull/736

maniootek:
+1 for this. I'd like to suggest a slightly broader version of the same idea: an option to keep the clipboard continuously in sync with the editor — i.e. after every edit action (arrow, box, crop, highlight, etc.), the edited image replaces what's on the clipboard, not only at save time.
This is how the built-in Windows Snipping Tool behaves: every edit is reflected on the clipboard automatically, so you can paste the current result at any moment without an extra copy step. I recently switched from Snipping Tool to Greenshot and this is the one workflow difference I miss.
Ideally this would be an opt-in setting (e.g. "Automatically copy edits to the clipboard"), since continuously overwriting the clipboard may not suit everyone. The on-save re-copy suggested here would already help; live sync would cover it fully.
I'd be willing to help implement this via a PR if the maintainers agree on the approach.

PR #736, opened 2025-12-05
Add option to recopy image to clipboard on editor save

danrhodes:
Introduces a new configuration setting 'EditorRecopyToClipboardOnSave' that, when enabled, automatically copies the image to the clipboard upon saving from the editor. Updates the settings UI to allow users to toggle this feature.

I raised an issue for this but had time to solve it also.


danrhodes:
Greenshot rules!

danrhodes:
Forgot to add also I have built the src in VS and it builds with just a few pre existing errors re confluence plugin.

danrhodes:
image

Christian-Schulz:
Nice feature, but is there a conflict with this feature?
image

danrhodes:
Likely if the option is selected yes. Easiest way would be to let the user choose, as each serves different end cases or I could amend to automatically untick that option if the new one was selected, I've never really understood for my workflow at least why you would want the text path and not the image? Views?

danrhodes:
I think the solution to this problem is slightly more complicated but to avoid having to programmatically adjust settings on the fly when the other is selected I will propose to swap the tickbox on the output tab for a dropdown box with the below options. This will provide a solid ground for users to choose their preferences with no ambiguity.

Clipboard behavior after save:
○ Copy image data
○ Copy file path
○ None (clipboard unchanged)

Christian-Schulz:
That's the same solution I wanted to write down this morning. You were quicker.
I would describe it as more general.

`PostSavingAction `- “Behavior after saving:”
- Copy image to clipboard 
- Copy file path to clipboard 
- None

A little logic should also be added to transfer the old configuration to the new one during the first start.

danrhodes:
The logic I 100% agree and can easily be added in. There is nothing worse than your preferences being overwritten. I'll also amend the wording to reflect your suggestion. I should have time to do this tomorrow and will push my code.

“Behavior after saving:”

- Copy image to clipboard
- Copy file path to clipboard
- None

Thanks for the feedback 👍

danrhodes:
Hey,

Previous changes have been reverted.

image

I have swapped the tickbox on Output for a dropdown and applied logic to keep user preferences intact.

image

I have checked all options, including the logic, to ensure they work.

Build passes with just warnings on the Greenshot.Plugin.Confluence plugin which is not associated with these changes.

danrhodes:
OCD having the additional : on the text made me tick, I have removed this to ensure consistency.

src/Greenshot.Base/Core/CoreConfiguration.cs
```diff
@@ -472,6 +472,15 @@ public override string PreCheckValue(string propertyName, string propertyValue)
...
 
+            // Migrate old OutputFileCopyPathToClipboard boolean to new PostSaveBehavior enum
+            if ("OutputFileCopyPathToClipboard".Equals(propertyName))
+            {
+                if (bool.TryParse(propertyValue, out bool copyPathToClipboard))
+                {
+                    return copyPathToClipboard ? "CopyFilePathToClipboard" : "None";
+                }
+            }
+
```
Christian-Schulz: I don't think this will work.
We have to read the old property and set the new property. 

I don't know if we have a feature to mark setting as deprecated.
Maybe with `IniPropertyAttribute.ExcludeIfNull` and setting the old property to `null` in `AfterLoad()`.
So that it could be read, but never written. You could give it a try.

src/Greenshot.Base/Core/CoreConfiguration.cs
```diff
@@ -547,11 +556,8 @@ public override void AfterLoad()
                 OutputDestinations.Add("Editor");
             }
 
-            // Prevent both settings at once, bug #3435056
-            if (OutputDestinations.Contains("Clipboard") && OutputFileCopyPathToClipboard)
-            {
-                OutputFileCopyPathToClipboard = false;
-            }
+            // Note: OutputFilePostSaveBehavior is independent of clipboard destination
+            // since it controls editor save behavior, not initial capture behavior
```
Christian-Schulz: This incompatibility between the different `OutputDestinations` still exists. The last `OutputDestination` that copies something to the clipboard wins. 

I.e. the `ImgurDestination` copies the URL to the clipboard.

I would say there is no solution and the comment can be removed here.

danrhodes:
I will take a look tomorrow, or maybe tonight if I get a spare hour 👍

danrhodes:
I have made the requested changes :-)

danrhodes:
Ive had to add a little more logic and a flag for migration as the migration logic was running each time GS loaded leading to "Copy path to clipboard on save" being set each time.

src/Greenshot.Base/Core/CoreConfiguration.cs
```diff
@@ -496,6 +502,22 @@ public override void BeforeSave()
         /// </summary>
         public override void AfterLoad()
         {
+            // Migrate old OutputFileCopyPathToClipboard boolean to new PostSaveBehavior enum
+            // Only run migration once - check if we've already migrated
+            if (!OutputFilePostSaveBehaviorMigrated.HasValue || !OutputFilePostSaveBehaviorMigrated.Value)
+            {
+                if (OutputFileCopyPathToClipboard.HasValue)
```
Christian-Schulz: Why is `OutputFilePostSaveBehaviorMigrated` needed? I thought that as long as `OutputFileCopyPathToClipboard` has a value, it hasn't been migrated yet.

danrhodes: I will take a further look today as the migration logic wasn't working and was reverting back to copy path each time GS was closed/opened and setting a new value in the ini file. Setting a migrated flag sorted this but it shouldn't be needed as you say so I will work it out and update the code instead of using a marker.

danrhodes: Made the changes you suggested - removed the OutputFilePostSaveBehaviorMigrated flag as it wasnt needed as you said. The migration now just checks if OutputFileCopyPathToClipboard.HasValue, and if it does then it migrates the value and sets the old property to null. With ExcludeIfNull on the old property it wont get written back to the ini file so on the next load it should be null and the migration wont run again. Removed the default value from the old property too so it only has a value if its actually in the ini file. This should work now without the extra flag cluttering things up

danrhodes:
Hey, I have made the changes as suggested. Tested and working well👍

danrhodes:
Linking the issue - https://github.com/greenshot/greenshot/issues/735

Christian-Schulz:
I debugged your solution a bit.
It works as it should, but only for the save button in the toolbar.

It does not work completely with the menu items “save as” and “save directly”. 
These menu items refer to `FileDestination` and `FileWithDialogDestination`.
Both only support the 2 options “Copy file path to Clipboard” and “None”. 

What do you think of the following?
Saving the file path to the clipboard is currently performed in `ImageIO.Save(... , bool copyPathToClipboard)`.
We are moving the functionality from `ImageIO`  to `File...Destination`.
We could remove the parameter `copyPathToClipboard`.
We then process all 3 options at the end of functions `FileDestination.ExportCapture()` and `FileWithDialogDestination.ExportCapture()`
Removing the call of `ClipboardDestination` in `BtnSaveClick()`.

That should then be the solution for all these cases:
- save button in the toolbar
- menu items “save as” and “save directly” 
- automatically executed destination “save as” and “save directly” 
- dynamically selected destination “save as” and “save directly” 

Tell me if you don't like something about the solution.

danrhodes:
I like the solution and feel very foolish for not checking the save options from the drop down menu myself so thank for you capturing this 🙏

I will update the code later to reflect the solution you specify 👍

danrhodes:
I've moved the clipboard logic exactly where you suggested, Chris.

Changes made:

- Removed clipboard logic from ImageEditorForm.BtnSaveClick()
- Removed copyPathToClipboard parameter from ImageIO.Save()
- Added post-save clipboard handling at the end of FileDestination.ExportCapture() (after successful save, before ProcessExport)
- Added the same logic to FileWithDialogDestination.ExportCapture()

I've built it in VS, and all is working.

Now handles all these cases:

- save button in the toolbar
- menu items “save as” and “save directly”
- automatically executed destination “save as” and “save directly”
- dynamically selected destination “save as” and “save directly”

src/Greenshot.Editor/Forms/ImageEditorForm.cs
```diff
@@ -37,6 +37,7 @@
 using Greenshot.Base;
 using Greenshot.Base.Controls;
 using Greenshot.Base.Core;
+using Greenshot.Base.Core.Enums;
```
Christian-Schulz: not needed anymore

src/Greenshot.Editor/Forms/ImageEditorForm.cs
```diff
@@ -60,6 +61,7 @@ public partial class ImageEditorForm : EditorForm, IImageEditor
     {
         private static readonly ILog Log = LogManager.GetLogger(typeof(ImageEditorForm));
         private static readonly EditorConfiguration EditorConfiguration = IniConfig.GetIniSection<EditorConfiguration>();
+        private static readonly CoreConfiguration CoreConfiguration = IniConfig.GetIniSection<CoreConfiguration>();
```
Christian-Schulz, line 65: not needed anymore

src/Greenshot.Base/Core/CoreConfiguration.cs
```diff
@@ -496,6 +499,17 @@ public override void BeforeSave()
         /// </summary>
         public override void AfterLoad()
         {
+            // Migrate old OutputFileCopyPathToClipboard boolean to new PostSaveBehavior enum
```
Christian-Schulz, line 490:
```suggestion
#pragma warning disable CS0618 // Type or member is obsolete
            // Migrate old OutputFileCopyPathToClipboard boolean to new PostSaveBehavior enum
```

src/Greenshot.Base/Core/CoreConfiguration.cs
```diff
@@ -496,6 +499,17 @@ public override void BeforeSave()
...
+            // Once migrated, the old property is set to null and won't exist in the INI file anymore
+            if (OutputFileCopyPathToClipboard.HasValue)
+            {
+                OutputFilePostSaveBehavior = OutputFileCopyPathToClipboard.Value
+                    ? PostSaveBehavior.CopyFilePathToClipboard
+                    : PostSaveBehavior.None;
+                // Set to null so it won't be written back to the INI file (ExcludeIfNull)
+                OutputFileCopyPathToClipboard = null;
```
Christian-Schulz, line 498:
```suggestion
                OutputFileCopyPathToClipboard = null;
                IsDirty = true;
```

To force a early saving after reading.

src/Greenshot.Base/Core/CoreConfiguration.cs
```diff
@@ -496,6 +499,17 @@ public override void BeforeSave()
...
+            }
```
Christian-Schulz, line 500:
```suggestion
            }
#pragma warning restore CS0618 // Type or member is obsolete
```

Christian-Schulz (approved):
I have tested different scenarios. 
It looks good to me. 👍 

I have only left minimal suggestions for adjustments.

danrhodes:
Committed the three you suggested and will make the other changes later this evening.

👍

src/Greenshot.Editor/Forms/ImageEditorForm.cs
danrhodes: Removed after review agreed no longer needed

src/Greenshot.Editor/Forms/ImageEditorForm.cs
danrhodes: Removed after review.

danrhodes:
All sorted, I hadn't seen that the two lines were no longer needed, so thanks for catching that 👌

src/Greenshot.Base/Core/CoreConfiguration.cs
```diff
@@ -129,8 +129,11 @@ public class CoreConfiguration : IniSection, INotifyPropertyChanged
         [IniProperty("OutputFileReduceColorsTo", Description = "Amount of colors to reduce to, when reducing", DefaultValue = "256")]
         public int OutputFileReduceColorsTo { get; set; }
 
-        [IniProperty("OutputFileCopyPathToClipboard", Description = "When saving a screenshot, copy the path to the clipboard?", DefaultValue = "true")]
-        public bool OutputFileCopyPathToClipboard { get; set; }
+        [IniProperty("OutputFileCopyPathToClipboard", Description = "When saving a screenshot, copy the path to the clipboard?", ExcludeIfNull = true)]
```
Christian-Schulz, line 133:
```suggestion
        [Obsolete("Exists only for backward compatibility")]
        [IniProperty("OutputFileCopyPathToClipboard", Description = "When saving a screenshot, copy the path to the clipboard?", ExcludeIfNull = true)]
```

Christian-Schulz (approved):
I'm sure GitHub didn't save this suggestion last time.

danrhodes:
It's been a good process this @Christian-Schulz thanks for you valued input 👌👍🙏

Lakritzator (approved):

Lakritzator:
I was a bit sceptical at first, changing the configuration UI, or rather add settings is my horror scenario. The pixel real estate in the configuration UI is small...

But after I read "post actions" I was sold, the current solution is confusing and that might make it clearer. I will need to take it for a spin though, code looks good but in the end it's something I have to experience... (I review with a tablet 🤷‍♂️)

It's also cool to see a joint venture 😄

One thing though, my horror is getting a huge PR which I have to deny, so I hope before the next big one, you ping the idea in an issue??

danrhodes:
Appreciate the feedback @Lakritzator its been good working with @Christian-Schulz to get this done.

I did put this as an issue first but had some time on my hands due to holiday leave from work to create the PR too - https://github.com/greenshot/greenshot/issues/735

Next time I'll let it ripen in the issues list as a new feature prior to going full steam ahead 🙈

Fingers crossed you like the flow of the new feature 🤞🙏 (I've got an office full of engineers waiting to reduce their daily click count 🤣)

❤️ for Greenshot as ever!

Lakritzator:
I was going to test this, when I hit: https://github.com/greenshot/greenshot/issues/758

I did manage to get to the settings before that, and at the moment I was reading them, I was wondering what "after saving" would mean to our users?

danrhodes:
I'm happy to change it to be more descriptive.

- Post-save action {None} / {copy image to clipboard} / {copy path to clipboard}

- On save {do nothing} / {copy image to clipboard} / {copy path to clipboard}

- When saved {do nothing} / {copy image to clipboard} / {copy path to clipboard}

danrhodes:
@Lakritzator any views on what this should be changed to? 🙏

Lakritzator (approved):

danrhodes:
Is there anything I can do to get this change pushed through and over any final hurdles? I'd love for it to be in the next release I think it will be super useful. 🙏

danrhodes:
I've merged any recent changes back into this PR to keep it current 👍

danrhodes:
@Lakritzator how are we fixed for merging this for release in v1.4?

Lakritzator:

danrhodes:
I need to get back doing some Greenshot fixes, I've been away too long, just so busy with life and kids 🤷‍♂️

Issue #725, opened 2025-11-23
`Highlight area` tool causes freezing/slowdown in the Editor


96273:
- Take a full-screen screenshot and open it in the editor.
- Use the `Highlight` > `Highlight area` tool.
- Scroll up and down with the middle mouse wheel.
- Scrolling is not smooth, causing stuttering/lag.

Video:
<details><summary>Screenshot</summary>

https://github.com/user-attachments/assets/acd5f450-ddb4-4da3-b5ee-c2bbb80d5f2c

</details>

---

Additionally, if you repeatedly click the `Highlight text` and `Highlight area` options, Greenshot will give an error.

Generally, if I repeat the steps below, I can reproduce the error, but this method does not always work.

- First, `Highlight text`
- Then click on the `Highlight area`
- Next, click on the `Undo` button above.
- Then click on the `Highlight area` button again.

Video:

<details><summary>Screenshot</summary>

https://github.com/user-attachments/assets/4ab817e7-a5e9-4ed9-8c7f-14b5275720dc

</details>

Error:
<details><summary>Screenshot</summary>

```

Software version: 1.3.301 - f5a29a2ed3 (64 bit)
.NET runtime version: 4.0.30319.42000+
Time: 2025-11-23 05:08:31 +03:00
OS: Windows 10 2009 x64 build 19045
GDI object count: 141
User object count: 86

Exception: System.MemberAccessException
Message: Could not set property 'Value' to 'TEXT_HIGHTLIGHT' [PreparedFilter] on [Field FieldType=PREPARED_FILTER_HIGHLIGHT Value= Scope=FieldAggregator]. Probably other type than expected, IBindingCoverter to the rescue.

Stack:
   at Greenshot.Editor.Drawing.Fields.Binding.BidirectionalBinding.Synchronize(INotifyPropertyChanged sourceObject, String sourceProperty, INotifyPropertyChanged targetObject, String targetProperty)
   at Greenshot.Editor.Drawing.Fields.Binding.BidirectionalBinding.ControlPropertyChanged(Object sender, PropertyChangedEventArgs e)
   at Greenshot.Editor.Controls.BindableToolStripDropDownButton.OnDropDownItemClicked(ToolStripItemClickedEventArgs e)
   at System.Windows.Forms.ToolStrip.OnItemClicked(ToolStripItemClickedEventArgs e)
   at System.Windows.Forms.ToolStripDropDown.OnItemClicked(ToolStripItemClickedEventArgs e)
[...]

PR #738, opened 2025-12-09
Add null check for prepared filter in HighlightContainer fix issue 725 crash

danrhodes:
Added a guard clause to handle null values returned by GetFieldValue for PREPARED_FILTER_HIGHLIGHT, preventing potential errors after undo operations.

Fix for crash in https://github.com/greenshot/greenshot/issues/725

### DIFF
--- a/src/Greenshot.Editor/Drawing/HighlightContainer.cs
+++ b/src/Greenshot.Editor/Drawing/HighlightContainer.cs
@@ -75,7 +75,15 @@ namespace Greenshot.Editor.Drawing
 
         private void ConfigurePreparedFilters()
         {
-            PreparedFilter preset = (PreparedFilter) GetFieldValue(FieldType.PREPARED_FILTER_HIGHLIGHT);
+            object fieldValue = GetFieldValue(FieldType.PREPARED_FILTER_HIGHLIGHT);
+
+            // Guard against null value which can occur after undo operations
+            if (fieldValue == null)
+            {
+                return;
+            }
+
+            PreparedFilter preset = (PreparedFilter)fieldValue;
             while (Filters.Count > 0)
             {
                 Remove(Filters[0]);


danrhodes:
Tested this morning and confirmed the fix works.

src/Greenshot.Editor/Drawing/HighlightContainer.cs
```diff
@@ -75,7 +75,15 @@ protected void HighlightContainer_OnFieldChanged(object sender, FieldChangedEven
...
-            PreparedFilter preset = (PreparedFilter) GetFieldValue(FieldType.PREPARED_FILTER_HIGHLIGHT);
+            object fieldValue = GetFieldValue(FieldType.PREPARED_FILTER_HIGHLIGHT);
+
+            // Guard against null value which can occur after undo operations
+            if (fieldValue == null)
+            {
+                return;
+            }
+
+            PreparedFilter preset = (PreparedFilter)fieldValue;
```
Lakritzator, line 86: This also works, and is a bit shorter:
PreparedFilter preset = GetFieldValue(FieldType.PREPARED_FILTER_HIGHLIGHT) as PreparedFilter;
if (fieldValue == null)
{
    return;
}

Lakritzator:
Thanks for providing this, I still wonder why it happens though...

danrhodes:
My understanding from what I found

The code tries to grab something (`PreparedFilter`) and immediately use it.

After you undo something, that storage is empty (null). The code grabs nothing, then tries to use nothing, which crashes.

The fix is toheck if the storage spot is empty first. If it’s empty, just stop and don’t try to use it.

It’s like reaching into a jar expecting a sweet, then trying to eat whatever you grabbed—even if your hand is empty. The fix is to look at your hand first, and if it’s empty, don’t try to eat.​​​​​​​​​​​​​​​​

PR #742, opened 2025-12-14
Added chocolatey push support for releases

Lakritzator:
Added chocolatey push support for releases, this ensures that the chocolatey package is updated at the same time as we release a new version.

### DIFF
--- a/.github/workflows/choco-publish.yml
+++ b/.github/workflows/choco-publish.yml
@@ -62,9 +62,10 @@ jobs:
       - name: Pack with cpack
         run: |
           cd chocolatey
-          cpack greenshot.nuspec
+          choco pack greenshot.nuspec
 
       # 5. Push package to Chocolatey
       - name: Push package to Chocolatey
         run: |
           choco push *.nupkg --source https://push.chocolatey.org/ --api-key ${{ secrets.CHOCO_API_KEY }}
+


Lakritzator:
This should solve issues like https://github.com/greenshot/greenshot-chocolatey/pull/4

Lakritzator:
I find it hard to test things like this offline, I did as much work as I could upfront...

.github/workflows/choco-publish.yml
```diff
@@ -0,0 +1,54 @@
...
+          $url = "https://api.github.com/repos/greenshot/greenshot/releases/tags/${{ github.event.release.tag_name }}"
+          $release = Invoke-RestMethod -Uri $url
+          $asset = $release.assets | Where-Object { $_.name -like 'Greenshot-INSTALLER-*.exe' }
+          if (-not $asset) { throw "No matching asset found!" }
+          $checksumType, $checksum = $asset.digest -split ":"
+
+          # Inject release notes into greenshot.nuspec
+          (Get-Content chocolatey/greenshot.nuspec) -replace '<version>.*</version>', "<version>$version</version>" | Set-Content chocolatey/greenshot.nuspec
+          # Inject release notes into greenshot.nuspec
+          (Get-Content chocolatey/greenshot.nuspec) -replace 'RELEASE_NOTES', $release.body | Set-Content chocolatey/greenshot.nuspec
```
jklingen: Thought: we should probably XML escape the release notes, just to be sure.

.github/workflows/choco-publish.yml
```diff
@@ -0,0 +1,54 @@
+name: Pack and Publish Greenshot to Chocolatey
+
+on:
+  release:
+    types: [published] # Only runs on full releases
```
jklingen, line 5: Question: as far as I understand, this also runs on (unstable) pre-releases, is this desired? If not, I think we'd need to switch to `release`.

jklingen, line 5: Thought: Maybe add `workflow_dispatch`? It would probably be good to be able to trigger this manually if needed.

chocolatey/greenshot.nuspec.template
```diff
@@ -0,0 +1,32 @@
...
+  <metadata>
+    <id>greenshot</id>
+    <version>version</version>
+    <packageSourceUrl>https://github.com/greenshot/greenshot</packageSourceUrl>
+    <owners>Greenshot</owners>
+    <title>Greenshot (Install)</title>
+    <authors>Greenshot</authors>
+    <projectUrl>https://getgreenshot.org/</projectUrl>
+    <iconUrl>https://upload.wikimedia.org/wikipedia/commons/9/95/GreenshotLogo.png</iconUrl>
+    <licenseUrl>https://www.gnu.org/licenses/gpl.html</licenseUrl>
```
jklingen, line 12: Thought: maybe better reference the license file in our repo, in case we ever change it?
https://github.com/greenshot/greenshot/blob/main/LICENSE

chocolatey/greenshot.nuspec.template
```diff
@@ -0,0 +1,32 @@
...
+<package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
+  <metadata>
+    <id>greenshot</id>
+    <version>version</version>
+    <packageSourceUrl>https://github.com/greenshot/greenshot</packageSourceUrl>
+    <owners>Greenshot</owners>
+    <title>Greenshot (Install)</title>
+    <authors>Greenshot</authors>
+    <projectUrl>https://getgreenshot.org/</projectUrl>
+    <iconUrl>https://upload.wikimedia.org/wikipedia/commons/9/95/GreenshotLogo.png</iconUrl>
```
jklingen, line 11: `iconUrl` is deprecated, we should use `icon` and provide a 128x128 PNG within the package.
https://learn.microsoft.com/en-us/nuget/reference/nuspec#iconurl

chocolatey/greenshot.nuspec.template
jklingen, line 1: Thought: maybe rename file to `greenshot.nuspec.template` and let the workflow create a copy before updating its contents. Just to clarify that it is updated automatically.

Also applies to powershell scripts with placeholders.

chocolatey/greenshot.nuspec.template
Lakritzator: Yeah makes sense!

.github/workflows/choco-publish.yml
Lakritzator: I thought about that, and left it out, as I do not think the context / environment would have all to information so it works if it's triggered manually. The code needs a version which is released.

Lakritzator: I read somewhere it didn't but I actually do not know... Let's check

chocolatey/greenshot.nuspec.template
Lakritzator: Do we have one available?

chocolatey/greenshot.nuspec.template
Lakritzator: Makes sense, I can fix that.

.github/workflows/choco-publish.yml
Lakritzator: Yes 👍

chocolatey/greenshot.nuspec.template
Lakritzator: Actually, although it does make sense, I can imagine some process looking for the exact string... I rather leave it for now. Besides even if we ever would change the license, all versions before still have that license, so it wouldn't be correct.

.github/workflows/choco-publish.yml
Lakritzator: I think I got it right, but I haven't been able to test this 🤔

Lakritzator:
I actually was able to test most of it

Lakritzator:
The [published] makes sure it's only run when we release a version

Issue #743, opened 2025-12-16
PNG MIME Type mismatch


joey4good:
Hi,

I think there is something wrong with the MIME format for PNG.
There seems to be some kind of mismatch.
When I upload a screenshot taken with Greenshot in PNG format, the web app (ServiceNow) doesn't show the preview – I can only download the file.

If I take the screenshot with a different tool, like the Snipping Tool, it works fine.
It also works if I save the screenshot in JPG format in Greenshot and upload it to ServiceNow.
Only PNG is not working correctly.

I opened the corrupt PNG file taken with Greenshot in MS Paint and saved it as a PNG file using "save as".
After that, the file was recognised correctly in ServiceNow and showed the preview. Thats why I think, the issue is related to Greenshot and probably the MIME type.

As a workaround, I can still use JPG, but I would prefer to use PNG.
Will there be a solution to this?

danrhodes:
**Changes Made**
PNG files were being saved with JPEG quality encoder parameters, which PNG doesn't support. This caused malformed PNG files that failed ServiceNow's MIME type validation. The fix excludes PNG from receiving encoder parameters, allowing it to save with proper PNG headers that pass validation.

**// BEFORE (lines 74-80):**
EncoderParameters parameters = new EncoderParameters(1)
{
    Param =
    {
        [0] = new EncoderParameter(Encoder.Quality, surfaceOutputSettings.JPGQuality)
    }
};

**// AFTER (lines 74-85):**
// Don't use quality parameter for PNG (it doesn't support it and can cause corruption)
EncoderParameters parameters = null;
if (imageFormat.Guid != ImageFormat.Png.Guid)
{
    parameters = new EncoderParameters(1)
    {
        Param =
        {
            [0] = new EncoderParameter(Encoder.Quality, surfaceOutputSettings.JPGQuality)
        }
    };
}

danrhodes:
This is fixed in PR https://github.com/greenshot/greenshot/pull/744

danrhodes:
Compiled and tested working 👍

Interestingly, both pre- and post-fix files work on GitHub, so ServiceNow and other platforms may be performing additional checks/validation on the files.

joey4good:
@danrhodes 

Wow, that was quick! Thank you.
When will the next release be?

danrhodes:
I couldn't tell you, I have nothing to do with GS. I am just a lowly dev with an hour free to fix a bug in an excellent open-source project, saving the owner from having to spend time on it.

[...]

PR #744, opened 2025-12-16
Avoid quality parameter for PNG in file format handler

danrhodes:
Prevents setting the quality encoder parameter when saving PNG images, as PNG does not support this parameter and it can cause file corruption. The quality parameter is now only set for formats that support it, such as JPEG.

This fixed https://github.com/greenshot/greenshot/issues/743

### DIFF
--- a/src/Greenshot.Editor/FileFormatHandlers/DefaultFileFormatHandler.cs
+++ b/src/Greenshot.Editor/FileFormatHandlers/DefaultFileFormatHandler.cs
@@ -71,13 +71,18 @@ namespace Greenshot.Editor.FileFormatHandlers
             {
                 return false;
             }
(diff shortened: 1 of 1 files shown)


danrhodes:
👍

Lakritzator:
Great catch, it would have taken me hours to detect that little glitch ❤️🥰

danrhodes:
Glad to be of service. I have a good few days left off in December for leave so I have a few spare hours so happy to help where I can. I've used GS for years it is epic, views not of just myself but the whole team I work with also.

Thank you for such an amazing tool! ❤️

P.S - please see the other pull for the feature request I submitted we're screaming out for it to save literally thousands of clicks 🤣👍❤️👌

PR #752, opened 2025-12-20
[Fix] Click on balloon tip and open Explorer

### LINKED ISSUE #659: Destination not opening on mouse click
Hi

Sorry i canty seem to create a post on the user side.
I installed new Greenshot version, but now new screenshot Destination not opening on mouse click

Christian-Schulz:
I am sure that these are left-overs from development.
`notifyIcon.Tag` is not used and is never set.
The `SurfaceMessageEventArgs `are already available as parameters.

That should fix #659.


robredrob:
Hi ... I'm confused as what's happening here ... is there a fix? Is it to be implemented?

PR #754, opened 2025-12-21
[Feature] Replace WM_COPYDATA with Named Pipes for IPC

Christian-Schulz:
This migrates our Inter-Process Communication (IPC) layer to a more modern stack.

The old method uses WinAPI and sends data via WM_COPYDATA.  
The data was serialized using BinaryFormatter. 

The new method uses Named Pipes and I replaced BinaryFormatter with MessagePack for IPC-Data.

This is the next step towards newer dotnet versions. 🚀 

### How it works

When Greenshot is running, every process listen on its own pipe. The name of the pipe is composed of a prefix and the process id. This ensures that the name is always known (e.g., `Greenshot_Pipe_ID1234`).

Communication is fairly simple. When a Greenshot process wants to send data, it searches for all Greenshot processes, generates pipe names for the instances, and sends the data to all instances.  

The data transfer object `IpcMessage` has also been kept very simple.

You can find more detailed information about the IPC [here](https://github.com/Christian-Schulz/greenshot/blob/feature/new_ipc/src/Greenshot/Ipc/readme.md).

### Changes

I renamed `CopyDataTransport` to a clearer name `AppCommands`.
All other classes in `CopyData.cs` are no longer needed.

I created separate classes for sending, receiving, serializing, and deserializing.

### Braking changes

The old and new versions of the communication are not compatible. For example, the new version cannot terminate an running old version with the /exit command.

### Max received bytes

There is a data limit of 1 MB. This protects the `IpcListener` to consume to much memory if _"someone"_ sends large data to the pipe.
For this case I added a Testfile `\Greenshot\Ipc\Test\Test_Ipc_BreakTransferIfMaxMessageSizeReached.ps1`


Lakritzator:
I'm sure I wrote code similar to this somewhere, but can't find it, which annoys the hell out of me.

Anyway, I do find it important to discuss the why. It worked good so far, and we actually don't do much with the functionality so far. I was able to come up with the following reasoning:

1. Binary formatter is bad, a security issue waiting to happen. (Thus obsolete)
2. Windows messages, using WM_COPYDATA, is very limited and not usable with other platforms. 

I do not see a reason to have more then one Greenshot running though, it seems your solution does.

While typing I think I remember where I wrote similar code... It was for work, and this actually made me think of some issues. Windows messages are nicely limited to the user desktop, but named pipes can be system wide. We need to make this work for terminal server / multi user environments. I'll check if I have the code somewhere...

Christian-Schulz:
I encountered this while attempting to migrate Greenshot to .NET 10. The primary reason was the deprecation of the BinaryFormatter.

After spending a few hours auditing where and why we were using it, I concluded that it could be entirely replaced with much simpler logic. Furthermore, I believe it is beneficial to minimize WINAPI calls in favor of managed .NET implementations wherever possible.

Yes, under normal circumstances, only one instance of Greenshot runs at a time because the application explicitly prevents multiple instances. However, if you use 'Open with' Greenshot in File Explorer, a second process starts. Then it uses WM_COPYDATA to pass the image path from the new process to the original instance.

Unfortunately, I didn't think about the terminal server and multi-user scenario.

In theory, the current solution should already work for normal users. I'm not so sure about users with admin rights, as they would be able to see other users' processes.

I would first create a test environment for myself and try it out a little.
NamedPipes supports the configuration of AccessControl. I will take a closer look at that as well.

Lakritzator:
You  might "just" take the users session ID to name the pipe, there will only be conflicts if the Greenshot was started in the same session.

I use this library from me in another application of mine: 
https://github.com/dapplo/Dapplo.Windows/blob/master/src/Dapplo.Windows.Advapi32/Advapi32Api.cs#L23

Christian-Schulz:
I marked this as draft because:
- Wait until the serializer has been definitively decided in #638 
- I haven't had time yet for terminal server and multi-user scenario

Christian-Schulz:
Force pushed and rebase on current main.
Is still WIP.
I'm using now System.Text.Json for serzalisation.

Lakritzator:
I've been thinking about this for a while, and I have the following **idea**:

As Greenshot is relatively heavy to start for just a file click, we create a (.NET 10 with AOT) "stub" to process the opening of files or the commandline in general, this forwards the information via a named pipe towards Greenshot. Only this stub is registered to handle the file clicks / open with. This partly makes a generic handling possible, for if we move Greenshot to the Windows App store.

The Stub (Gatekeeper)
The stub is purely a dispatcher, it guarantees that only one greenshot is started (and starts it if needed) and forwards the command (commandline, whatever). 

Logic:

- Acquire local\Stub_Mutex (using WaitOne()) with a timeout, this allows other instances to handle the startup.
- Inspect: Use Mutex.TryOpenExisting("local\Greenshot_Mutex").
- Execute: If Greenshot App is missing, Process.Start("greenshot.exe").
- Handoff: Use NamedPipeClientStream with a polling loop (max 2 seconds) to send the file path to Greenshot.
- Clean up: Dispose Mutex and exit.

Greenshot (Receiver)
Greenshot does not care who sent the file (or other commandline options); it only cares that a path arrived.
Logic:

- Mutex Acquisition: During startup immediately acquire local\Greenshot_Mutex.
- Pipe listener: As soon as processing is possible, create the pipe listener
- ActivationManager: A singleton service that handles the received path string (pipe listener forwards to this)
- Clean up: Dispose pipe listener, Mutex and exit.

The communication should be a "formal" format, secure on the side of Greenshot (not allowing some dangerous action to be taken), so other applications could also benefit, having a "formal" handoff makes things way more flexible. This concept would make a step towards a browser extension, for capturing web-pages, using native messaging possible: https://learn.microsoft.com/en-us/microsoft-edge/extensions/developer-guide/native-messaging either by creating another small executable or using the same stub.

Issue #222, opened 2020-08-05
Capture Window not working correctly on Primary Monitor only


Simpuhl:
This GitHub repository is for developers, if you want to report a bug for Greenshot as a user please do so in our JIRA issue system here: https://greenshot.atlassian.net/projects/BUGS/issues/filter=allopenissues

**Describe the bug**
Weird issue, when I use the capture window it won't take the screenshots correctly - it just tries to capture full screen (only on screen 1. If I drag the window to my second screen it will let me choose specific boxes, etc. For example taking a screenshot within any web browser - edge, chrome, ie. If I try to take the capture window screenshot on screen 1 it will do fully screen and not just do the bottom section. If I drag the same browser to screen 2, it will allow me to capture just the bottom of the window.

If I reboot my machine, it will work for a while and then eventually stop again. Taking screenshots on screen 2 always works as expected. No idea what causes it but its been happening for months and its bugging the hell outa me lol.

**To Reproduce**
Example steps to reproduce the behavior:
1. Capture Window
2. Try to take screenshot in screen 1

**Expected behavior**
It should allow you to select window just like it does on other screen.

**Screenshots**

**Versions (please complete the following information):**
 - Greenshot version: 1.2.10 Build 6
 - Windows version: Windows 10 2004

**Additional context**
You guys are amazing

Simpuhl:
Now its not working at all, no matter which screen, if I reboot, or restart the app... it just tries to capture the full screen

toddatctera:
Hi @Simpuhl , 

You might want to try and isolate the issue from any other running applications. 
Do you have any kind of programs running that render over the whole screen like [Flux](https://justgetflux.com/) or a screen sharing/capture program like [OBS](https://obsproject.com/) or video conferencing apps?

In my case, I *think* I reproduced the issue you described while I was sharing my screen in Zoom and Microsoft Teams. 
Whichever screen I was currently sharing would not allow dynamic window/area selection. Just your basic region select or full screen screenshot.

If I activate the 'Capture Region' mode (Print Screen key for me), I can hit `Space` to toggle between Region selection (click and drag the mouse cursor) and this nifty Dynamic Window/Area selection (not sure exactly what this is called) which lets me screenshot the window or area under my mouse cursor. Like this:

![image](https://user-images.githubusercontent.com/47750691/113449510-04519900-93cc-11eb-8933-0e624745100b.png)

I hope that helps.

Lakritzator:
[...]

Issue #765, opened 2025-12-29
Window capture not working correct on Google Chrome


chuim:
**Describe the bug**
When I activate the capture window option, windows from Google Chrome are being ignored. A Windows File Explorer window is handled correctly, though.

**To Reproduce**
Example steps to reproduce the behavior:
1. R-click the Greeshot tray icon and select "Capture window".
2. Wait for the crosshair cursor to show up and hover over a Google Chrome window.
3. To compare, hover over a Windows File Explorer window.

**Expected behavior**
When hovering the crosshair over the Google Chrome window, the usual green area highlight should show up to cover it (or its parts). Clicking on it should trigger a capture.

**Screenshots**
Screenshot of a Chrome Canary window being ignored by the crosshair while a File Explorer in the background is still highlighted:

<img width="2118" height="1375" alt="Image" src="https://github.com/user-attachments/assets/b2a0e87c-842e-4e4b-a551-0da630ef9136" />

Note that the same happens with regular/stable Chrome windows. Also tested that the Discord window is correctly handled.

**Versions (please complete the following information):**
 - Greenshot version: 1.3.304
 - Windows version: Windows 11 Pro, 25H2, 26200.7462

**Additional context**

This used to work in the past. I'm not sure how long ago the issue started.

code1ine:
I was about to create a new issue, but I see that it’s already been well-explained here. I really hope to see this improvement implemented.

danrhodes:
PR for a similar issue which this will fix has been posted for review.

Lakritzator:
This should be fixed with 1.4, thanks to @danrhodes, if you like you can already download a pre-release to test this.
Closing this issue, you are probably automatically informed if the new version is released.
You are welcome to report back if the issue is not solved for you.

Issue #757, opened 2025-12-22
wrong file name


AaronSantos2021:
In previous versions the screenshot image with ${capturetime:d"yyyy-MM-dd"} ${title} was saved as - for example - 2025-12-22 Facebook.jpg, when I have taken a screenshot of a Facebook page.

${title} no longer works in any browser window tab, neither in Edge nor in Chrome. I would understand if the browser name were used, but "Greenshot" is used as the ${title}.

So, when I take a screenshot of a browser tab, "Greenshot" is always used as the ${title}, i.e. 2025-12-22 Greenshot.
Regardless of the tab's name in the browser, the file name is always "{Date}-Greenshot".
If I don't specify the time {HH_mm_ss}, the first screenshot is always overwritten by the last screenshot, because of the same title.

This only happens with screenshots from browser tabs. 
Screenshots from application windows (Office, Calc, VLC, etc.) work normally.

Lakritzator:
Can you give me a few more details:

- Does the browser window actually have a title? You can see this in the 
- Is the browser window running as the same user as greenshot? Check in the task manager (under details) what the different processes have as an owner. (It's a bad idea to run your browser as Admin / elevated) Greenshot can't get the title of a window if it doesn't belong to the same user under which Greenshot is running.

danrhodes:
I'll do some testing tomorrow also as I've experienced this issue frequently. I do wonder wether it could be the location of the mouse cursor when the capture is initiated and or the current active  window etc. The title isn't something I use often in my workflow but I have noticed it doesn't always capture the title correctly.

danrhodes:
Hmmm its not working for Edge it just returns the default Greenshot yet for such as notepad++ its fine

<img width="1232" height="718" alt="Image" src="https://github.com/user-attachments/assets/3f469f51-82e9-453a-b818-ff72d9c096d4" />

<img width="812" height="394" alt="Image" src="https://github.com/user-attachments/assets/4ae46566-44bb-4322-a464-d499720a8fd1" />

danrhodes:
Ok I’ve done some research on this and found that it’s not actually a Greenshot bug—it’s Edge and Chrome’s AppContainer sandboxing blocking the window title capture by design.

Chromium browsers run their content processes at low/untrusted integrity levels for security isolation.

Windows UIPI (User Interface Privilege Isolation) prevents normal applications like Greenshot from accessing window properties of these sandboxed processes.

[...]

Issue #729, opened 2025-11-28
Filename gets populated with an incorrect name (not the application being 'captured')


tveyben:
**The bug**
When capturing a screenshot (either using PRINTSCREEN to select an area or using ALT+PRINTSCREEN to select an applications window) the filename most often does not match the area that is captured, meaning that it often uses a name of a different application.

Most often Outlook or OneDrive appears in the filename. OneDrive even appears after Quit'ing OneDrive (though background processes are still present...)

The following attemps to identify a pattern have been tried
* Making sure that the captured area is 100% covering only one application
* Capture a full window thus only target one application
* Minimize all other applications thus no other application is drawn behind the one being screenshotted to potentially confuse Greenshot

**To Reproduce**
Example steps to reproduce the behavior:

1. Capture a screenshot of a Chrome Window

**Expected behavior**
Filename contains Chrome

**Screenshots**
<img width="1944" height="569" alt="Image" src="https://github.com/user-attachments/assets/d238602e-ee78-4d4a-b213-1b509009fb88" />

**Versions (please complete the following information):**
 - Greenshot version 1.3.304 (latest as of issue reporting time)
 - Windows version 11 25H2 (latest as of issue reporting time)

**Additional context**
n/a

danrhodes:
PR for a similar issue which this will fix has been posted for review.

danrhodes:
This is now fixed the PR was merged.

Lakritzator:
That is, if you would upgrade to the latest 1.4 which is a preview and isn’t released yet…

PR #767, opened 2025-12-31
Fix-issue-757 - Chromium based browser titles

danrhodes:
This is a PR to fix Chromium-based browsers (Edge, Chrome, Brave) not being detected for window capture and showing incorrect titles. Two issues were addressed: first, these browsers use the WS_EX_NOREDIRECTIONBITMAP extended style for hardware-accelerated rendering which was causing them to be filtered out of the window list - added IsChromiumWindow check to allow Chrome_WidgetWin_1 windows through. Second, sandboxed browsers block GetWindowText across process boundaries - added WindowTitleHelper with a fallback to Process.MainWindowTitle which only triggers when the fast native call returns empty. Explicit exception handling for process race conditions included.

Also addresses issues
https://github.com/greenshot/greenshot/issues/765
https://github.com/greenshot/greenshot/issues/757
https://github.com/greenshot/greenshot/issues/729
https://github.com/greenshot/greenshot/issues/517
https://github.com/greenshot/greenshot/issues/681

Possibly also fixes issues
https://github.com/greenshot/greenshot/issues/222
https://github.com/greenshot/greenshot/issues/658


src/Greenshot.Base/Greenshot.Base.csproj
```diff
@@ -21,6 +21,8 @@
     <PackageReference Include="Svg" Version="3.4.3" />
     <Reference Include="Accessibility" />
     <Reference Include="CustomMarshalers" />
+    <Reference Include="UIAutomationClient" />
```
Lakritzator: Are the UIAutomation references used anywhere?

Lakritzator:
Nice work, looks good at a first glance, need a night sleep though before I approve it.

Lakritzator:
I did find some bug, that we do not correctly handle the cloacked information. That "might" be another way to solve this generically, need to find my fix first, it's on some branch on one of my systems...

src/Greenshot.Base/Greenshot.Base.csproj
danrhodes: Good catch, no, this is left in from when I was trying this method, I will remove.

danrhodes:
UIAutomation references removed 👍

Lakritzator:
I'm not very fond of the IsChromiumWindow, as it sounds very maintenance heavy, with that I mean that this solution works for Chromium and not for other windows.

What might be interesting to know, if my solution works for your issue too:

```
        private static bool IsVisible(WindowDetails window, NativeRect screenBounds)
        {
            // Ignore invisible
            if (!window.Visible)
            {
                return false;
            }

            // Ignore minimized
            if (window.Iconic)
            {
                return false;
            }

            if (IgnoreClasses.Contains(window.ClassName))
            {
                return false;
            }

            if (window.IsCloaked)
            {
                return false;
            }

            // On windows which are visible on the screen
            var windowRect = window.WindowRectangle.Intersect(screenBounds);
            if (windowRect.IsEmpty)
            {
                return false;
            }

            if (window.Visible)
            {
                return true;
            }
            // Skip everything which is not rendered "normally", trying to fix BUG-2017
            var exWindowStyle = window.ExtendedWindowStyle;
            if (!window.IsWin10App && (exWindowStyle & ExtendedWindowStyleFlags.WS_EX_NOREDIRECTIONBITMAP) != 0)
            {
                return false;
            }

            return true;
        }
```

Btw. if you change the "GetVisibleWindows()" to contain the following line as the last one inside of the for, you can see what windows are skipped and the details on them:
Log.Debug("Skipped:\r\n" + window.ToString());

danrhodes:
I suspect this will have issues with sandboxed browsers because the `WS_EX_NOREDIRECTIONBITMAP` check is filtering out Chrome/Edge renderer processes—they typically set that flag on their tab content windows. Also not sure if the parent-child window hierarchy that Chromium uses will be handled correctly here, since tab content shows up as child windows rather than top-level.

No I Agree that hardcoding browser-specific class names is brittle and doesn’t scale—every browser update or new entry means maintaining a whitelist. Better to identify sandboxed/child windows based on actual rendering characteristics rather than string matching. The `WS_EX_NOREDIRECTIONBITMAP` flag itself might be a useful indicator if we flip the logic—windows *with* that flag in certain contexts are likely content being actively rendered.

I’ll test it against some live browser instances to see what actually gets through and whether there’s a pattern we can leverage without browser-specific detection.​​​​​​​​​​​​​​​​

danrhodes:
I'll give it a go though to be sure 👍

danrhodes:
This could work with no hard coded classes?

Also do we want to ignore minimized if we want to be able to capture windows from the context menu?


```

private static bool IsVisible(WindowDetails window, NativeRect screenBounds)
{
    // Ignore invisible
    if (!window.Visible)
    {
        return false;
    }

    // Ignore minimized
    if (window.Iconic)
    {
        return false;
    }

    if (IgnoreClasses.Contains(window.ClassName))
    {
        return false;
    }

    if (window.IsCloaked)
    {
        return false;
    }

    // On windows which are visible on the screen
    var windowRect = window.WindowRectangle.Intersect(screenBounds);
    if (windowRect.IsEmpty)
    {
        return false;
    }

    if (window.Visible)
    {
        // Check if it's a child window (potential sandboxed content)
        var parentHandle = User32.GetParent(window.Handle);
        var isChildWindow = parentHandle != IntPtr.Zero;
        
        var exWindowStyle = window.ExtendedWindowStyle;
        
        // Child windows with WS_EX_NOREDIRECTIONBITMAP are likely sandboxed browser content
        if (isChildWindow && (exWindowStyle & ExtendedWindowStyleFlags.WS_EX_NOREDIRECTIONBITMAP) != 0)
        {
            return true;
        }
        
        // Original check for non-child windows
        if (!window.IsWin10App && (exWindowStyle & ExtendedWindowStyleFlags.WS_EX_NOREDIRECTIONBITMAP) != 0)
        {
            return false;
        }

        return true;
    }

    return true;
}

```

Lakritzator:
We first need to see what we want to achieve, there are a few use-cases:

1. Make a screenshot of the active window via a hotkey, this should be fairly easy, as Win32 will tell us. In this case we should always be able to get the title (eventually via the process).
2. Make a screenshot from the screen, in this case we could already take the title of the active window IF the active window is on the screen(s) we captured or we have to look for one...
3. Interactive Window selection, this is either primary selected in the settings, or the user switches the mode with the spacebar. Here we make a screenshot from the screen, to present it in the region selection, get a list of visible windows on that screen, so the user gets the "you are over this window" overlay. To make this possible we use the list of top level windows visible on the screen(s), these are the ones which are primary our interest. This uses the IsTopLevel. If something isn't seen as a top level window, it's children are excluded. IF it IS a top-level window, the children are retrieves, so one can select an area of a window -> this won't function if the application uses it's own composition.
4. Capture Window from list, this is from the context-menu. Here we also would like to have the iconic windows, we can either move them back to visible, or use capture methods to get their content (working on using Windows.Graphics.Capture to solve the black edges & HDR issues)

About the different types of information we have:

1. Since Windows 10, many windows (especially UWP apps) can be "Cloaked." A cloaked window is technically "visible" to the Win32 API but is hidden from the user (e.g., a window on a different Virtual Desktop or a suspended app).
I think we got that easily covered, cloaked windows should be ignored.

2. WS_EX_NOREDIRECTIONBITMAP
This style is often used by modern applications (like Chrome !! or supposedly Discord) that use direct composition.
For me this might mean that we should handle this differently, maybe we can even ignore it in the selection of the windows, but need it to decide HOW we capture it.

It says:
Windows with this style don't have their contents managed by the standard GDI redirection bitmap. If you use PrintWindow or BitBlt on these, you might get a black box. For these windows, you must use the Windows.Graphics.Capture API (introduced in Win 10, version 1803) or the Desktop Duplication API for a reliable screenshot.

As soon as I use the Windows.Graphics.Capture API this might solve some issues... and maybe checking this is really unneeded.

3. Child or not, if it's a child it cannot be a top level window.

4. Window-Text / Title
If the title is empty, it's rarely a window the user interacts with.
Also if it has a title itself, it's most likely a top level window.

7. WS_EX_TOOLWINDOW
These are mostly overlays, there might be however a need for them...

Some hints about the code:

- You won't need to check is visible again, as it was already a fail fast.
- Actually I noticed that the window.Visible flag already checks the cloaked information. (I forgot I added that)
- FYI: It's interesting how IsWindowVisible and the WS_VISIBLE are related:
image

Maybe we should look at the methods used to get the information for the different use-cases, and clean them up?

danrhodes:
Cheers for the breakdown—makes way more sense now. Dyslexia makes walls of text a bit rough sometimes so appreciate you laying it out like that.

You’re spot on that we’re mixing visibility filtering with capture method logic. Splitting this into separate methods for each use-case (interactive selection, hotkey capture, context menu, etc.) sounds like the right move—keeps everything cleaner and more explicit.

Do you want me to make a start on the refactor or are you planning to tackle it yourself? Happy to have a go if you can lay out your rough idea?

Lakritzator:
I more or less jump back and forth between issues / PRs and stuff I find important, as I only get snippets of time due to my challenge handling a full-time job and a kid with special needs... so if I do not respond or are a bit "short" in answering, it's due to that challenge I have.

You are more then welcome to look at it, but only if it satisfies an itch, I mean if this is something you like to do, then go for it.

I currently need to think about how I continue with the feature/WGC branch (Windows.Graphics.Capture), which is the tech I need to introduce to solve issues we have with (and be able to throw away) legacy capture methods. I started quite rigorously, which might not be the best choice if there are issues down the path.

danrhodes:
Happy to help where I can, and yes it is something I enjoy.

I also work full time and have two kids so I can appreciate your comments.

Parenting comes first, always!

Keep strong 💪

danrhodes:
Ive done some additional work on this @Lakritzator and have listed below what I have tested with the latest update.

## Testing Checklist

### Chromium-based Browsers (Primary Fix)
- [x] Chrome - Active window capture via hotkey
image

- [x] Chrome - Window selection from context menu list
image

- [x] Chrome - Interactive window selection (spacebar mode)
image

- [x] Chrome - Window title displays correctly in filename/editor
Showing correct in all capture methods

- [x] Edge - Same tests as Chrome
Confirmed Edge also working the same as Chrome

### Other Direct Composition Apps (WS_EX_NOREDIRECTIONBITMAP)
- [x] Discord - Appears in window list and captures correctly
- [x] VS Code - Appears in window list and captures correctly
- [x] V Studio - Appears in window list and captures correctly
image

### Windows UWP Apps (IsWin10App)
- [x] Calculator - Captures correctly
image

- [x] Settings app - Captures correctly
image

- [x] Microsoft Store - Captures correctly
image

### Standard Applications (Regression Testing)
- [x] Notepad - Still works as before
image

- [x] File Explorer - Still works as before
image

- [x] Other Win32 apps - No regressions

### Window Filtering
- [x] Minimized windows appear in context menu list
- [x] Cloaked windows (other virtual desktops) do NOT appear

### Capture Modes
- [x] Active window hotkey
- [x] Full screen capture
- [x] Region selection
- [x] Interactive window selection
- [x] Context menu window list

danrhodes:
My bad, just noticed I didn't re capture the settings capture to show the title but can confirm it captured correctly.

src/Greenshot.Base/Core/WindowsEnumerator.cs
```diff
@@ -50,6 +51,11 @@ public WindowsEnumerator GetWindows(IntPtr hWndParent, string classname)
             if (hasParent)
             {
                 parentText = User32Api.GetText(hWndParent);
```
Lakritzator: Does applying the "title" (text) of the parent to the children actually make sense here, as we test "window.Text.Length" on other places. Maybe we should get the title parent if we are really interested in the title elsewhere?
(I know that it was there before, that is not something you added 😂)

Lakritzator:
Very impressive, although I didn't test it, already just looking at the tests and the results.
I only found that one comment I created, let's discuss that.

src/Greenshot.Base/Core/WindowsEnumerator.cs
danrhodes: You're right - this makes window.Text.Length checks elsewhere misleading since they'd test the parent's title, not the child's own text. Better to fetch the parent title on-demand when needed via ParentHandle.

danrhodes:
Long delay in getting back to you, apologies. I've removed code that was copying the parent window's title to all its child windows. This was misleading as other parts of the code check if a window has a title to decide whether to include it - but they were checking the parent's title instead of the child's own title. Now each window keeps its own title, and the parent's title can still be looked up when needed.

Lakritzator:
That's very cool and really great work. That while I was traveling, giving my kid a birthday party, taking care the same kid being sick and doing silly stuff like investigating something that bothered me for years: https://github.com/greenshot/greenshot/pull/863

We owe the community many thanks...

danrhodes:
Thanks! Hope your kid's feeling better and that the birthday party went well despite the timing.

PR #769, opened 2026-01-02
Fix the spacing of the checkboxes on the Expert tab in preferences.

danrhodes:
This has bugged me for a while with the inconsistent spacing hence the PR to amend the spacing to make it look clearner.

Before
000034Visual Studio Code

After
000046 - Greenshot - DefaultFileFormatHandler cs - Microsoft Visual Studio


danrhodes:
Closed as extra code looks to have snook in here which shouldn't be there!?

PR #778, opened 2026-01-05
Adjust layout and sizing of expert settings controls

danrhodes:
Updated the location and size properties of several checkboxes, labels, and textboxes in the expert settings group to improve alignment and consistency in the SettingsForm.

Before
before

After
image


danrhodes:
A minor change, but one that has bugged me not being aligned for a while....

PR #786, opened 2026-01-07
Update silentArgs for Greenshot installation

hjenkel:
I added /ALLUSERS to the Chocolatey install to support automatic rollouts with system user.

### DIFF
--- a/chocolatey/tools/chocolateyInstall.ps1.template
+++ b/chocolatey/tools/chocolateyInstall.ps1.template
@@ -9,8 +9,9 @@ $packageArgs = @{
   softwareName  = 'Greenshot*'
   checksum      = 'CHECKSUM_PLACEHOLDER'
   checksumType  = 'CHECKSUM_TYPE'
-  silentArgs   = '/VERYSILENT /SUPPRESSMSGBOXES /NORESTART /SP-'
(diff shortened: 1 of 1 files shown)


Lakritzator:
I'm not sure if we want to force all chocolatey users to simply have the installation for all users.
I'm no expert, but I expect that this makes it impossible for users with lower rights to update their application.

hjenkel:
Hi @Lakritzator, thanks for your reply. As far as I understand, Chocolatey is designed to install software system-wide by default. It even shows a warning when you try to install packages from a non-elevated shell. 

We ran into an issue where Greenshot cannot be installed at all without administrative privileges, because there is no write access to the Chocolatey directory. As a result, the Chocolatey package is unusable in a non-admin context. This assessment is based solely on our environment.

In theory, it might be possible to add a package parameter to support a per-user installation. However, I would strongly assume that system-wide installations are the preferred and intended approach for Chocolatey packages.

lunndal:
@hjenkel is right. The [Chocolatey FAQ](https://docs.chocolatey.org/en-us/faqs/) specifically states that Chocolatey is a "_machine level package manager_".

Lakritzator:
That helps, thanks for clarifying this and I agree.

BTW. I'm looking at integrating the Windows Restart Manager, which would support a quick update without users losing information

Lakritzator (approved):

PR #798, opened 2026-01-10
Improve Inno Setup installer script

bovirus:
@Lakritzator 

Improve installer script

- Change to use RTF GPL license (not plain text) - Better use of Inno Setup graphic UI
- Change default installer folder (only for new installation -- not for update) - Now use ProgramData - No admin right required
- Add option for Inno Setup Modern Wizard UI
- Add string for Win11 default snippet tool and Italian language

If you can please update Inno Setup to a newer version (current seems 6.1.0 - available 6.7.0).
In this way you can setup better WiuzardModern GUI options

### DIFF
--- a/src/Greenshot.Editor/Helpers/BinaryFormatterHelper.cs
+++ b/src/Greenshot.Editor/Helpers/BinaryFormatterHelper.cs
@@ -25,6 +25,7 @@ using System.Runtime.Serialization;
 using System.ServiceModel.Security;
 using Greenshot.Base.Interfaces.Drawing;
 using Greenshot.Editor.Drawing;
+using Greenshot.Editor.Drawing.Emoji;
 using Greenshot.Editor.Drawing.Fields;
 using Greenshot.Editor.Drawing.Filters;
 using log4net;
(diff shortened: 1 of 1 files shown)


Lakritzator:
Hey, I already created a PR for this some days ago, based upon your suggestions.
But using commonappdata is a BAD idea, we still want to provide the user the possibility to install in program files.
This is actually the most secure way of installing, commonappdata is not.

bovirus:
@Lakritzator

I posted the PR because I didn't see the changes in latest stable and beta version released some days ago.

I added also WizardStyle=modern option to make the UI more modern.

Sorry but I didn't udnerstand why use of ProgramData installation folder is a bad idea and most secure use "Program Fidles".

The common general approach in last time about program installation is to use "ProgramData" installation folder to avoid issues withI user rights that can be make the "Program Files" folder not available (partially or fully) for all users or create issue about partial access to "Program Files" folder that can create problem with the application.

Any users has full access to ProgramData then no issue with user rights.

If the user has previously installed it in Program Files folder new version (if you setup as defaulkt instalaltion for new installation as ProgramData folder) continue to use Program Files folder.

bovirus:
@Lakritzator 

I restored the default installation folder {autopf}.

Please if you can approve the PR.

bovirus:
@Lakritzator 

I see that beta version has now WizardStyle enable. Thanks you.
Remain the strings untranslated.
Could you please fix the issue in beta version regarding string untranslated in component list regarding Win11 snippet default tool? 

Could you please approve this PR to enable changes also on stable version?

Thanks.

bovirus:
@Lakritzator 

New beta (1.4.56) with same issue regarding untranslated string for Win11 default snippet tool.

What do you think to have a separate file only for custom messages?

In this way the changes about custom messages didin't change master inno setup scxript.

Could you please apply changes with this PR to Custom message for Italian language?

Thanks.

PR #813, opened 2026-01-14
Fix GCHandle memory leak in DibFileFormatHandler

tsiakoulias:
## Summary
Fix inverted condition in `finally` block that prevents `GCHandle` from being freed.

## Change
`src/Greenshot.Editor/FileFormatHandlers/DibFileFormatHandler.cs` line 113:
- Before: `if (gcHandle == IntPtr.Zero)`
- After: `if (gcHandle != IntPtr.Zero)`

## Problem
The condition was inverted, causing:
- Handle never freed when valid (memory leak)
- Attempted free when null (no-op)

## Impact
Memory leak on every DIBV5 clipboard paste operation.

### DIFF
--- a/src/Greenshot.Editor/FileFormatHandlers/DibFileFormatHandler.cs
+++ b/src/Greenshot.Editor/FileFormatHandlers/DibFileFormatHandler.cs
@@ -110,7 +110,7 @@ namespace Greenshot.Editor.FileFormatHandlers
             }
             finally
             {
-                if (gcHandle == IntPtr.Zero)
+                if (gcHandle != IntPtr.Zero)
                 {
                     GCHandle.FromIntPtr(gcHandle).Free();
                 }


Lakritzator:
Oh wow, another impressive find 🥰

tsiakoulias:
Thanks for the warm welcome, glad I could help a tiny bit 🥳
Keep up the good work!

danrhodes:
Very impressive! 👌🚀

danrhodes:
@tsiakoulias do you have any numbers pre/post fix out of interest?

I've noticed lately a slow down in GS in terms of how long it takes to load the cross hairs and then how long it takes to put the image on the clipboard.

Obviously not related but just interested.

tsiakoulias:
@danrhodes Greetings!
No metrics captured for this fix. Regarding GS cross hairs on release versions, they appear almost instantly (might say sub 250ms) for me. Regarding paste lag from clipboard I haven't noticed any at all as well (by the time I paste, its already there).
Could be a multi-plugin and/or multi-monitor behavior?
I found some other potential memory wastes in other parts of the code, subject to a future PR when I can verify them, but they are probably unrelated to the effects you are describing.

PR #816, opened 2026-01-14
Use cryptographically secure random for OAuth nonces (corrected pr)

tsiakoulias:
**Problem:**
OAuth nonce generation used `System.Random`, which is not cryptographically secure and predictable.

**Suggested Implementation:**
- Replaced `System.Random` with `RandomNumberGenerator` (OS-level CSPRNG)
- Changed nonce format from numeric to Base64URL-encoded string
- Limited to 64-bit entropy per nonce (8 bytes are enough)
- Nonce output is URL-safe (replaced `+` with `-` and `/` with `_`)

**Impact:**
Prevents nonce prediction and replay attacks in authentication flows.

**Standards Compliance:**
- RFC 5849 (OAuth): Cryptographically secure nonces
- RFC 4648 (Base64URL): URL-safe encoding (without padding)

**Files Changed:**
- `src/Greenshot.Base/Core/OAuth/OAuthSession.cs`

### DIFF
--- a/src/Greenshot.Base/Controls/EnumComboBox.cs
+++ b/src/Greenshot.Base/Controls/EnumComboBox.cs
@@ -1,0 +1,109 @@
+﻿/*
+ * Greenshot - a free and open source screenshot tool
+ * Copyright (C) 2007-2026 Thomas Braun, Jens Klingen, Robin Krom
+ * 
+ * For more information see: https://getgreenshot.org/
+ * The Greenshot project is hosted on GitHub https://github.com/greenshot/greenshot
+ * 
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 1 of the License, or
+ * (at your option) any later version.
+ * 
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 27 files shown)


Lakritzator:
The idea is good, I mean the code comes from the time that mostly oauth 1 was used. Still I fear that this won't help, in this form. 

First while I actually expect that most upload plugins don't anyway, but also because the nonce is probably bigger this way, I think 12 characters instead of 8?
The base 64 encoding makes the string bigger, and I'm not sure about the size limitations. I mean I believe the nonce was supposedly 8 characters.

I do have repairing the plugins for Box, Imgur, DropBox, Flickr and Photobucket on my TODO list, I would believe these all use OAuth 2 by now...

tsiakoulias:
Good point on the length after conversion (and apparently Flickr might expect just 8 (numeric only?) at least from their example which is a bit unclear until tested).

Let's just propose the change from System.Random to RandomNumberGenerator and call it a day until OAuth2, if that finds you agreeing.

(Even thought they could accept alphanumeric (RFC allows and some services use it like that, instead of just numbers, length might also vary above 8) and we could reduce the bytes to 6 randoms, so we get a base64 of 8, we can easily skip since OAuth2 is abundant and there's also the plan to migrate.)

ClubSpeedBogdan:
overengineering nonsense

Lakritzator (approved):

Issue #801, opened 2026-01-12
Trying to configure the Confluence plugin


irLunatik:
**Describe the bug**

**To Reproduce**
Example steps to reproduce the behavior:
1. Settings, Plugins
2. Select Confluence Plugin
3. Click Configure
4. Resize

**Expected behavior**
Not sure

**Screenshots**
<img width="529" height="480" alt="Image" src="https://github.com/user-attachments/assets/496a3345-c05a-433f-b996-6340164d0ec8" />

Sony, an unexpected error occuœd
The good news is: you can help us getting rid of it by filing a bug report.
Please visit the URL below. create a new bug report and paste the contents from the text area into
the desctiption.
Please add a meaningful summary and enclose any information you considerto be helpful for
reproducing the issue.
Aso, we would highly appreciate if you checked whether a tracker item already exists forthis bug.
(You can use the search to find those quickly Thank you :)
https://qetqreenshot.orq,•tickets/

**Versions (please complete the following information):**
Software version: 1.3.300 - c1ad1d4a93 (64 bit)
.NET runtime version: 4.0.30319.42000+
Time: 2026-01-12 13:08:09 +01:00
OS: Windows 10 2009 x64 build 22631
GDI object count: 126
User object count: 91

Exception: System.Windows.Markup.XamlParseException
Message: Provide value on 'System.Windows.Baml2006.TypeConverterMarkupExtension' threw an exception.

Stack:
   at System.Windows.Markup.XamlReader.RewrapException(Exception e, IXamlLineInfo lineInfo, Uri baseUri)
   at System.Windows.Markup.WpfXamlLoader.Load(XamlReader xamlReader, IXamlObjectWriterFactory writerFactory, Boolean skipJournaledProperties, Object rootObject, XamlObjectWriterSettings settings, Uri baseUri)
   at System.Windows.Markup.WpfXamlLoader.LoadBaml(XamlReader xamlReader, Boolean skipJournaledProperties, Object rootObject, XamlAccessLevel accessLevel, Uri baseUri)
   at System.Windows.Markup.XamlReader.LoadBaml(Stream stream, ParserContext parserContext, Object parent, Boolean closeStream)
   at Greenshot.Plugin.Confluence.ConfluencePlugin.Configure()
   at System.Windows.Forms.Control.OnClick(EventArgs e)
   at System.Windows.Forms.Button.OnMouseUp(MouseEventArgs mevent)
[...]

PR #817, opened 2026-01-14
Fix Confluence plugin settings

Christian-Schulz:
I dont think this plugin is in use anymore.
But I had the fix locally for a while.

This references #801


Lakritzator:
That's cool anyway, I'll take it 😉

PR #820, opened 2026-01-15
Fix #804: Use Greenshot's UI language for date formatting in print footer

### LINKED ISSUE #804: Print Footer Date Format Ignores Configured UI Language
When printing an image captured with Greenshot, the date displayed in the footer of the printed document consistently uses an English format (e.g., MM/DD/YYYY or Month DD, YYYY). This occurs even when the Greenshot UI language is explicitly set to German (Deutsch). The expected behavior is that the date format in the footer should align with the selected UI language, using its corresponding locale-specific date representation (e.g., DD.MM.YYYY for German).

This issue leads to an inconsistent user experience where a localized application fails to apply the chosen localization to all output elements.

Steps to Reproduce:
    Launch Greenshot.
    Open Greenshot settings (right-click the Greenshot tray icon, select "Einstellungen...").
    Navigate to the "Sprache" (Language) tab.
    Ensure that "Deutsch" (German) is selected as the UI language and apply the settings.
    (Optional but recommended for consistency) Verify that the operating system's regional format is also configured to German (e.g., "Deutsch (Deutschland)").
    Capture any screenshot using Greenshot.
    In the Greenshot editor, select "Datei" -> "Drucken..." (or directly print from the context menu if enabled).
    In the print dialog, ensure that the "Datum" (Date) option is selected for inclusion in the footer.
    Print the image (e.g., to a PDF printer or a physical printer).
    Examine the footer of the resulting printed document or PDF.

Expected Behavior: The date in the print footer should be formatted according to the configured UI language in Greenshot. For German, this would typically be DD.MM.YYYY or a similar localized format.

Actual Behavior: The date in the print footer is displayed in an English format (e.g., MM/DD/YYYY), irrespective of Greenshot's UI language being set to German.

Environment:

    Greenshot Version: 1.3.312 64bit
    Greenshot UI Language: German

tsiakoulias:
Date/time in print footer uses system locale instead of Greenshot's configured UI language. 
https://github.com/greenshot/greenshot/issues/804

The proposed fix now uses CultureInfo.GetCultureInfo(Language.CurrentLanguage), so the date formatting respects the GS language and not the Windows system locale.

### DIFF
--- a/src/Greenshot/Helpers/PluginHelper.cs
+++ b/src/Greenshot/Helpers/PluginHelper.cs
@@ -43,7 +43,6 @@ namespace Greenshot.Helpers
         private static readonly ILog Log = LogManager.GetLogger(typeof(PluginHelper));
         private static readonly CoreConfiguration CoreConfig = IniConfig.GetIniSection<CoreConfiguration>();
 
-        private static readonly string PluginPath = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), Application.ProductName);
         private static readonly string ApplicationPath = Path.GetDirectoryName(Application.ExecutablePath);
         private static readonly string PafPath = Path.Combine(Application.StartupPath, @"App\Greenshot");
 
@@ -208,7 +207,6 @@ namespace Greenshot.Helpers
             }
             else
             {
-                pluginFiles.AddRange(FindPluginsOnPath(PluginPath));
                 pluginFiles.AddRange(FindPluginsOnPath(ApplicationPath));
             }
 


Lakritzator:
I currently can't accept this PR, as this functionality is used for generating the filenames as well as for the footer (I assume you are right in that). But this means that we change existing functionality, where some people might rely on.

Lakritzator:
I was actually thinking we should pass the culture from the places where we need another culture. But this works too for me.
I'm sorry that I didn't write that down and had to shorten my comment.
I often start commenting and then my family/kid "distracts" me a bit and I have to "comment" an issue, as I don't know when I have time again... the life of a dad with a special needs kid. 🤷‍♂️

Anyway, I take it like this, thanks.

tsiakoulias:
Greetings, no apologies required... haha. I didn't comment but you pulled already, took a deeper look and found the usage you mentioned, and proposed the addition so both cases can be satisfied. As always, if you feel this isn't up to par with what you have in mind you can reject any time. I have in mind to create some other PRs for unreported bugs as well some time, when I can have some quiet as well :D (these might require building and re-running more often than I'm used to). Take care!

Issue #819, opened 2026-01-15
Greenshot does not work anymore after Update, that failed


norbneut:
**Describe the bug**
A few days ago, I was prompted to perform a software update for Greenshot.

I started the update, but it crashed. Since then, Greenshot has been unavailable to me.

Our company has a software center that allows us (as regular users) to install and uninstall standard software.

There, I also have the option (sometimes) to initiate a repair for a piece of software. I did that for Greenshot as well, but it didn't help.

I am very desperate because I use Greenshot often. Without Greenshot, my work is much more complicated.

I therefore request quick support.

**To Reproduce**
Example steps to reproduce the behavior:
1. Capture '....'
2. Open Editor
3. Add a line
4. Resize

**Expected behavior**

**Screenshots**

**Versions (please complete the following information):**
 - Greenshot version
 - Windows version

**Additional context**
Software version: 1.3.304 - f46c1b7f31 (64 bit)
.NET runtime version: 4.0.30319.42000+
Time: 2026-01-15 08:44:56 +01:00
OS: Windows 10 2009 x64 build 22631
GDI object count: 56
User object count: 16

Exception: System.InvalidOperationException
Message: Sequence contains more than one element

Stack:
   at System.Linq.Enumerable.SingleOrDefault[TSource](IEnumerable`1 source)
   at Greenshot.Forms.MainForm..ctor(CopyDataTransport dataTransport)
   at Greenshot.Forms.MainForm.Start(String[] arguments)

Configuration dump:

---
**Stack Trace Hash:** [83c36af0b23b]

Lakritzator:
I think the previous Greenshot version wasn't correctly deleted, or maybe Greenshot is deployed without removal script.
This means Greenshot finds plugins in multiple directories.

If Greenshot is installed & running in Program files, which from my point of view is the right choice, delete the greenshot files here:

%LOCALAPPDATA%\Greenshot

For me that is C:\Users\robin\AppData\Local\Greenshot

Lakritzator:
I have made a change in future Greenshot versions which should prevent this issue. With the hints I gave, you should be able to fix the issue, I assume you do have write permissions in the location I gave you. Let me know if that helped.

norbneut:
Hello,
our internal support team has now come up with a solution.

I was prompted to update Greenshot again today, and since then it has been running as usual.

Thank you very much for your support. You can now close this ticket.

Best regards,
Norbert

PR #826, opened 2026-01-15
[Experiment] Extend Serviceprovider

Christian-Schulz:
I tested a bit, how it is pssible to force this error in #819 

```
Exception: System.InvalidOperationException
Message: Sequence contains more than one element

Stack:
at System.Linq.Enumerable.SingleOrDefault[TSource](IEnumerable`1 source)
at Greenshot.Forms.MainForm..ctor(CopyDataTransport dataTransport)
at Greenshot.Forms.MainForm.Start(String[] arguments)
```

It was possible when Greenshot found the plugin Win10Plugin.dll twice in different locations. While starting the `INotificationService` was added in ServiceProvider two times.

And the check `if (SimpleServiceProvider.Current.GetInstance<INotificationService>() == null)` in `MainForm`  calls `SingleOrDefault()`;

I think the reason for #819 is a defect installation. 

I extend the message in the error dialog and added the plugin list. I think this is a helpful information.
image

I know the Win10Plugin no longer exists, but the actual cause remains as long as the DLL is in the installation folder.

In `SimpleServiceProvider.GetInstance()` i replaced `SingleOrDefault()` with `FirstOrDefault()` and checked the number of instances by hand for a more informative error message.
This shows early if the `SimpleServiceProvider` is configured incorrectly. 

Some services are required, others are not. 
To make this clearer, I have added the parameter `isRequired` and throw an error if the service is missing.

@Lakritzator  Any thoughts on my changes?


src/Greenshot/Forms/MainForm.cs
```diff
@@ -57,6 +57,7 @@
 using Greenshot.Plugin.Win10;
 using Greenshot.Processors;
 using log4net;
+using Newtonsoft.Json;
```
Lakritzator: I'm a bit confused about this include? Might it be an unused using?

Lakritzator:
I was wondering why we even could have 2 plugins..

I do not directly against your solution, but look here:

```
        private static readonly string PluginPath = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), Application.ProductName);
        private static readonly string ApplicationPath = Path.GetDirectoryName(Application.ExecutablePath);
```

And in Load plugins:

```
                pluginFiles.AddRange(FindPluginsOnPath(PluginPath));
                pluginFiles.AddRange(FindPluginsOnPath(ApplicationPath));
```

To be honest, I really think we should remove the first one, and just stick with the application path, problem solved.
This also solves potential security issues.

Lakritzator:
This might also cause the other issues with "missing DLLs" or stuff like that!

Lakritzator:
I made the change, so removed the "PluginPath", as I really think that is important.
But I do really want to thank you for the assist, that saved me a lot of investigation time, thanks to you I knew exactly where to look. ❤️

src/Greenshot/Forms/MainForm.cs
```diff
@@ -614,8 +615,8 @@ private void HandleDataTransport(CopyDataTransport dataTransport)
                         break;
                     case CommandEnum.FirstLaunch:
                         LOG.Info("FirstLaunch: Created new configuration, showing balloon.");
-                        var notifyIconClassicMessageHandler = SimpleServiceProvider.Current.GetInstance<INotificationService>();
-                        notifyIconClassicMessageHandler.ShowInfoMessage(
+                        var notifyIconClassicMessageHandler = SimpleServiceProvider.Current.GetInstance<INotificationService>(isRequired: false);
```
Lakritzator: Why do we want to do this optionally? I would assume something is very broke if there is no INotificationService?

Christian-Schulz:
@Lakritzator Do you think it's worth the effort to extend SimpleServiceProvider?
Currently, it is possible to add multiple instances, even though other places assume that exactly one instance in the service list exists. That shouldn't be possible.

Lists of services are only needed because of plugins, destinations, and processors. I think this should be handled differently. There should be explicit types PluginList, DestinationList, and ProcessorList, and these should be registered exactly once in the ServiceProvider.

Lakritzator:
It was added as a cheap di solution, Microsoft.Extensions would offer a much better solutions.
I do not think that extending it brings us much, I would leave it the way it is, until we have better use cases.

My .NET 10 prototype/experiment, and it's really that, uses this. I gave you the permissions to look at it.

Christian-Schulz:
> I would leave it the way it is 

👍

> My .NET 10 prototype/experiment, and it's really that, uses this. I gave you the permissions to look at it.
 
Oh, I think I had to read a lot.

However, I think the plugin list in the error message is useful.
It also helps to find out which plugins may be actively disabled by many users.

Lakritzator:
So I already removed the extra "pluginFiles.AddRange", to reduce the error cases as it should no longer find the plugins in multiple locations, in another commit.

And I think removing the "SingleOrDefault()" does make a lot sense, but still having a log entry if something is weird.
I also agree on having the list of plugins.

But I do not see "isRequired" the way to go forward, I am okay with an isOptional with the default being false.
Can you change this PR,  so we can reduce more errors?

Christian-Schulz:
Because the default behavior has now changed.
I checked all calls to “GetInstance()”. 
I think `IOcrProvider` is the only Service wich is really optional. Can you double check that ?

Lakritzator:
The OCR Provider is no longer optional, as I assume Greenshot is only used on Windows 10/11, we cannot test it on other versions anyway.

Christian-Schulz:
Now are all Services are required. 

After some of the old `GetInstance()` calls are null checks, wich are actually no longer nessesary. In that case, an exception would be triggered before.

Lakritzator (approved):

Issue #829, opened 2026-01-16
export screenshots as .pdf format


trufik-ctrl:
<img width="645" height="554" alt="Image" src="https://github.com/user-attachments/assets/b388a5e8-baf2-48c8-83c3-c258fa691660" />

It will be nice if greenshot could export screenshots as pdf format

Lakritzator:
Tell me more, why would you need that, how should it look? I personally don’t see much benefit of having a single screenshot in a pdf, without having text or anything else. So O never came up with a good use case, thus it wasn’t build.

Try to explain me the requirements.

trufik-ctrl:
1 reason: you can bind screenshots las pages, is great for doing instructions, tutorials etc. And keep theese pages in one file 

2 reason: on pdf you can Insert editable comments for other people who you for example share binded pdf with screenshots as pages,

3 reason: you can lock pdf with password 

4 reason: in every pdf editor after inserts arrows, marks, text, other people can edit it after saving file.

Of course you can print mamy PNG jpg to one pdf document using virtual printer but when you using it with many screenshots many times a day is very uncomfortable 

I work as cad engineer and in my company using screenshots as pdfs is very easier to communicate with other company departments and production

Lakritzator:
Just that I understand your requirements:

- You make a screenshot
- While saving you select .pdf
- Greenshot creates a PDF file with the filename in the save dialog, containing the image.
That is it?

At the end of the day you have multiple PDF files with a single screenshot in it, and that helps you more then to save them as a .png or .jpg?

I'm just a bit unclear & unsure that it really saves time, why this doesn't bring multiple screenshots, let alone instructions... And how would you get additional screenshots to your tutorial?

trufik-ctrl:
[binded.pdf](https://github.com/user-attachments/files/24687204/binded.pdf)
[2026-01-17 at 11-25-47.pdf](https://github.com/user-attachments/files/24687205/2026-01-17.at.11-25-47.pdf)
[2026-01-17 at 11-24-46.pdf](https://github.com/user-attachments/files/24687206/2026-01-17.at.11-24-46.pdf)

here is example (i did screenshots using other program). edit with simple version acrobat reader.

just i would like greenshot to do screenshots exacly the same as png jpg etc only in pdf format, thats all. Then it will be perfect program for me

Christian-Schulz:
Please have a look at #831

PR #831, opened 2026-01-18
[Experiment] save as .pdf

Christian-Schulz:
This is a test for #829.
My fist attempt to save a screenshot as pdf looks quite nice.
I wrote the PDF on my own without a an external lib.  Since it only contains one image, it's easy to implement. 
Internally, I have to use a JPEG, which unfortunately reduces the quality.

The correct scaling and conversion to print dimensions is still incomplete.

image

### DIFF
--- a/src/Greenshot.Plugin.Pdf/Configuration/PdfExportSettings.cs
+++ b/src/Greenshot.Plugin.Pdf/Configuration/PdfExportSettings.cs
@@ -24,58 +24,52 @@ using Greenshot.Base.IniFile;
 namespace Greenshot.Plugin.Pdf.Configuration
 {
     /// <summary>
-    /// PDF export settings for page layout and image scaling
+    /// PDF export settings for document layout and image placement
     /// </summary>
     [IniSection("Pdf", Description = "Greenshot PDF Plugin configuration")]
     internal class PdfExportSettings : IniSection
     {
         /// <summary>
-        /// Page width in millimeters
+        /// Document width in millimeters
         /// </summary>
-        [IniProperty("PageWidth", Description = "Page width in millimeters", DefaultValue = "210.0")]
-        public double PageWidth { get; set; } = 210.0;
+        [IniProperty("DocumentWidth", Description = "Document width in millimeters", DefaultValue = "210.0")]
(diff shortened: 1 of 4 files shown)


Lakritzator:
I love doing experiments like that...

It's a great, simple implementation. I tested it, and it integrates very easily, I'm happy that I created the FileFormatHandler for these cases...

I would almost merge it like this, but I'm slightly hesitant, as these things open the door into many discussions on extending the features.

Writing PDFs via a library makes it a bit more challenging, I recently (for another project) had to dive into PDFs and found that there are paid and free solutions, and they all come with a limitation that they use some native dependencies (SkiaSharp in most cases).

What do you mean with the following:

- The correct scaling and conversion to print dimensions is still incomplete.

Christian-Schulz:
I would go one step further.

What if the output format could also be expanded dynamically?
Currently, the list of file extensions is still static. If this were moved to the ServiceProvider, it would be flexible. I don't know if it has side effects, but it might be easy to do.
After that you could create a PDF-Plugin and offer a settings dialog for PDF export.

With the scaling i mean:
In PDF you have to define the document in mm (or inch for the non metric folk). Currently i choose the width of DINA4. I think the user should decide whether the output document is always the same size or depends on the screenshot dimensions.

_(I have also used various PDF libraries in my other projects. Over time, there are always license changes. However, I have no experience with their use in the context of open source.)_

Christian-Schulz:
This is the next try,
I created a PDF export plugin with its own settings.

Settings_example

Christian-Schulz:
Any thoughts?
Should I go ahead and create a complete plugin with translation and integrate it into installations?

Lakritzator:
I had to take some "time off" of Greenshot to study, but I finished today.
The only thing I did work on, was PR 1000, (didn't want to link it just for a remark) where I tried to get the Restart Manager code integrated. Was more difficult than I thought... and I changed way more than I should...

I just finished but my eyes close by their selves, it's late...

Christian-Schulz:
Be careful not to overwork yourself. There are still plenty of days left in the year.

Christian-Schulz:
Note for me: #1376  should help to define the supported file extension `.pdf` for the “Save As” dialog in the plugin.

PR #68, opened 2017-08-01
Fix for shadows

Arvidas:
circles

This fix makes drawings shadows look more professional.

PR #835, opened 2026-01-19
Improved shadows by remaking #68 as this wasn't merge-able.

Lakritzator:
Hey @Arvidas, I recreated your pr #68 as I agree it's an improvement.
Unfortunately it was laying around so long, it was impossible to merge, that's on us...

FYI: I had to remove the part which probably costed you the most time, this was for the Rectangle, the part you copied from the speech bubble. It didn't work nicely with a large border. So I used the simple approach.

I do not want to take credit for this, so I hope you might be able to take this and re-apply your PR.
Or maybe you can do just a small change, so you are visible as that you contributed to Greenshot?

### DIFF
--- a/src/Greenshot.Editor/Drawing/DrawableContainer.cs
+++ b/src/Greenshot.Editor/Drawing/DrawableContainer.cs
@@ -406,7 +406,7 @@ namespace Greenshot.Editor.Drawing
 		protected static void DrawShadow(int lineThickness, Action<int, int, Pen, Brush> drawShadowStepAction)
         {
             double alpha = 100;
-            double stepsCount = 5; // increase shadow width according to thickness 
+            double stepsCount = 5;
             double alphaStep = alpha / stepsCount;
             int currentStep = 0;
             using (var brush = new SolidBrush(Color.Black))


Lakritzator:
After looking at it again, I am not so sure what I like more.

Before:
Test capture - without improve shadow

After
Test capture - improve shadow

Lakritzator:
@jklingen I'd like to have your opinion, I am not sure what looks better...

Cipriux:
Whatever shadow type it is, it should not appear gray on black background

Lakritzator:
You are right, didn't check that... Let me check it.

Lakritzator:
Old:
image

New:
image

Lakritzator:
I guess that tells enough...

Lakritzator:
@jklingen got time to look it over? What do you think?

Arvidas:
For me it's (not) science - shadows must improve information readability.
If shadow is too sharp or on high contrast it adds distraction, visual clutter - informational noise.
Effective shadows often use subtle, soft, and consistent blurs.

Gray shadow issue on dark background should be simple to fix.

Lakritzator:
@Arvidas yes, but what did you want to say with that? I wouldn't have bothered to reproduce your PR if I didn't think your solution is the right one.

I just wanted a second opinion from my colleague, if I overlooks something for which is from my point of view high profile feature (very visible).

jklingen:
Sorry once again... my Github inbox has so much traffic that I often don't see the ones that matter most 😞 

It was bugging me for years that our shadows are bright on dark backgrounds 🙈 so I am totally in for it, even if it means that the shadows aren't visible at all on very dark backgrounds, but I guess that's just what shadows do 😏 

On the screenshots with bright background I like the old version better, because the shadows are softer - but I guess we can do some fine tuning with the constants used in `alpha` and `stepsCount`.

Unfortunately, I cannot do this at the moment, I am having trouble with the latest (unstable) versions on my machine and can no longer start the editor - I'll need to investigate.

Christian-Schulz:
> Unfortunately, I cannot do this at the moment, I am having trouble with the latest (unstable) versions on my machine and can no longer start the editor - I'll need to investigate.

@jklingen Maybe I had the same problem. try this PR #981

Lakritzator:
I made sure the most recent changes were merged into this, **should** work now if there were any issues.

jklingen:
Thanks for the fix @Christian-Schulz . Since no one else had complained I had assumed it would be specific to my machine.

Also thanks @Arvidas for the changes and @Lakritzator for picking them up again. 💪 

So I tried around a bit with the shadows, in my opinion:

* we should definitely keep that shadows are black instead of gray (making the shadows look waaay more natural on dark background)
* we should definitely keep the refactoring to avoid all the copy/paste code

I am just not happy with changing the shadow appearance depending on line thickness, as it makes shadows for thin lines really hard. (When increasing the constants, the shadows for thicker lines start to look weird (that's not because of the changes in here, but because of the way how we draw them in iterations)). 
Also, when a shadow stretches further, it seems as if the object was further away from the canvas/screenshot. To me, it looks as if a thick arrow is nearer to me than a thin arrow.

So, my suggestion is to revert **only that part** of the changes and go back to

```
int basealpha = 100;
int steps = 5;
```

But I am open for other opinions/suggestions.

Here are two screenshorts using the current implementation in this PR vs. changing `basealpha` and `steps` back to constants

### Using Dynamic Values
dynamic

### Using Constant Values
constant

Christian-Schulz:
I would go with the constant values.

jklingen:
No vetoes, so I added the constant values. Thanks everybody who has been involved, especially @Arvidas for your contribution

PR #837, opened 2026-01-19
Amend setup.iss to-fall inline with standard practices

danrhodes:
Updated the Inno Setup script to exclude the application version from the uninstall display name, showing only the executable name.

The version information is already stored in DisplayVersion (set by Inno Setup's AppVersion directive), so administrators and tools can still query the installed version when needed.


Lakritzator:
I get it, valid point. It's good to not always work alone, this would have hit us later.

PR #843, opened 2026-01-20
[Fix] open .greenshot file with Emoji

Christian-Schulz:
Deserialization is aborted with SecurityAccessDeniedException when a file containing emoji.

The EmojiContainer was missing in the SerializationBinder


Lakritzator:
Impressive changes, I need to think about it, the amount of lines changes is overwhelming ;)

PR #844, opened 2026-01-20
[Fix] open older .greenshot files with StepLabels

Christian-Schulz:
If you open an old .greenshot file with `StepLabels` you get an exception because the fields `SHADOW` and `LINE_THICKNESS` are missing.
I added them now while deserialization.

I had to fix `AbstractFieldHolder` as well because it was not fully initialized after deserialization.

And i ran into an edge case. `AddField()` uses the given value only as a default. If a different value is specified in the greenshot.ini file, that value will be used instead. The result was that after opening old files, the step labels could always look different than the time before. 
Because of this i had to set the value actively again.


Lakritzator:
❤️

Issue #471, opened 2023-02-16
Win + Shift + S doesn't work as a hotkey


weary-adventurer:
**Describe the bug**

When hotkeys are set up like this:

```ini
[Core]
WindowHotkey=Win + Shift + S
RegionHotkey=Win + S
```

This is how they appear in the settings:

![image](https://user-images.githubusercontent.com/30081929/219488895-d97660b3-42ad-4666-b392-5d74deac1bda.png)

* Shift + S captures window
* Win + S captures region
* Win + Shift + S doesn't work (prints S if Notepad is in focus)

I also have the Win + S hotkey disabled in the registry so that the default Snipping Tool does not appear and apps are free to use this hotkey:

```reg
Windows Registry Editor Version 5.00

[HKEY_CURRENT_USER\Software\Microsoft\Windows\CurrentVersion\Explorer\Advanced]
"DisabledHotkeys"="S"
```

**Versions (please complete the following information):**
 - Greenshot version 1.2.10 Build 6 (64 bit)
 - Windows version 10 21H2 19044.2604

Taxelas:
I have the same issue
Disabled all apps that are using Win + Shift + S
Edited config file as suggested here:
https://getgreenshot.org/faq/how-can-i-use-the-windows-key-as-global-hotkey-for-greenshot/
`; Hotkey for starting the region capture
RegionHotkey=Win + Shift + S`
Still Win part is omitted.

Win + S works fine

Looks like application doesn't read "Win + Shift/Ctrl  as valid option.
I remember it worked before in Win 10
Current win version:
OS Name	Microsoft Windows 11 Pro
Version	10.0.22621 Build 22621

Lakritzator:
The [current hotkey logic](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-registerhotkey) in Greenshot is somewhat restricted by Windows. I believe the Win + key is mostly ok as long as it's not already taken by Windows itself, but Ctrl / Shift combinations with the Win are no longer allowed.

I am looking at going forward with a different hotkey routine, using a project I maintain (https://github.com/dapplo/Dapplo.Windows),but I didn't get to integrating this into Greenshot yet.

Lakritzator:
> I also have the Win + S hotkey disabled in the registry so that the default Snipping Tool does not appear and apps are free to use this hotkey:

[...]

PR #855, opened 2026-01-23
Support multiple modifiers when using Win modifier in hotkeys

stdlib-dot-h:
Previously, trying to set hotkeys like Win+Alt+S would just cause the app to register a hotkey for Alt+S. I've changed that so now it actually registers Win+Alt+S instead of Alt+S.

2026-01-22 16_21_51-Settings


stdlib-dot-h:
It seems this fixes #471.

Lakritzator:
I have some hesitations about this change, as it is actually documented that one should not use hotkeys with the Win key, These are more or less reserved by Windows and can cause many issues.

And I do not remember why I never made it possible, but I will try it.

Lakritzator:
I was forced to migrate the hotkey handling in #1000 to use direct input, which simply means we are most likely no longer competing hotkeys with other applications. I need to test this change there, due to the complete change of technologies, it doesn't make sense to merge this now and break it with #1000...

stdlib-dot-h:
i'll just close this for now

PR #857, opened 2026-01-24
Pave the road to modern .NET

jairbubbles:
Multi target to `net8.0-windows10.0.19041` **while keeping the `net481` target untouched**. This will allow side by side testing and also ensures that what is now changed/added is future proof.

The minimum is working : **the app runs and we can edit the screenshot**. I didn't test any plugin.

Problem not yet resolved:
- Confluence plugin is compiled only for `net481`
- `ComWrapper` / `IAppVisibility` disabled 
- No installer for new framework
- `App.config` disabled, the winfoms section throws an exception but the other settings probably have no effect.
- There are still some warnings:
  -  `Thread.Abort()` does nothing at all with .NET Core
  -  `WebRequest.Create` should be replaced by an `HttpClient`
  -  `SymmetricAlgorithm.Create()`:  'The default implementation of this cryptography algorithm is not supported.' (https://aka.ms/dotnet-warnings/SYSLIB0007)
  - Some packages only targetting `net4xx`: 'MicrosoftOfficeCore 15.0.0' / 'Unofficial.Microsoft.mshtml 7.0.3300'. 

Keep in mind, **it's just a first step**! But I really think we have a lot to gain to make all this in the main branch. With proper care we can make sure it has no impact on the published version.

Also, I adopted a protective attitude to make sure it's not affecting the current version but some changes are also compatible with .NET Framework and we could reduce from time to time (after proper testing) some `#if` and conditions in the `.csproj` so the code is less divergent.

### DIFF
--- a/src/Greenshot.Plugin.ExternalCommand/ExternalCommandDestination.cs
+++ b/src/Greenshot.Plugin.ExternalCommand/ExternalCommandDestination.cs
@@ -260,7 +260,7 @@ namespace Greenshot.Plugin.ExternalCommand
                 }
 
                 LOG.InfoFormat("Starting : {0} {1}", process.StartInfo.FileName, process.StartInfo.Arguments);
-                Process.Start(new ProcessStartInfo() { UseShellExecute = true });
+                process.Start();
                 process.WaitForExit();
                 if (config.RedirectStandardOutput)
                 {


Lakritzator:
Let's agree that before bigger steps are taken, you create an issue discussing it.
We already did a lot of things behind the scene in a different project, because taking such changes with us, places us in a bit of a split. Currently I am trying to get Greenshot working again, there are some places which are broken, so we can release something which helps the people waiting for that. .NET is on our roadmap, and there are many very cool possibilities coming from making that step, you especially know about the ImageSharp restrictions. It's the way we need to take...

But we have some obstacles which are very big:

- File size, the Greenshot installer will jump up from <5MB to between 20MB for a separate .NET installation to >120MB for a independent installation. This is not only something the users need to get used to, we need to make sure our infrastructure is okay with that. On top of it, I already had some very negative installation experiences, as from a user perspective it was just sitting there and doing "nothing".
- There is old tech which will not work, and we need to see what direction we go. For most there is a solution, and we are in fact already making such things possible in other issues.

My biggest point currently is time, e.g. having issues with the office plugin, most services do not work, I need to get Confluence either working or I need to kick it out. That is where I currently need to focus on... I will look at it in the comings days, but having another .NET version to build for makes everything slower and bigger.

jairbubbles:
I completely understand @Lakritzator that it's not your current focus. It doesn't mean we actually ship Greenshot any time soon to .NET 8, but mainly to make sure that it's future proof.

I discussed it a bit with @Christian-Schulz over here: https://github.com/Christian-Schulz/greenshot/discussions/2 I ended up creating this PR so that we see clearly what it involves.

Also, if you need help on a subject you listed let me know.

Issue #859, opened 2026-01-24
MAPI Client destination: Wrong name/or not detected if in HKCU


htcfreek:
**Describe the bug**
If the MAPI Client setting is located in the registry in HKCU the following symptoms can happen:
- HKCU only: MAPI Client not detected.
- HKCU and HKLM: Name of HKLM and open app from HKCU.

HKCU entry can happen for example if an Mail Client sets the Mail App association as "current user".

**To Reproduce**
Example steps to reproduce the behavior:
1. Add MAPI-Value in registry in HKCU.
2. Capture something.
3. Send to Mail.

**Expected behavior**
Client is detected and opens the client shown in the menu.

**Screenshots**
-/-

**Versions (please complete the following information):**
 - Greenshot version: 1.3.312
 - Windows version: Win10 22H2 ESU

**Additional context**

htcfreek:
I will open a PR to fix.

Lakritzator:
I guess you fixed it yourself 😂
Closing this, should be available and ready for testing in the latest build (in the next ~ 20 minutes tops)

PR #861, opened 2026-01-24
Fix Mapi Client detection on HKCU

htcfreek:
### Description
Some mail clients write the Mapi key in HKCU and not in HKLM.  This happens mostly if the user has no admin permission and instructs the mail client to set itself as default app.

This PR adds code to handle the case where the key is located in HKCU only or in both places (HKCU and HKLM) with preferring the value from HKCU.

### Changes

1. Implemented a ReadKey() extension for reading HKCU value.
2. Updated the EmailDestination code to respect both base hives.

### Closes
Issue #859

### DIFF
--- a/src/Greenshot/Destinations/EmailDestination.cs
+++ b/src/Greenshot/Destinations/EmailDestination.cs
@@ -43,11 +43,21 @@ namespace Greenshot.Destinations
         static EmailDestination()
         {
             // Logic to decide what email implementation we use
-            _mapiClient = RegistryHive.LocalMachine.ReadKey64Or32(@"Clients\Mail");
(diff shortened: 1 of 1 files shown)


Lakritzator:
Unfortunately I do not use email clients which register as MAPI clients, so I cannot directly test this.
But it looks quite good, and I assume you made the changes as you had the issue yourself, therefor I can also assume it works.

Thanks for supporting us!! ❤️

htcfreek:
Should I create a PR for 1.3 branch/release too?

danrhodes:
No need to do that it will all get handled from main. Love the fix 🚀❤️

PR #863, opened 2026-01-26
Improved cursor capture logic to reflect real size

Lakritzator:
Improved the way we take the mouse cursor for the capture. Before the cursor was the small default one, now we look at the settings and take the size as the user has configured. This was a long standing issue...

In fact, I was reminded of this due to a google notification, which was trigged due to this here: https://www.computerbase.de/forum/threads/screenshot-von-genau-dem-was-man-auf-sieht-inkl-korrekter-cursor-groesse.2264060/

Now it should be fixed, this is before my change (my mouse cursor was set to HUGE):
image

And this after, which makes it look exactly like it was on the screen: 
image

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -50,7 +50,7 @@
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
     </PackageReference>
-    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="8.0.0" PrivateAssets="All"/>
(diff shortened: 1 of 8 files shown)


Lakritzator:
There might still be issues with multi-monitor and different DPI settings, which I will check, but I will merge this to make sure we do not get conflicts here.

To be clear, most of the logic for this is in Dapplo.Windows, this is why the change is relatively small in Greenshot. The complete code change is documented here: https://github.com/dapplo/Dapplo.Windows/commit/28398503a6dcd7ef8ddaa14b6f3c8ca4516ee25d

PR #864, opened 2026-01-26
Fix pointer conversion for bitmap data

danrhodes:
Changed Scan0 pointer conversion from ToInt32() to ToInt64() to correctly handle 64-bit environments and prevent potential data loss or errors when working with bitmap data.


Lakritzator:
Btw Windows 10 still supports 32 bit... but this should just work, only slower.

danrhodes:
Yeah I checked and it's full compatible on 32bit

PR #865, opened 2026-01-26
Dispose GDI resources on application exit

danrhodes:
Registers cleanup for static brushes and pens used in CaptureForm by disposing them when the application exits. Prevents potential resource leaks from undisposed GDI objects.


Lakritzator:
I never found these, as the way I searched for leaks didn't involve the stuff which is thrown again anyway, but yeah it does make sense 👍

PR #866, opened 2026-01-26
Refactor zoomer opacity ImageAttributes handling

danrhodes:
Moved ImageAttributes creation for zoomer opacity into a dedicated method and replaced manual disposal with a using declaration to ensure proper resource management. This improves code clarity and reliability by encapsulating the opacity logic.

### DIFF
--- a/src/Greenshot/Forms/CaptureForm.cs
+++ b/src/Greenshot/Forms/CaptureForm.cs
@@ -852,22 +852,10 @@ namespace Greenshot.Forms
                 return;
             }
 
-            ImageAttributes attributes;
-
-            if (_isZoomerTransparent)
-            {
-                //create a color matrix object to change the opacy
-                ColorMatrix opacyMatrix = new ColorMatrix
-                {
-                    Matrix33 = Conf.ZoomerOpacity
-                };
-                attributes = new ImageAttributes();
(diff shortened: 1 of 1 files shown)


src/Greenshot/Forms/CaptureForm.cs
```diff
@@ -852,22 +852,10 @@ private void DrawZoom(Graphics graphics, NativeRect sourceRectangle, NativeRect
...
-                };
-                attributes = new ImageAttributes();
-                attributes.SetColorMatrix(opacyMatrix, ColorMatrixFlag.Default, ColorAdjustType.Bitmap);
-            }
-            else
-            {
-                attributes = null;
-            }
+            // Create ImageAttributes for transparent zoomer (using declaration ensures disposal even if exception occurs)
+            using ImageAttributes attributes = _isZoomerTransparent
```
Lakritzator, line 856: I am not the biggest fan of "using", where the compiler more or less decides when things get cleaned up... but it does work... so why not 🤷‍♂️

Lakritzator:
It doesn't really solve something, but I am all for better code.

src/Greenshot/Forms/CaptureForm.cs
Christian-Schulz: > where the compiler more or less decides when things get cleaned up...

Small friendly note: 
The object `attributes` is disposed of exactly at the end of the function (on return or on exception). No decision is necessary.

PR #867, opened 2026-01-26
Handle unsupported COM event subscription gracefully

danrhodes:
Added logging and safe return for COM event subscription and unsubscription attempts. This avoids breaking callers by warning that event handlers on COM objects are not implemented, rather than throwing errors.

### DIFF
--- a/src/Greenshot.Base/Interop/COMWrapper.cs
+++ b/src/Greenshot.Base/Interop/COMWrapper.cs
@@ -364,12 +364,20 @@ namespace Greenshot.Base.Interop
             else if (1 == argCount && typeof(void) == returnType && (methodName.StartsWith("add_") || methodName.StartsWith("remove_")))
             {
                 bool removeHandler = methodName.StartsWith("remove_");
-                methodName = methodName.Substring(removeHandler ? 7 : 4);
-                // TODO: Something is missing here
+                string eventName = methodName.Substring(removeHandler ? 7 : 4);
+
                 if (!(callMessage.InArgs[0] is Delegate handler))
                 {
                     return new ReturnMessage(new ArgumentNullException(nameof(handler)), callMessage);
                 }
+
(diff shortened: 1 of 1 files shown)


danrhodes:
Apologies for the four PRs in quick succession they were sat on my notepad as I've not had time recently to create them.

Lakritzator:
Can you explain this to me? Just to be clear, the whole COM stuff I got it working by trying and fixing errors, it was never my best topic... It's more or less a miracle it works, mostly...

But what does this change fix? When does this throw an error, or is it just cosmetics?

Lakritzator:
Btw... GitHub actions seems to act up, saw it at my other projects before... 🙁

danrhodes:
Hey, I'll add more context to the PR @Lakritzator there was a specific reason for this that I can't recall right now but I have a note about this on my laptop.

danrhodes:
Evening,

So when Greenshot interacts with Office apps via COM (like th Outlook plugin), it uses this COMWrapper class to proxy calls. The wrapper handles things like property gets/sets and method invocations, but there was also a code path for event subscription (the `add_` and `remove_` handlers) that was never finished - just had a TODO comment.

The problem is if any code tried to subscribe to a COM event, it would hit this incomplete path and either error or behave unexpectedly. This change makes it fail gracefully - logs a warning saying “COM events aren’t supported” and returns cleanly instead of erroring.

Honestly, I can’t remember the exact scenario that triggered this for me. The fix doesn’t add anything other than it just stops it from breaking when something tries to use it.

danrhodes:
> Btw... GitHub actions seems to act up, saw it at my other projects before... 🙁

Is something broken on continuous builds?

Lakritzator:
I skipped building, using [skip ci] in the commit message, as I didn't want to have 7 builds for small changes...

PR #871, opened 2026-01-26
Trying to fix the JIRA plugin

Lakritzator:


Lakritzator:
This is really weird, using await in the OnLoad of the Jiraform closes the ShowDialog call directly with DialogResult.Cancel.

Been trying to tackle the issue, but I really do not have a clue, looking again another day.

Lakritzator:
@Christian-Schulz or @danrhodes does anyone of you use the Jira plugin? I can't seem to find why the upload form closes.

These are the steps (if you can't do anything with this, that's okay, need to document them anyway):

1. Configure the plugin with credentials (email / api token)
2. Capture something
3. Select the jira destination (main entry)
4. The upload screen comes up
5. cancel it
6. (re)load a JIRA issue in the browser
7. Capture something, move to the jira destination, see the issue, but click on the main jira destination entry entry 
8. The form shows, and it's gone again, the destination picker sees "cancel"... WHY?

When debugging this happens as soon as an await is used, but I do not see any exceptions.

Christian-Schulz:
Unfortunately, no. 
I don't use the plugin. I also don't have an environment to test it.

danrhodes:
Sadly no I don't use Jira either, sorry. Maybe reach out to the wider community somehow?

Lakritzator:
It more or less doesn't have to do with JIRA specifically, those things I can handle, it seems I am still in the top 3 of nuget JIRA packages.

I will try to reduce the issue to "work" (or rather not work) without jira, so far it looks like it's an await which breaks the UI...

danrhodes:
Happy to take a look if you would like?

Lakritzator:
I wanted to make it easier to trigger the issue, let me try that and if I still don't see it, I'd be very grateful

Lakritzator:
@Christian-Schulz after you did #911 I noticed the behavior for the Jira Plugin improved (it wasn't weird anymore) but there was a problem with the context menu, as it was disposed it causes weird effects... I changed from Dispose to Close and now things seem to work.

Lakritzator:
Sorry for the AbstractDestination, somehow the merge is weird now... so the diff shows that everything changed... odd.

PR #874, opened 2026-01-27
Memory Leak - Dispose _image in Surface.Dispose method

danrhodes:
The Surface class holds an _image field containing the captured screenshot, but it was never disposed in the Dispose(bool) method. Since screenshots can be several megabytes each, this caused significant memory leaks when surfaces were created and destroyed (e.g., when opening/closing the editor).

Added disposal of _image in the Dispose method, following the same pattern used for other disposable fields (_buffer, _transparencyBackgroundBrush, etc.).

### DIFF
--- a/src/Greenshot.Base/Core/Capture.cs
+++ b/src/Greenshot.Base/Core/Capture.cs
@@ -113,7 +113,7 @@ namespace Greenshot.Base.Core
             set
             {
                 _cursor?.Dispose();
-                _cursor = (Bitmap) value.Clone();
+                _cursor = (Bitmap) value?.Clone();
             }
         }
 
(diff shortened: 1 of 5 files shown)


Lakritzator:
I really want to believe I somehow forgot... and screwed up the dispose.
BUT I can't suppress the feeling this might be more complex then it looks... why would I forget to dispose the most obvious item?
Did you try to undo / redo after making this change?

danrhodes:
I'll be honest, no I didn't so don't merge and I will try tomorrow as I've logged off for the evening now. 👍

Lakritzator:
No problem, I'll have a look... I hope it doesn't give an issue, otherwise we are sooooo screwed... 😂

danrhodes:
I hope it doesn't break undo redo too as if it does there is a much deeper issue that's going to be fun to fix 😭🤣

I will test tomorrow and report back.

Lakritzator:
I didn't see any issues, I'm going for it... 🤞

Lakritzator:
This also triggers the pre-release build, didn't skip it... 😉

PR #876, opened 2026-01-27
Dispose unused file streams in ClipboardHelper

danrhodes:
In ClipboardHelper.IterateFileDescriptors(), a MemoryStream is created for each file descriptor, but it was only yielded to the caller when fileData?.Length > 0. In cases where:

The stream was created but had zero length
The stream was created but an exception occurred during content reading The MemoryStream was never disposed, causing a memory leak.

Added an else branch to dispose the MemoryStream when it won't be yielded to the caller (i.e., when it's empty or has zero length).


Lakritzator:
I owe you more then one beer... 😱

danrhodes:
You really don't, I wish I could help more so happy to keep donating my free time to a tool I use all day 🚀👍🎉

PR #877, opened 2026-01-27
Refactor clipboard retry logic in ClipboardHelper

danrhodes:
The GetDataObject() retry loop had a logic flaw where retryCount was decremented in a finally block, which executes on both success and failure. This meant:

On success, the counter was unnecessarily decremented before returning The loop structure was confusing - using while (retryCount >= 0) with a decrement in finally made it hard to understand the actual retry count.

Refactored to use a cleaner for loop pattern:

Clear maxRetries constant (3 attempts total)
Loop counter only advances on each iteration (not in finally) isLastAttempt boolean makes the intent clear
Same functional behavior (3 attempts with 100ms delay between failures)

### DIFF
--- a/src/Greenshot.Base/Core/ClipboardHelper.cs
+++ b/src/Greenshot.Base/Core/ClipboardHelper.cs
@@ -208,8 +208,8 @@ EndSelection:<<<<<<<4
         {
             lock (ClipboardLockObject)
             {
-                int retryCount = 2;
-                while (retryCount >= 0)
+                const int maxRetries = 3;
+                for (int attempt = 0; attempt < maxRetries; attempt++)
                 {
                     try
                     {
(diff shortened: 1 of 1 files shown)


Lakritzator:
I had in mind to replace all Clipboard logic that from my Dapplo.Windows project:
https://github.com/dapplo/Dapplo.Windows/blob/28398503a6dcd7ef8ddaa14b6f3c8ca4516ee25d/doc/articles/clipboard-usage.md

This would even allow to only put something on the clipboard IF it was requested by another application (delayed rendering).

I guess the part you tripped on, is something like the logic here:
https://github.com/dapplo/Dapplo.Windows/blob/28398503a6dcd7ef8ddaa14b6f3c8ca4516ee25d/src/Dapplo.Windows.Clipboard/Internals/ClipboardSemaphore.cs#L58

I agree with your code, but it wasn't that horrible 😁

danrhodes:
I must admit the work you are doing over at Dapplo looks very impressive 👌

PR #887, opened 2026-01-28
Use 'using' statements for GDI+ resources in CaptureForm

danrhodes:
Problem: In the OnPaint method, bgBrush and rulerPen are created without using statements and disposed manually at the end of the block. If an exception occurs between creation and disposal, these GDI resources will leak.

Solution: Changed bgBrush and rulerPen declarations to use using statements, ensuring proper disposal even if an exception occurs. Removed the now-redundant manual .Dispose() calls.

Impact: Ensures GDI resources are properly disposed in all code paths, preventing potential resource leaks during exception scenarios in the paint handler.


Lakritzator:
Makes sense, I almost think the using keyword was introduced after I wrote that code 😂

PR #888, opened 2026-01-28
Dispose capture object in Dispose method

danrhodes:
The Dispose(bool) method in CaptureHelper sets _capture = null without disposing it first. While some code paths dispose _capture explicitly (e.g., after successful capture), other paths - notably when user cancels the capture via CaptureWithFeedback() - leave _capture undisposed, causing an image memory leak.

Added _capture?.Dispose() in the Dispose(bool) method to ensure the capture is always disposed when the CaptureHelper is disposed. Since ICapture.Dispose() is idempotent (safe to call multiple times), this won't cause issues in code paths that already dispose the capture.

Prevents memory leaks when users cancel captures, ensuring the captured image is properly released in all scenarios.

### DIFF
--- a/src/Greenshot/Helpers/CaptureHelper.cs
+++ b/src/Greenshot/Helpers/CaptureHelper.cs
@@ -83,10 +83,10 @@ namespace Greenshot.Helpers
         {
             if (disposing)
             {
-                // Cleanup
+                // Dispose the capture if it hasn't been disposed in another code path
+                _capture?.Dispose();
             }
 
-            // Unfortunately we can't dispose the capture, this might still be used somewhere else.
             _windows = null;
             _selectedCaptureWindow = null;
             _capture = null;


Lakritzator:
I'm VERY hesitant to do this without investing a bit more thought, I added the line with reasons:
// Unfortunately we can't dispose the capture, this might still be used somewhere else.

You're sticking a stick in a bee nest... 🐝

There is in fact still something broken which needs fixing, and I'm not even talking about memory leaks.
Since the SetupAutoDispose was introduced by @Christian-Schulz , I have issues with the destination picker.

I use the destination picker all the time, this is the flow:

1. Capture something
2. Pick a destination
3. Sometimes I cancel the destination, e.g. save as, and want to place it on the clipboard
4. The picker comes back and I can select clipboard

Since the changes this no longer works reliably.
Why do I mention it here? Because this means that the capture needs is available for a longer time, and I am no longer sure this is in the scope of the life-time of the CaptureHelper or beyond.

Lakritzator:
Maybe we need a clean process (or flow), maybe not directly linked to the UI, for handling the capture.
Where did I see this again, ah yes, in the develop branch of Greenshot I started that. 😂

I completely (well not completely as it wasn't complete yet) rewrote the way Greenshot works internally: https://github.com/greenshot/greenshot/tree/develop/src/Greenshot.Core

Sources are where a "screenshot" comes from, can be screen, file, cursor, clipboard etc.
Processors are things which modify that which a source provides, e.g. add ocr, automatically obfuscate sensitive information, use an LLM to do whatever, etc.
Destinations are like we already have.

Different Flows could be triggered by different triggers, e.g. hotkeys or have something trigger when an image is on the clipboard, or a certain application which is in focus... who knows.

I wanted to add the possibility to define recipes, which can be provided by the community.

Christian-Schulz:
I found the reason why the destination picker didn't reopen after canceling the action. 
The closed event is called even before we manually call menu.Close().

I write a PR in the next days.

PR #889, opened 2026-01-28
Handle exceptions during cursor capture

danrhodes:
I had a good 10+ of these errors today so have created this to be reviewed which should sort the issue. I use a green cursor at work for vision and i suspect this is possibly the reason why.

Software version: 1.4.68 - 1cb951119f (64 bit)
Time: 2026-01-28 08:08:45 +00:00
OS: Windows 11 x64 build 26100
GDI object count: 60
User object count: 23
Exception: System.ComponentModel.Win32Exception
Message: Invalid cursor handle
ErrorCode: 0x80004005
Stack:
   at Dapplo.Windows.Icons.CursorHelper.TryGetCurrentCursor[TBitmapType](TBitmapType& returnBitmap, NativePoint& hotSpot)
   at Greenshot.Base.Core.WindowCapture.CaptureCursor(ICapture capture)
   at Greenshot.Helpers.CaptureHelper.MakeCapture()
   at Greenshot.Helpers.CaptureHelper.CaptureRegion(Boolean captureMouse)
   at Greenshot.Base.Controls.HotkeyControl.HandleMessages(Message& m)
   at Greenshot.Forms.MainForm.WndProc(Message& m)
Configuration dump:

Wrapped the cursor capture logic in a try-catch block to handle potential Win32Exceptions. If capturing the cursor fails, a warning is logged and the capture proceeds without the cursor, improving robustness.

### DIFF
--- a/src/Greenshot.Base/Core/WindowCapture.cs
+++ b/src/Greenshot.Base/Core/WindowCapture.cs
@@ -86,17 +86,24 @@ namespace Greenshot.Base.Core
                 capture = new Capture();
             }
 
-            Bitmap cursorBitmap;
-            NativePoint cursorHotSpot;
-            if (CursorHelper.TryGetCurrentCursor(out cursorBitmap, out cursorHotSpot))
+            try
             {
-                NativePoint cursorLocation = User32Api.GetCursorLocation();
-                // Align cursor location to Bitmap coordinates (instead of Screen coordinates)
-                var x = cursorLocation.X - cursorHotSpot.X - capture.ScreenBounds.X;
-                var y = cursorLocation.Y - cursorHotSpot.Y - capture.ScreenBounds.Y;
-                // Set the location
-                capture.CursorLocation = new NativePoint(x, y);
-                capture.Cursor = cursorBitmap;
+                Bitmap cursorBitmap;
+                NativePoint cursorHotSpot;
+                if (CursorHelper.TryGetCurrentCursor(out cursorBitmap, out cursorHotSpot))
+                {
+                    NativePoint cursorLocation = User32Api.GetCursorLocation();
+                    // Align cursor location to Bitmap coordinates (instead of Screen coordinates)
+                    var x = cursorLocation.X - cursorHotSpot.X - capture.ScreenBounds.X;
+                    var y = cursorLocation.Y - cursorHotSpot.Y - capture.ScreenBounds.Y;
+                    // Set the location
+                    capture.CursorLocation = new NativePoint(x, y);
+                    capture.Cursor = cursorBitmap;
+                }
+            }
+            catch (System.ComponentModel.Win32Exception ex)
+            {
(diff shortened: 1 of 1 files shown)


Lakritzator:
Can you explain me how you set this cursor?
We should definitively not catch an exception, as there is fallback logic in place, but obviously not good enough.

danrhodes:
Sure, it's part of Windows 11.

Open settings
Go to ease of access
Under vision select cursor and pointer
From here you can set colour and size
image

danrhodes:
I was maybe too hasty with this.

On reflection the code should handle the exception inside TryGetCurrentCursor() and return false, rather 
than catching.

The Try pattern should signal  failure via the return value.

I think maybe a better way would be to refactor to move the exception handling into TryGetCurrentCursor() 
itself? Views?

Lakritzator:
Yeah it has to be fixed inside the method, but it should in fact have a fallback to the old logic. I couldn't reproduce it though. 

I think the handle here is already invalid, but I didn't validate it, so it might throw invalid cursor handle: 
https://github.com/dapplo/Dapplo.Windows/blob/master/src/Dapplo.Windows.Icons/CursorHelper.cs#L73

Lakritzator:
I'll check it Friday, have another business trip to do, if I'm feeling fit tomorrow(kid is spreading bacteria like crazy)

danrhodes:
I'll also confirm tomorrow if it is indeed the green cursor causing the issue by reverting back to a standard cursor.

Argh, yeah, it's this time of year coughs and colds spread like wildfire.

Christian-Schulz:
Hey, you're also in the #GreenMouseGang? I thought I was the only one. 🤣
image

danrhodes:
Ha ha no I'm part of that crowd too so much easier to see on screen 🤣

Also what version are you running as that is clearly a capture of a green cursor!

Lakritzator:
For me it's working too, I already expected it's something else (additionally):
image

Lakritzator:
I think I found it, let me provide a fix in Dapplo.Windows

Christian-Schulz:
> Also what version are you running as that is clearly a capture of a green cursor!

I think it was v1.4.78

Lakritzator:
The error was that I disposed the hCursor in the CursorInfo, but this is shared. Who knows what happens when if someone does that... With the update it should be fixed.

danrhodes:
Legend 👌🚀

danrhodes:
@Lakritzator tested today and your Dapplo fix worked perfectly no more bugs today!

Lakritzator:
It was an unneeded bug, I wasn’t aware of the shattered state of the handle und documentation was very limited. Happy it worked

danrhodes:
Unsure if it helps but the cause I found out was a web based remote access too we use to remotely control servers, as I tried to screenshot my changes it caused the crash.

PR #890, opened 2026-01-28
[Fix] Add emoji files in portable app

### LINKED ISSUE #840: Unfortunately, the editor won't open.
After I took a screenshot, the editor doesn't open.

Christian-Schulz:
This fixes #840

The emoji files were missing in the output folder.


Lakritzator:
Good catch, I never use the portable

PR #911, opened 2026-02-03
[Fix] destination picker

Christian-Schulz:
The destination picker menu was disposed too early. 
If an export was canceled, the menu could not be reopened.

### DIFF
--- a/src/Greenshot.Base/Core/AbstractDestination.cs
+++ b/src/Greenshot.Base/Core/AbstractDestination.cs
@@ -183,8 +183,6 @@ namespace Greenshot.Base.Core
                 TopLevel = true
             };
 
-            menu.SetupAutoDispose();
-
             menu.Opening += (sender, args) =>
             {
                 // find the DPI settings for the screen where this is going to land
@@ -211,6 +209,7 @@ namespace Greenshot.Base.Core
                         else
                         {
                             Log.DebugFormat("Letting the menu 'close' as the tag is set to '{0}'", menu.Tag);
+                            menu.Dispose();
                         }
 
                         break;
@@ -219,13 +218,13 @@ namespace Greenshot.Base.Core
                         // The ContextMenuStrip can be "closed" for these reasons.
                         break;
                     case ToolStripDropDownCloseReason.Keyboard:
-                        // Dispose as the close is clicked
+                        // Menu closed via keyboard (e.g., ESC key)
                         if (!captureDetails.HasDestination("Editor"))
                         {
                             surface.Dispose();
                             surface = null;
                         }
-
+                        menu.Dispose();
                         break;
                     default:
                         eventArgs.Cancel = true;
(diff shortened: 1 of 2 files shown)


Lakritzator:
That looks good, and sounds good too... this was something I kept having as an issue while trying to fix the jira plugin and also the share destination. Just didn't want to get off track with my work, so thanks for fixing it.

src/Greenshot.Base/Core/AbstractDestination.cs
```diff
@@ -211,6 +209,7 @@ public ExportInformation ShowPickerMenu(bool addDynamics, ISurface surface, ICap
                         else
                         {
                             Log.DebugFormat("Letting the menu 'close' as the tag is set to '{0}'", menu.Tag);
+                            menu.Dispose();
```
Lakritzator, line 212: I think this should haven been a close, see #871

Christian-Schulz: I thought that in the AppFocusChanged event, we allow the menu to close.  Since it will never be opened again after that. 
Isn't that right?

Lakritzator: The focus would change if the menu is selected and a new window pops up, e.g. the share, Jira etc... if this is cancelled I want to be able to go back to the menu again, as I do not want to loose the capture. The only reason to dispose is ECS, Cancel or a successful export.

Christian-Schulz: I tested to cancel all export plugins. But I never got to this code line. 
Can you describe your exact steps?

Lakritzator: It's a bit weird behavior, as with the limited testing I did, can't explain why it is triggered with certain actions and with others it's not. Maybe you can figure it out.

I saw it with the JIRA destination, BUT only if the destination has sub-destinations.
And this is actually the best way of testing this, using a destination with a sub-destination, if found it works when using the editor or printer, editor is the easiest to test.

Capture and open an editor. Capture again and select the sub-destination for the editor you just opened (btw. that's a not well known feature, it will add the capture a bitmap container to the other capture)
At the moment you select the menu item, that line is triggered.

Now with the editor you cannot cancel, with the printer you can if you have the bottom printer setting ...
image

Why I find it weird? I would have expected the line to trigger with clicking the main item too. 🤷‍♂️
And actually, THAT is what happens with the Jira destination, but ONLY if there is sub item. (which only shows up if you actually configured it to use a JIRA server, AND looked at a JIRA issue in the browser)

Christian-Schulz: Next week I may be very busy at work, so it may take a while before I can test it.

Christian-Schulz: Yes you were right. I tested your steps. 
And found an other bug. #958

Issue #534, opened 2024-03-27
Greenshot not compatible with new Outlook version (beta)


Johnsinoo:
We at our organization received a complaint that the ‘Microsoft Outlook ’ shortcut in the greenshot menu is not working with the newest version of Outlook. Perhaps this is because the outlook process changed names from Outlook.exe to olk.exe in the newest version?

I’d kindly want to request it as a feature in an upcoming version of Greenshot.

Example steps to reproduce the behavior:
1. Open Outlook and press the button top right to switch to the newest version of Outlook.
2. Make a screenshot with Greenshot and select the 'Microsoft Outlook' function in the menu.
3. You'll notice it's trying it's best to open the screenshot in a new mail, but visually, it's not doing anything. For X-amount of seconds, you are not able to use Greenshot.

When you normally select the 'Microsoft Outlook' function in the menu, it should open the image inside a new e-mail.

The stable version of Greenshot 1.2.10.6 has this problem. I then I tried with the latest version of Greenshot 1.3.277 and it did also not work. This is both an issue on Windows 10 and Windows 11 PC's.

Omarlilyo:
Was this ever solved?

drizzt09:
This is still an issue and New Outlook is more usable now and more are starting to roll to it.
Can this get some support please?

Omarlilyo:
Yep - still an issue here too, and all my staff have moved to new outlook - so i can no longer get them to use greenshot and they are now using windows snipper - which is worse.

reconsierra:
Just adding my vote to this as we have just rolled to the new Outlook globally and this was a disappointing surprise as I used this a lot.

Thanks :)

danrhodes:
I'll take a look later should just need the new version adding in of it's not there

Omarlilyo:
Thanks @danrhodes

Lakritzator:
@danrhodes I am not sure if we need to do something extra...
For me it works for me with the current 1.4 builds, but I am not sure what Outlook version we are having issues with...

(I need to check all the Office apps, if these still work...)

danrhodes:
I'll have a look to see if it's working already on a later build.

Omarlilyo:
@danrhodes  @Lakritzator  

There is "classic" outlook, and "new" outlook.

In typical microsoft fashion, there is very little difference in the icons and they are very easy to confuse.

My current greenshot 1.3.312 (and every one before that) only lets us post out to "classic" outlook. Not the New one.

[...]

PR #919, opened 2026-02-05
Fix Windows Share & Rename OCR to work with Windows 10 & 11

danrhodes:
Rename the "Windows 10 Share" and "Windows 10 OCR" features to just "Windows Share" and "Windows OCR" now that Win10 is EOL.

Fixed the share feature not working on Windows 11 - turns out there were two issues: a missing .resx file was causing a silent MissingManifestResourceException that made the whole thing look like it did nothing, and the COM interop was passing the wrong GUID (IUnknown instead of the DataTransferManager IID) to GetForWindow.

Updated the interface to return DataTransferManager directly instead of messing around with raw IntPtr marshalling, which matches what Microsoft's own reference sample does.

Also made the SharingForm invisible since it's just there to own the HWND for the share dialogue. Internal designation strings left as-is so existing user configs don't break.

https://github.com/greenshot/greenshot/issues/534 - Fixing GS to use New Outlook doesn't look like it will work given that New Outlook is a Web App; however, the best solution looks to be to update GS to use Windows Share in Windows 11 to work.

Works both from the editor and from having Windows Share as a destination.

Backwards compatibility with Windows 10 is maintained.

image

### DIFF
--- a/src/Greenshot/Forms/SharingForm.Designer.cs
+++ b/src/Greenshot/Forms/SharingForm.Designer.cs
@@ -50,14 +50,15 @@ namespace Greenshot.Forms {
 		/// not be able to load this method if it was changed manually.
 		/// </summary>
 		private void InitializeComponent() {
-            System.ComponentModel.ComponentResourceManager resources = new System.ComponentModel.ComponentResourceManager(typeof(SharingForm));
             this.SuspendLayout();
-            // 
+            //
             // SharingForm
-            // 
-            this.ClientSize = new System.Drawing.Size(284, 261);
-            this.Icon = ((System.Drawing.Icon)(resources.GetObject("$this.Icon")));
+            //
+            this.ClientSize = new System.Drawing.Size(0, 0);
+            this.FormBorderStyle = System.Windows.Forms.FormBorderStyle.None;
             this.Name = "SharingForm";
+            this.Opacity = 0;
+            this.ShowInTaskbar = false;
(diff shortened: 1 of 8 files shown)


Lakritzator:
I don't know what happened, I fixed it while merging in https://github.com/greenshot/greenshot/commit/c066830acc86c17248c1cc31ed665d8583e357d6#diff-a17ff5dde6fac2569d45ae04f78080dcaddcf0cfe27c09b0fdae469c4892da4d
I even tested it, and it worked flawless... Anyhow, sometimes it's better to move on...

src/Greenshot/Forms/SharingForm.Designer.cs
```diff
@@ -50,14 +50,15 @@ protected override void Dispose(bool disposing) {
...
-            System.ComponentModel.ComponentResourceManager resources = new System.ComponentModel.ComponentResourceManager(typeof(SharingForm));
             this.SuspendLayout();
-            // 
+            //
             // SharingForm
-            // 
-            this.ClientSize = new System.Drawing.Size(284, 261);
-            this.Icon = ((System.Drawing.Icon)(resources.GetObject("$this.Icon")));
+            //
+            this.ClientSize = new System.Drawing.Size(0, 0);
```
Lakritzator, line 57: I was thinking that if something went wrong, at least something would be visible, I believe I even had a cancel button on the form... but to be honest, I don't care to much about that, it should just work.

PR #921, opened 2026-02-05
Included support for ImageSharp decoders / encoders

Lakritzator:
Implemented ImageSharpFileFormatHandler to handle all reading/writing of the formats that ImageSharp supports.

Placed this behind a BetaTester feature toggle, which is currently only available via the greenshot.ini file.

### DIFF
--- a/src/Greenshot.Base/Core/ImageIO.cs
+++ b/src/Greenshot.Base/Core/ImageIO.cs
@@ -29,7 +29,6 @@ using System.Reflection;
 using System.Runtime.InteropServices;
 using System.Text;
 using System.Text.RegularExpressions;
-using System.Web;
 using System.Windows.Forms;
 using Greenshot.Base.Controls;
 using Greenshot.Base.Core.Enums;


src/Greenshot.Base/Core/ImageIO.cs
```diff
@@ -571,13 +571,10 @@ public static Image FromStream(Stream stream, string extension = null)
                 return null;
             }
 
-            if (!string.IsNullOrEmpty(extension))
-            {
-                extension = extension.Replace(".", string.Empty);
-            }
+            var fileFormatHandlers = SimpleServiceProvider.Current.GetAllInstances<IFileFormatHandler>();
+            var supportedExtensions = fileFormatHandlers.ExtensionsFor(FileFormatHandlerActions.SaveToStream).ToList();
```
Christian-Schulz, line 575: Copy&Paste ? 
`supportedExtensions` is never used.

Christian-Schulz:
Have the file sizes of PNG files changed?

src/Greenshot.Base/Core/ImageIO.cs
Lakritzator: yes.. but I need to clean up the extension part anyway, having the OutputFormat enum makes things a bit inflexible.

Lakritzator:
What do you think, of course!
I captured the edge browser default background image, for me some photo of Tossa de Mar in Spain.
Opened this in the editor and then saved it twice:

1. With System.Drawing it was 2.149.510 bytes
2. With ImageSharp it was 1.541.733 bytes

Issue #723, opened 2025-11-20
Multiple Counters


EsRa13:
The already quite good Greenshot would be even better if it were possible to have multiple counter series. For example, start a new counter with a different color from starting with No. 1. This probably means that there must also be a way to continue an existing counter series (e.g., the red one), e.g., by right-clicking >> “Continue counter.”

Lakritzator:
Interesting idea, maybe we will come to that, but currently I wouldn't know the best approach for the UI. 🤷‍♂️

Christian-Schulz:
I agree. I've missed this feature several times already.
I don't have a practical solution for the UI either.

_Additional technical information: Internally, the StepLabels-Counter is very tightly linked to the Editor.Surface and is not as independent as all other containers. This would require major reimplementations._

EsRa13:
Perhaps a simple (or poor) solution could be to install a second counter that counts A, B, C etc.
Yes, it's really poor 🙈

EsRa13:
How about this?
Counters get a new attribute “type” in addition to fillcolor, linecolor and size.
Each time the “type” icon is clicked, the type of the counter changes (in a loop): e.g., "number, uppercase letters, lowercase letters, Roman numeral, etc.).
Greenshot must save the current counter value for each type in the background. The next time the currently set "type" is used, it is incremented.

![Image](https://github.com/user-attachments/assets/57ff0a1d-ed56-4774-ac80-27898a828066)

Issue #103, opened 2018-11-30
Feature request: Option for numbered or lettered counter


Gadgetgeek2000:
When writing documentation and I tab in for a process with a screenshot, my lists turn into a, b, c, lists instead of 1, 2, 3, lists. It would be helpful to have an option to choose between 1,2,3 or a,b,c, so my screenshot matches the documentation. 

I would envision just switch or selector in the toolbar when I have the counter tool selected.
![2018-11-30 15_22_51-setup visual studio - onenote](https://user-images.githubusercontent.com/5576809/49320014-60cae380-f4b4-11e8-8633-a628ea1234bc.png)

Thank for such a great tool!

PR #926, opened 2026-02-06
Add letter-mode and reset for step labels

danrhodes:
Added the ability to use letters (A-Z) instead of numbers for the step label counter tool. There's a new toggle button on the toolbar that shows "Abc" when you're in number mode and "123" when you're in letter mode - click it to switch between the two. Each label remembers whether it was placed as a number or a letter, so switching modes only affects new labels and doesn't change existing ones. Numbers and letters have independent counters, so if you place 1, 2, 3, then switch to letters, you get A, B, C, and switching back continues at 4. There's also a reset button (↺1) that starts a fresh sequence from 1 or A without affecting any labels already on the image.

This PR is off the back of some old issues - 
https://github.com/greenshot/greenshot/issues/103
https://github.com/greenshot/greenshot/issues/164
https://github.com/greenshot/greenshot/issues/723

image

image


Christian-Schulz:
In my NewFileFormat branch I struggled a lot with serializing StepLabels and the global counter start. 
Can you try to Copy&Paste your label groups between two editor windows, or save it as .greenshot file and reopen it?

Personally, I think it's going in the wrong direction if the `StepLabelContainer` are so tightly bound to the `Surface`. I think it needs to be decoupled. By that I also mean the current `Surface.CounterStart`. 
You could define a StepLabelService that encapsulates all functionality.

and just an other idea:
Instead of `bool IsLetterMode` you could be define a `enum StepLabelMode`

danrhodes:
Hey, thanks for the feedback! Kids are napping so here’s my thoughts before I forget 😄

You’re spot on with the StepLabelService idea - decoupling from Surface makes way more sense, especially for the serialization stuff you ran into on NewFileFormat. I’ll get that that changed.

Good shout on the enum too, StepLabelMode is cleaner than a bool and leaves the door open for roman numerals or whatever else down the line.

Haven’t tested copy/paste between editors or .greenshot save/reopen yet - I’ll make sure that works. 

Got double kids birthday parties this weekend so probably won’t get back to this until next week 🤣

danrhodes:
Is this your concern re copy/paste @Christian-Schulz the numbers don't keep between editor windows, and revert back to the next highest number in the editor being pasted into? I think this is a pre existing issue not introduced by this PR but needs addressing too.

image

danrhodes:
If so, this has been addressed in the PR I will push shortly.
image

danrhodes:
Yet the issue with saving is still there 🤔
image

danrhodes:
Ah thats better!

Two screenshots / copy+paste counters to the destination.

image

Save as Greenshot files - numbers are now preserved.

This is now solved because each label now remembers its own number when it's pasted from another editor. Before, the number was always calculated on the fly by counting how many labels come before it in the list - so pasting "8" into a surface with 5 labels would count 5 before it and show "6". Now, when a label is pasted from a different surface, it gets flagged to use its stored number directly instead of recalculating. That flag is saved with the file too, so it survives save/load. Same-surface copy/paste still renumbers dynamically like it always did.

image

danrhodes:
Refactored the whole step label system so StepLabelContainer no longer casts directly to Surface - everything now goes through a new StepLabelService that handles all the counting, registration and mode tracking. Swapped out the bool IsLetterMode for a proper StepLabelMode enum like Chris suggested. Also fixed the long-standing copy/paste bug where pasting a label from one editor to another would renumber it - label 8 now stays as 8 when pasted across editors. That fix works by flagging cross-surface pasted labels to use their stored number instead of recalculating dynamically, and that flag persists through save/load too. Same-surface paste still renumbers as you'd expect.

Christian-Schulz:
I meant something slightly different with the `StepLabelService`. 
The service should be globally unique. 
Register at start like 
   
    SimpleServiceProvider.Current.AddService(new StepLabelService());

and used like

    SimpleServiceProvider.Current.GetInstance<StepLabelService>();

In my opinion, the `Surface` should contain as little special code for one specific Conatainer as possible.  
There have to be a solution without these properties:
- Surface.StepLabelService
- Surface.CounterStart
- Surface.UseLetterCounter
- Surface.CounterGroup 

and without this function:
- Surface.CountStepLabels()

All functionality should be in the `StepLabelContainer` or in the `StepLabelService`.

The longer I think about it...
Why is there “CountStepLabels()”?  It defines the content to be drawn for each `StepLabelContainer` (and the icon in `EditorForm`). So the position in the container list was translated to the number. This was fine for numbers. Now we have letter.
The `StepLabelContainer`  already contains a property `Number` but this was only used for serialization.
Why not use it properly and use it directly when drawing? After that, sorting the list by number is no longer needed, this was more or less a workaround.
For letter mode there could be new property `TextLabel`. Or there could be a new property `Label` wich stores both the text and the number.

Yes, you have to reimplement every case in wich the Number could change e.g. removing a SetpLabel in the middle or Copy&Paste.
This should be handle in the `StepLabelService`.
In addition, it must be taken into account that it is possible that several EditorForms may be open, all uses the same `StepLabelService`.

What is your opinion on this?

danrhodes:
Evening,

Thanks for taking the time to think through this properly — really useful having your perspective from the NewFileFormat branch work.

Honestly, I hadn’t thought about it that deeply, I just wanted letters on my screenshots 😄 But you’re right, bolting it onto the existing Surface properties might be the wrong approach. Reading your breakdown it’s obvious the current design doesn’t scale too well— I was just adding more duct tape.

A global StepLabelService via SimpleServiceProvider feels like right call. No reason the Surface should know or care about step label counters, letter modes, or any of that. Let the service own it all.

If the container just store their own label value and the service handles assignment and renumbering, it’s cleaner across the board.

Serialisation, copy/paste, multiple editors all of it becomes simpler when the data lives where it should.

This feels a lot more complicated  so will take a bit longer than the original PR but it’ll actually be done right. I'll take a looo later and push the pr for your thoughts 💭

danrhodes:
I haven't forgotten about this, just grappling with making it work, who'd have thought it would be so difficult to swap from numbers to letters and back and have it all tracked....!

danrhodes:
- Added IStepLabelService interface (IStepLabelService.cs) - new global service contract with StepLabelMode enum (Number/Letter), counter start, mode, counter groups, renumbering, and label registration
- Created StepLabelService (StepLabelService.cs) - global singleton implementation registered via SimpleServiceProvider in EditorInitialize.cs. Manages all step label state, assigns values via Renumber() grouped by (mode, counter group)
- Decoupled step labels from ISurface/Surface - removed StepLabelService, CounterStart, UseLetterCounter, CounterGroup properties from ISurface. Surface no longer owns or manages step label state
- Refactored StepLabelContainer - each label stores its own mode (_isLetterMode), group (_counterGroup), and display value (_assignedValue). Uses HasFixedValue flag to preserve values during cross-surface paste and file save/load
- Updated ImageEditorForm - uses SimpleServiceProvider.Current.GetInstance<IStepLabelService>() directly instead of going through Surface
- Added toolbar controls (ImageEditorForm.Designer.cs) - "Abc"/"123" toggle button to switch between letter and number mode, "↺1" reset button to start a new counter sequence

Ive tested all the scenarios i can come up with and each came back fine.

@Christian-Schulz @Lakritzator - please let me know your thoughts on this 
image

Christian-Schulz:
First , it's amayzing 🤩

**Reset**
I'm not sure if the average user understands how to use the reset button.
But I don't have any useful ideas either. 

**ModeButton**
I would swap the text on the button. “123” when number mode is active. That way, you can see the current mode in the toolbar.

**CounterBox**
There are now multiple StepLabel groups, but the CounterBox in the Toolbar is for all of them. 
Is it poosible that every group has it own counter start. So the start value for the currently selected group is shown in the toolbar? 
Maybe the Service have to re-calculate the start value from the current group.  A global counter is then no longer needed.

I have no idea what should happen when multiple groups are selected.

****

Get an `ArgumentException`:
- open editor
- switch to LetterMode
- add "A"
- set counter start in Toolbar to 0

****

Add StepLabel in multiple Editor windows at the same time mixed all labels.

danrhodes:
Thanks, appreciate that, I've been chipping away at it all week it's not been easy!

I'll address your comments later and see what I can do 👍

danrhodes:
Hey @Christian-Schulz ,

Thanks again for the feedback, really glad you like it!

I've made changes for points 2, 3 and 4.

For the mode button, I've swapped the text so it now shows the current active mode - "123" when in number mode and "Abc" when in letter mode. Makes much more sense to see what you're currently using rather than what you'd switch to.

For the counter box, each mode and group combination now has its own independent start value. So if you set numbers to start at 5 and then switch to letters, letters will still default to 1 unless you change them separately. When you switch modes or reset the counter, the toolbar updates to show the start value for whatever group you're currently working with. Under the hood, the single counter start integer has been replaced with a dictionary keyed by mode and group.

For the counter start 0 issue, the letter conversion is 1-based, so 0 produced an empty string, which caused the problem. The service now clamps the minimum to 1 when in letter mode, and the toolbar spinner minimum updates dynamically - minimum 1 for letters, minimum 0 for numbers. So you simply can't get into that broken state anymore.

For the multi-editor mixing, this is a side effect of the global singleton approach :-(. Because all editors share one label list, labels from different editors interleave in creation order. So if you place 1, 2, 3 in one editor, then switch to another and add a label it becomes 4 instead of 1. In practice, most users work on one screenshot at a time, so it shouldn't come up often, but it's worth discussing whether it needs addressing.

image

There are a few possible approaches.

1 - The service could track which surface each label belongs to and renumber per surface, but that reintroduces coupling between the service and surfaces.

2 -  Each surface could get its own service instance, but that breaks the singleton pattern and makes the toolbar tricky when multiple editors are open.

3 - Or we could renumber based on spatial position within each surface, but that makes numbering unpredictable when labels get moved around - not good!

Each option adds quite a bit of complexity for what's likely an edge case. Happy to hear your thoughts on whether it's worth tackling or leaving it as a known limitation for now?

Christian-Schulz:
yea, you're right. The global singleton creates more problems than it solves.
Since the service contains all StepLabels, it would be better if it only store and handle those from the current editor. 
That would be your second approach. _(and your initial implementation 😉 )_ 

***
I compared the behavior of the toolbar in other containers a little more.
The toolbar displays all fields of the current container typewich is active in the left tool menu. 

If no container is selected and you change the values, these values are used for the next container you added.
If you have one container selected and you change the values, this changes only the selected container.
If you have multiple container selected, you see only these field wich are available in all contaner types combined. 

The behavior of StepLabels is currently slightly different, but it should actually be the same.
Changing the mode value should change the mode of the selected StepLabel. 
Klick reset button should reset the selected StepLabel. 
Changing the CounterStart should change the start value of the selected StepLabel.

The behavior of StepLabels would only be special because of the groups. 
If you change a special field in a StepLabels ( `LabelMode` or  `CounterStart`) all StepLabels in the same group should be changed also. ( Note: there is a `HandleFieldChangedEvent` in every container, maybe the special `OnPropertyChanged` in the Service no longer needed )

Currently Toolbar displays the global counter start but it should be the counter start of the selected StepLabel/StepabelGroup. 
So the ´_counterStart` in the service is not needed for the value in the toolbar, but still for the image in the tool menu.
It is not nessesary to store the CounterStart in the Service. It could return the CounterStart from one of the StepLabels in the last group. 

The reset button should be disabled if no StepLabel is selected. I think it should be disabled also if more then one Steplabel is selected.

If you reset the selected StepLabel, it should move this and all following StepLabel in this group to a new group and start at 1/A.

***
There is a bug in the current implementation. If you resize a StepLabel it changes the mode if the toolbar display a other mode. 
***
There is a field `_isLetterMode` and a property `LabelMode`. They store the same information. I think only `LabelMode` is needed.
***
would it be simpler if you store the StepLabes in the service lin a dictionary like this?
`Dictionary<(int group), List<StepLabelContainer>> _stepLabels`

danrhodes:
Wow, thank you @Christian-Schulz for the notes, I will go through them thoroughly tomorrow. 🙏

Lakritzator:
I love the discussion & energy of working together.
But as I see this is growing much bigger then it initially started, I just wanted "intervene" a bit.

Here is the vision I had when I created the Step-Label:

The idea was to provide a user a simple way of pointing towards steps a person had to take to reach a goal. For example if I would like to explain when in Greenshot the language can be changed. Generally this means that it's a unbroken range of numbers placed at "strategic" locations, so the person can write a list in an email or documentation... Click on 1, change at 2... or whatever.

This also means that the numbers automatically need to change, if one of the step-labels is removed, as it looks weird to have a gap. And it should be a no-brainer to add new ones, it just automatically increases.

That said, it doesn't mean it can't serve other/additional goals, but it's important to have an intuitive feature, easy to use.
I'm not sure the direction you are going, especially with keeping numbers when copying to other editors etc., is what helps a person for the use-case I designed it too. Also a certain complexity, of selecting if something is a 1-2-3 or a-b-c steplabel, might be unclear. Although the grouping seems nice, but which ones belong together, they all look alike!?

Before you continue, maybe can we focus on the use-case(s) that we want to serve?
Then look at what would be the best way to implement this, from a user journey and from a technical point of view.

Some guidelines that come to mind:

1. It has to be simple to use, if someone wants to do more with it then the standard, then it's okay to increase complexity.
2. It would be unacceptable if this doesn't work, due to conflicts, when multiple editors are open.

PR #929, opened 2026-02-08
Update setup.iss

xqtp:
Removing localization for this string to avoid duplicate app names, as the installer automatically appends 'Greenshot' to the description.

<details><summary>Screenshot</summary>

gst

</details>


Lakritzator:
That text doesn't look very Turkish though?

xqtp:
You're right "Uninstall" doesn't look very Turkish at all

I understand your concern, but there is a linguistic issue here. In Turkish, the object comes before the verb.

The installer automatically appends the {#ExeName} to the end of this string.

If I write 'Kaldır' (Uninstall), it becomes 'Kaldır Greenshot', which sounds like broken/unnatural Turkish (word-for-word translation from English).

If I write 'Greenshot uygulamasını kaldır', it becomes 'Greenshot uygulamasını kaldır Greenshot', which is a redundant duplication (as seen in my previous screenshot).

Since the software's name is forced to the end by the installer's logic, there is no way to create a grammatically correct Turkish sentence like 'Greenshot'ı kaldır'.

To avoid this 'broken Turkish' look or the 'double name' error, I decided to use the universal term `Uninstall`. This is very common in Turkish IT circles when a proper localization isn't possible due to hardcoded suffixing. That's why I reverted it to the original."

Lakritzator:
I actually can understand your reasoning, so merging this.

PR #932, opened 2026-02-08
Add new capture technology

Lakritzator:
This PR adds WindowsGraphicsCapture functionality to capture the Window (and screen, but this is not yet included) as a Beta Test feature.

WGC is a relative new feature in Windows, arrived with Windows 10, made to capture Windows and screens with the latest technologies. This should work with HDR, high DPI and also with games, but I didn't put it through extensive tests... Which is why we put it behind a feature toggle.

WGC captures the window contents from the composition engine, sometimes it takes a bit longer if there are no changes, so we need to monitor this. I also noticed that it uses transparency, let's see how that works.

Anyway, to make this work on .NET 10 would be easy, unfortunately we are on .NET framework, so I had to use some tricks and DirectX code, including SharpDx (no longer maintained) as a dependency would not be very satisfactory, so I had to create the needed interop classes myself.

For now I only enabled Window capture, due to limited time. The screen capture will need to be added later, as this is monitor based and not 100% in line with the current GDI (bitblt) capture.

### DIFF
--- a/src/Greenshot/Helpers/CaptureHelper.cs
+++ b/src/Greenshot/Helpers/CaptureHelper.cs
@@ -19,7 +19,6 @@
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
  */
 
-using log4net;
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
@@ -42,6 +41,8 @@ using Greenshot.Configuration;
 using Greenshot.Editor.Destinations;
 using Greenshot.Editor.Drawing;
 using Greenshot.Forms;
+using Greenshot.Native;
+using log4net;
 
 namespace Greenshot.Helpers
 {
@@ -908,6 +909,14 @@ namespace Greenshot.Helpers
             if (captureForWindow == null)
             {
                 captureForWindow = new Capture();
+            }
+
+            // New simplified logic with 1.4, using WindowsGraphicsCapture
+            if (CoreConfig.IsBetaTester)
+            {
+                captureForWindow.Image = WindowsGraphicsCaptureInterop.CaptureWindowToBitmap(windowToCapture.Handle);
+                captureForWindow.CaptureDetails.Title = windowToCapture.Text;
+                return captureForWindow;
             }
 
             NativeRect windowRectangle = windowToCapture.WindowRectangle;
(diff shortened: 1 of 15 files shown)


Lakritzator:
It's ✨🪄

Lakritzator:
Placed this behind a BetaTester feature toggle, which is currently only available via the greenshot.ini file.

Those wanting to help beta-testing this, just drop a file called greenshot-fixed.ini (if there isn't one already) with the following in the %APPDATA%\Greenshot directory and restart Greenshot:

```
[Core]
BetaTester=True
```

Christian-Schulz:
🤩

Lakritzator:
Can you take her for a ride? Wasn't easy to get her running...

Christian-Schulz:
I played around with it a bit.
1. the new version has transparent borders.😍 
image



2. Endless loop and/or wrong capture
I was stuck here:
https://github.com/greenshot/greenshot/blob/47191921cddeb15ed1c1c6b66fd8e5286cc1693c/src/Greenshot/Native/WindowsGraphicsCaptureInterop.cs#L350-L354
Steps:
- minimize all windows (used ShowDesktop function)
- used Greenshort SystemTray menu "capture window from list"
- pick a random window
- then nothing happens
- until i open the same window manually
- sometimes it captures the window correct
- sometimes it captures only parts of the window
- and somtimes i stuck in the loop for ever (i think this was one of the Greenshot editor window from the same debug session)

Lakritzator:
Thanks for having a look!

I also like the transparency, even though it said somewhere that it doesn't do that 😂

The loop _somewhat_ can be explained, Windows waits until something changes. Why nothing changes is a good question.

What do you mean with parts of the Window? More or less, Microsoft advertises this Technology to solve exactly THAT.

Chaython:
I set 
```
[Core]
BetaTester=True
```
and greenshot still doesn't seem to handle HDR correctly, preview/screenshot is very blown out still.
Is there something else I need to do?

Dmi2014:
just wanted to thank you, man. This thing was so fucking annoying.

Lakritzator:
@Chaython 
> I set
> 
> ```
> [Core]
> BetaTester=True
> ```
> 
> and greenshot still doesn't seem to handle HDR correctly, preview/screenshot is very blown out still. Is there something else I need to do?

It's currently only for the window capture, not the whole screen.

Lakritzator:
With V1.4.147 the new tech is also used with the region and screen captures. Still needs the BetaTester as it's not tested enough.

PR #933, opened 2026-02-09
Show beta tester info in about dialog

Christian-Schulz:
Now it is displayed in the about dialog if the “BetaTester” option is enabled in the INI file.

    [Core]
    BetaTester=True

image

### DIFF
--- a/src/Greenshot/Forms/AboutForm.Designer.cs
+++ b/src/Greenshot/Forms/AboutForm.Designer.cs
@@ -81,7 +81,7 @@ namespace Greenshot.Forms {
 			// lblLicense
 			// 
 			this.lblLicense.LanguageKey = "about_license";
-			this.lblLicense.Location = new System.Drawing.Point(109, 34);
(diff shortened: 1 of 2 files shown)


Lakritzator:
To be honest, I don't know the added value in there? Convince me 😁

What I could imagine, is to have the checkbox to enable/disable it in the expert settings, and the information in the "exception" details (so we understand if someone hit something due to the feature toggle)

Christian-Schulz:
No, this is more or less for developers like me. I sometimes have up to 6 different versions on my system for testing. One installed, some in dev-working-trees and some in portable ZIPs.
All could use different INI files.

I needed a way to see if the option is active in the running app.

I thought the about dialog next to the build version would be a good place, even if other people are testing features.

Lakritzator:
I'm not sure why the form has that many changes?

Christian-Schulz:
I had to move the text around a bit because the title got longer.
image

Issue #788, opened 2026-01-07
Big black spots in Greenshot window


Paredroi:
**Describe the bug**
When attempting to obscure parts of an image (or even before), large black spots appear or disappear in some areas.

**To Reproduce**
Example steps to reproduce the behavior:
1. Capture any (larger!?) part of the screen
2. Open with Paint.NET
3. [Modify &] Select all -> Ctrl+C to put in clipboard
4. Open image from clipboard with Greenshot
5. Black spots are visible

**Expected behavior**
no black spots!? :D

**Screenshots**
see GIF below

**Versions (please complete the following information):**
 - Greenshot version
latest 1.3.312
 - Windows version
latest Win10

**Additional context**

![Image](https://github.com/user-attachments/assets/6a6b8f9b-5f61-4ad8-a91d-8cc3fe55b17c)

jklingen:
Does it only happen when the editor zoom is at 66 or 75% zoom? Sounds like it could be the same issue as that one: https://github.com/greenshot/greenshot/issues/511

If yes, the issue is already fixed (but not included in a stable release yet). You might want to check out the latest (unstable) 1.4 release from our [version history](https://getgreenshot.org/version-history/). Or just wait until the fix is included in a stable release.

Paredroi:
Makes sense, it was at 75%, couldn't reproduce it at 100%.

But release **1.4.24-g7c2a1f4db5** seems not to have fixed it for me.

<img width="929" height="626" alt="Image" src="https://github.com/user-attachments/assets/e4fdd4f9-6320-496e-b026-f2695da72dc2" />

lynxus:
A simple fix would be to allow us to hardcode the editor to always be 100%

Annoyingly, it seems to always open at 75% and create the bug.

Lakritzator:
Still seeing it in the most recent builds, will look at it.

<img width="782" height="531" alt="Image" src="https://github.com/user-attachments/assets/04aef69a-2d14-4dbb-bbc3-476fab1ad390" />

Lakritzator:
These should be solved with the latest 1.4 builds which isn't released yet, but we have "unstable" versions which are for people who take the latest and live with risk ;)
I placed it on the changes for 1.3 if we release a bugfix.

Issue #511, opened 2023-08-21
Drawing on an image shows a black box


4nqyi:
**Describe the bug**
Any shape is drawn with a solid black box. The shape is only shown after a new shape is drawn.

**To Reproduce**
Example steps to reproduce the behavior:
1. Capture '....'
2. Open Editor
3. Use any of the tools on the left toolbar to draw a shape on the image
4. Draw another shape.

**Expected behavior**
The shape is drawn with a transparent background.

**Screenshots**
![2023-08-21 12_34_34-2023-06-30 12_48_40-Quick Start Guide - OM2200 png  -  FastStone Image Viewer 7](https://github.com/greenshot/greenshot/assets/11065729/9f0d4bc0-9070-4d0a-8f3d-4cdbd9b5a590)

**Versions (please complete the following information):**
 - Greenshot version - 1.3.277 64 bit
 - Windows version - Windows 10 Version 22H2 (OS Build 19045.3324)

ssuets110:
For the same reason I have tried reverting to an old version but the problem still exists.

**Greenshot version**
![image](https://github.com/greenshot/greenshot/assets/108708873/649f3c47-e1b0-40a7-8e25-55f923011a98)

**OS details**
![image](https://github.com/greenshot/greenshot/assets/108708873/3aec0d19-3642-4d68-b906-3480b0bf4dff)

**My experience**: One way to get rid of the black box is to hit the ESC key but it would reappear if the object (arrow, rectangle, eclipse, or text box) is moved or resized. It can be quite a hassle :(

wanyedude:
In my tests I found out that this behavior only seems to affect if the zoom mode is set to 66% or 75%. In all other zoom modes, i was not able to reproduce this behavior.



![Zoom](https://github.com/greenshot/greenshot/assets/143075815/1a9967c8-7033-48f7-bd1b-5126164bf79d)

ssuets110:
**Magnificent work on updating to 1.3.290, however**
This issue persists. hopefully to be fixed soon.

FF-Brown:
Looking into this one

raabma2:
The issue is still exists, also in 1.3.304

<img width="780" height="449" alt="Image" src="https://github.com/user-attachments/assets/2c9aaec9-1dee-44eb-9ea7-094dff0a52fa" />

PR #944, opened 2026-02-11
Fix black boxes / artifacts when using e.g. zoom level 75%

Lakritzator:
This change should fix the issues we had on some tickets. Still have to test it a bit more in detail...


Lakritzator:
This should fix #788 (and #511)

Issue #906, opened 2026-02-03
Feature Request: Add Optional Timestamp Watermark to Screenshots


julioalmeidamacedo:
Hello Greenshot Team,

I would like to suggest a new feature that I believe could be very useful for many users.

It would be great if Greenshot could offer, at the moment a screenshot is taken, the option to automatically add a watermark containing a timestamp. This watermark could include the date (year, month, and day), and optionally the time (hour, minute, and second).

Having this information embedded directly in the screenshot would be extremely helpful for documentation, auditing, troubleshooting, and evidence purposes, especially when screenshots are shared later and the original capture time is important.

Ideally, this could be configurable, allowing users to:

Enable or disable the timestamp watermark

Choose whether to include only the date or both date and time

Define basic formatting or positioning of the watermark

Thank you for your time and for developing such a great and useful tool. I appreciate your consideration of this suggestion.

Best regards,
Julio Macedo

Christian-Schulz:
Similar feature request to #468 and #349

ZloyIngrav:
Hi guys, I developed this Plugin to add CaptionBar. to the screenshots:

How can I share it with you to include it in the release?

<img width="1457" height="958" alt="Image" src="https://github.com/user-attachments/assets/cda913ac-8d86-4a6b-ab07-1b5c74f89805" />

Christian-Schulz:
@ZloyIngrav That looks interesting. 🤩
Do you need any help collaborating on the project?

These are the usual steps:
- Create your own fork of the greenshot repository ( https://github.com/greenshot/greenshot/fork )
- Check out the main branch
- Create a new local branch for this feature
- Do your changes in the code _(add your new plugin project to the solution )_ and commit
- Push the feature branch to your GitHub
- Create a new pull request here https://github.com/greenshot/greenshot/pulls

ZloyIngrav:
@Christian-Schulz Thank you, I will push it soon. Just found one more bug with text positioning. Аfter some tests will push Plugin like you explained.

danrhodes:
Keeping an eye on this and waiting for the PR, looks good! 👍 

👀

ZloyIngrav:
https://github.com/greenshot/greenshot/pull/953

danrhodes:
I'll check this out tonight 👀

ZloyIngrav:
> I'll check this out tonight 👀

It was my first PR in Open Source projects. How does it work? When can we expect the next build with this Plugin?

[...]

PR #953, opened 2026-02-13
Add CaptionBar plugin - automatic caption bar for screenshots

ZloyIngrav:
https://github.com/greenshot/greenshot/issues/906#issuecomment-3853076207

### DIFF
--- a/.github/agents/release-documentation-specialist.md
+++ b/.github/agents/release-documentation-specialist.md
@@ -1,0 +1,28 @@
+---
+name: release-documentation-specialist
+description: aevent specialising in maintaining changeligs and release documentation
+---
+
+As a release documentation specialist, you are responsible for:
+
+* generating changelogs for all stable releases
+* create drafts for blog posts about new releases
+
+user change logs
+
+* should be user friendly, leave out purely technical changes, focus on provided value for user, natural (non-technical)  language
+* most important changes first
+* link to detailed (technical) change log at the end
+
+technical change logs
+
+* contains technical changes and
+* contains changes from user change logs but may use more technical language
+
+blog posts
+
+* should have similar content as user change log
+* use full sentences
+* highlight mist important changes.
+* link to user change log at the end
+
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -10,6 +10,11 @@
 - **Build System**: MSBuild (requires Visual Studio or MSBuild Tools for Windows)
 - **Versioning**: Nerdbank.GitVersioning (version base: 1.4.x)
 
+## Specialised Agents
+
+-translation-manager: always consult them about UI messages, esp. after addition/change/removal of UI messages
+- release-documentation-specialist: always ask them to update release notes after applying changes to the code base or when there is demand to update release notes or write a release blog post
+
 ## Build Requirements & Environment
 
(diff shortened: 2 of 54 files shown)


danrhodes:
@ZloyIngrav Thanks for the contribution — the core concept of the CaptionBar plugin is really interesting and the main processing logic in CaptionBarProcessor.cs looks substantial. However, there are two blocking issues found during testing that need to be resolved before I can test further.

1. Project not added to solution file
Greenshot.Plugin.CaptionBar.csproj exists on disk but was never added to Greenshot.sln. As a result, the plugin is completely absent from any build when working within the solution — it will not compile, will not be copied to the output directory, and will not be discoverable at runtime. Anyone cloning the repo and building via Visual Studio or dotnet build on the solution will not get this plugin at all.

The project entry, build configuration mappings, and solution-level dependency on the main Greenshot project all need to be added to Greenshot.sln.

2. Settings dialogue crashes Greenshot on open
Clicking "Configure" for the CaptionBar plugin in the Preferences window causes Greenshot to hard-crash (exit code 0xffffffff) with the error **_enumType** was null.

Root cause: The two GreenshotComboBox controls (comboBoxTimestampAlignment and comboBoxCustomTextAlignment) in the Designer are configured with SectionName = "CaptionBar" and bound PropertyName values, which means the GreenshotForm base class expects to manage them entirely via its INI binding system — calling Populate(enumType) to set _enumType before any items are touched. However, the constructor in CaptionBarSettingsForm.cs also manually calls Items.AddRange(...) on both controls before GreenshotForm has had a chance to call Populate(). Adding items fires SelectedIndexChanged → StoreSelectedEnum() → _enumType.Name → null reference crash.

The fix is to pick one approach and be consistent. Either:

Remove the manual Items.AddRange calls and let GreenshotForm populate the combo boxes from the enum type via its binding system (the intended pattern for GreenshotComboBox), or
Replace both GreenshotComboBox controls with plain ComboBox controls, remove the SectionName/PropertyName designer properties, and manage them fully manually alongside the existing NumericUpDown handling.

The mixed approach — GreenshotComboBox with designer binding and manual item population — is not viable.

Both issues were reproduced on Greenshot 1.3.312, Windows 11, .NET 4.8, debug build from this branch.

Lakritzator:
I just wanted to say that I love contributions, and this looks like a nice one.

Before you stick a lot of work in this, let me explain how I generally work:

1. If we accept contributions, we need to be able to maintain this in the future, as the person providing it might no longer be available. So it needs to be something we understand. I think I can maintain it, so "check".
2. Features need to make little to no extra work in our issues, we already get a lot of doubles / issues which are already solved, these take time. Then we have issues where people have problems with Windows, and all the other things. So I like low maintenance on the issue side. I don't know about that one yet, as I didn't use it.
3. We need translations for all the UI elements, this is now MUCH easier with Copilot, BUT the sizes of the texts change, and might add a lot of work testing and aligning.
4. We need a simple usage, default it should work without thinking, okay to make it complexes for users who invest time.

I am currently working on something complex, as it cuts through some legacy stuff in Greenshot, so I prefer to finish that.
But you can already look at the suggestions @danrhodes made, he knows his stuff...
There is still no guarantee we integrate this, I am not against the idea, but I need to use it to get a feeling for it.

Issue #952, opened 2026-02-12
Greenshot making constant CPU load?


ryukisai:
Hi! i recently noticed that Greenshot is constantly and consistently using ~10% of my CPU in task manager.
I’m using GS 1.3.312 on win11 pro 25H2. My cpu is not high end (i5-8400H @2.5 GHz and 4 cores).

I also use ESET endpoint security (antivirus enforced by my employer) if it helps…

Not sure this is a bug, but is anyone elso noticing this?
I already tried reinstalling, no luck. Thanks in advance!

<img width="789" height="538" alt="Image" src="https://github.com/user-attachments/assets/466e1e6d-4609-4ccd-9b4f-1eceb551b4d7" />

Christian-Schulz:
We were unable to reproduce this. see #924

ryukisai:
Hi, thanks for the quick reply. sorry i didn't notice the former ticket. I'll try with the beta and report asap.
Edit: just for full disclosure, i only have this issue on 1 PC out of 3 for now… All are on 25H2 but they all have different hardware.

Lakritzator:
If you have the about open, there is an animation, so that causes some CPU.
The question is what happens if you have all windows closed?
It shouldn't really do something, so I am also interested in what happens!

Unless you have the JIRA Plugin (JIRA is an issue tracker application) active and some roque application is changing window titles like crazy, I did actually see that once, then it might do something in the background. It will process the windows title, in the hope you open a JIRA ticket and Greenshot will recognize that from the title...

One other option is possible, that it somehow can't reach out to download the list of Greenshot versions, to detect if there is a newer version. Or if it can't write the configuration... Do you see anything in the [logfile](https://getgreenshot.org/faq/where-can-i-find-greenshots-log-file/)?

Fixer-007:
I had the same problem just now with greenshot 1.3.312 5a60aff83e on windows 10. 10+% cpu usage on 6-core cpu, had to exit the program, it was in background. I've restarted it and I still have constant high cpu usage.

Lakritzator:
I cannot reproduce this, so it's important to understand what your usage is like.
One other important question, how did you install Greenshot? Did you download the installer from OUR website? (https://getgreenshot.org)

Fixer-007:
[...]

Issue #822, opened 2026-01-15
Crashing


gooftrpr-ops:
**Describe the bug**
Crashes when I try to use. below is the report.

Software version: 1.3.300 - c1ad1d4a93 (64 bit)
.NET runtime version: 4.0.30319.42000+
Time: 2026-01-15 07:07:24 -06:00
OS: Windows 10 2009 x64 build 22631
GDI object count: 122
User object count: 85

Exception: System.ArgumentOutOfRangeException
Message: The added or subtracted value results in an un-representable DateTime.
Parameter name: value

Stack:
   at System.DateTime.AddTicks(Int64 value)
   at Greenshot.Base.Core.FileDescriptor..ctor(BinaryReader reader)
   at Greenshot.Base.Core.FileDescriptorReader.<Read>d__0.MoveNext()
   at Greenshot.Base.Core.ClipboardHelper.<IterateFileDescriptors>d__24.MoveNext()
   at Greenshot.Base.Core.ClipboardHelper.<IterateClipboardContent>d__22.MoveNext()
   at Greenshot.Base.Core.ClipboardHelper.ContainsImage(IDataObject dataObject)
   at Greenshot.Editor.Forms.ImageEditorForm.UpdateClipboardSurfaceDependencies()
   at Greenshot.Editor.Forms.ImageEditorForm.ImageEditorFormActivated(Object sender, EventArgs e)
   at System.Windows.Forms.Form.OnActivated(EventArgs e)
   at System.Windows.Forms.Form.set_Active(Boolean value)
   at System.Windows.Forms.Form.WmActivate(Message& m)
   at System.Windows.Forms.NativeWindow.Callback(IntPtr hWnd, Int32 msg, IntPtr wparam, IntPtr lparam)

Configuration dump:

Lakritzator:
[...]

PR #956, opened 2026-02-14
Stability fixes for 1.3

Lakritzator:
This PR fixes for 1.3 the following which was already fixed in 1.4:

- #822 - Crash due to unexpected information on the clipboard
- #952 - CPU load due to background task scheduler bug
- #788 & #944 - Weird black areas in the editor when using 75% zoom

Additionally:

- Greenshot doesn't Windows 11 correctly and says Windows 10 in the error window
- Starting to draw an line/arrow starts with a wrong destination, this means it is briefly showing a line/arrow with an endpoint which wasn't picked yet.
- A glitch in the region capture, where the zoom/cross starts on the wrong position.

### DIFF
--- a/src/Greenshot.Base/Core/Capture.cs
+++ b/src/Greenshot.Base/Core/Capture.cs
@@ -113,7 +113,7 @@ namespace Greenshot.Base.Core
             set
             {
                 _cursor?.Dispose();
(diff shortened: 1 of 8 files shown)


jklingen:
Great, thanks a lot 👍

PR #958, opened 2026-02-14
[Fix] Cancel destination picker menu with ESC

Christian-Schulz:
This fixes an exception when the destination picker is canceled with ESC.


Lakritzator:
I hate fixes like that, feels so bloated. That's not against you, just complaining on how hard UI code can be...

Issue #472, opened 2023-02-21
Greenshot stuck in capture screenshot mode


eblis:
**Describe the bug**
Sometimes when I press PrintScreen hotkey to take a screenshot it takes a looong time for the Greenshot window to appear, and when that happens Greenshot freezes in capture mode, I cannot take the screenshot even if I select a region.
I can still press alt+tab and switch to other windows, but I don't see their content as the capture window from Greenshot is on top.
The only way to get rid of this issue is to restart the machine or, if you know what you're doing open TaskManager, navigate to Greenshot and terminate the process, all without seeing what you're doing as the Greenshot window is drawn over everything else.

**To Reproduce**
Unfortunately I don't know how to reproduce this issue, it just happens from time to time. It happened to me twice in 2 consecutive days.
I just press the PrintScreen hotkey and if it takes a loong time for the capture window to appear I know that something is wrong.
Is there a debug log that I can enable ?

**Expected behavior**
Capture screenshot should finish once I select aregion.

**Versions (please complete the following information):**
 - Greenshot version: 1.3.274 - 73005f741a (x64)
 - Windows version: Windows 10 x64 22H2 19045.2546.

Lakritzator:
There is already a log file: https://getgreenshot.org/faq/where-can-i-find-greenshots-log-file/
But it doesn't show the debug level by default: https://getgreenshot.org/faq/how-can-i-make-greenshot-write-debug-information-to-its-log-file/

Strange, I saw this issue in older versions of Greenshot and was expecting that it has to do with the older technology that was using. Maybe it's a different issue, so we need to have a look. It would be good to know what the last thing it was doing before hanging.

Please consider what other applications you are using, some are known to cause issues with Greenshot. (e.g. Slack as described here #320, or maybe power tools)

PerplexedPeach:
I'm also getting this issue. Specifically it occurs often when I'm playing BG3 in fullscreen (Vulkan).

rafaelrwrc:
[...]

PR #959, opened 2026-02-14
Fix CPU spikes, memory leaks, and resource management issues

danrhodes:
Evening all,

Lately, there have been a good few issues popping up around CPU and RAM usage, so I thought I'd get ahead of the game and do a thorough audit of the codebase for potential resource-related problems. Here's what was found and fixed:

**CPU fixes:**

- Fixed a busy-wait spin loop in WindowDetails.Restore() that would pin a CPU core at 100% while waiting for a window to un-minimise — now bounded with a sleep interval and timeout
- Added a safety limit to the capture mode switching loop in CaptureHelper to prevent infinite cycling between GDI/Aero/Screen capture modes

**Memory leak fixes:**

- Fixed a Font leak in AbstractDestination where a new GDI Font was allocated every time the context menu opened without disposing the previous one
- Fixed a Bitmap leak in CredentialsHelper where an intermediate Bitmap was never disposed after extracting its GDI handle
- Fixed event handler leaks in 4 plugins (Box, Dropbox, Flickr, ExternalCommand) that subscribed to Language.LanguageChanged and CoreConfig.PropertyChanged but never unsubscribed in their Shutdown() methods, preventing garbage collection
- Fixed the Cache<TK,TV> class where internal System.Timers.Timer instances were stopped but never disposed — CachedItem now implements IDisposable and timers are properly cleaned up on removal and replacement
- Fixed repeated lambda event handler subscriptions in EmojiPicker that would accumulate on every popup load — converted to named methods with a guard flag
- Added proper timer disposal in AnimatingForm (was only stopped, not disposed) and MainForm (double-click timer never disposed)

**Thread and async fixes:**
- Added .ConfigureAwait(false) to all async calls inside Task.Run lambdas in Confluence.cs and the OCR processors to reduce deadlock risk
- Added a timeout to the ConfluenceUpload space-loading poll loop to prevent indefinite blocking
- Added a 60-second timeout to MailHelper.WaitOne() to prevent hanging if the MAPI mail thread throws an exception
- Added debouncing (300ms) and cancellation to JiraForm.JiraKeyTextChanged which was firing an API call on every single keystroke with no cancellation of previous in-flight requests

**GC pressure reduction:**

- Removed an unconditional GC.Collect() + GC.WaitForPendingFinalizers() in EditorDestination that was causing unnecessary full GC pauses on every screenshot export
- Moved GC.Collect() in ImageEditorForm behind the MinimizeWorkingSetSize user opt-in so it only runs when explicitly requested

### DIFF
--- a/src/Greenshot.Plugin.Jira/Forms/JiraForm.cs
+++ b/src/Greenshot.Plugin.Jira/Forms/JiraForm.cs
@@ -23,6 +23,7 @@ using System;
 using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
+using System.Reactive.Linq;
 using System.Threading.Tasks;
 using System.Windows.Forms;
 using Dapplo.Jira.Entities;
@@ -40,7 +41,7 @@ public partial class JiraForm : Form
     private readonly JiraConnector _jiraConnector;
     private IssueV2 _selectedIssue;
     private readonly GreenshotColumnSorter _columnSorter;
-    private System.Threading.CancellationTokenSource _jiraKeyCts;
+    private IDisposable _jiraKeySubscription;
 
     public JiraForm(JiraConnector jiraConnector)
     {
@@ -60,6 +61,13 @@ public partial class JiraForm : Form
 
         uploadButton.Enabled = false;
         Load += OnLoad;
+        FormClosed += (_, __) => _jiraKeySubscription?.Dispose();
+
+        _jiraKeySubscription = Observable
+            .FromEventPattern(jiraKey, nameof(jiraKey.TextChanged))
+            .Throttle(TimeSpan.FromMilliseconds(300))
+            .ObserveOn(this)
+            .Subscribe(async _ => await JiraKeyTextChanged());
     }
 
     private async void OnLoad(object sender, EventArgs eventArgs)
@@ -270,7 +278,7 @@ public partial class JiraForm : Form
         jiraListView.Sort();
     }
 
-    private async void JiraKeyTextChanged(object sender, EventArgs e)
+    private async Task JiraKeyTextChanged()
     {
         string jiranumber = jiraKey.Text;
         uploadButton.Enabled = false;
@@ -278,26 +286,13 @@ public partial class JiraForm : Form
         int dashIndex = jiranumber.IndexOf('-');
         if (dashIndex > 0 && jiranumber.Length > dashIndex + 1)
         {
-            // Cancel any previous in-flight request to debounce rapid typing
-            _jiraKeyCts?.Cancel();
-            _jiraKeyCts = new System.Threading.CancellationTokenSource();
-            var token = _jiraKeyCts.Token;
-
             try
             {
-                // Wait briefly to debounce keystrokes
-                await Task.Delay(300, token);
-                if (token.IsCancellationRequested) return;
-
                 _selectedIssue = await _jiraConnector.GetIssueAsync(jiraKey.Text);
-                if (_selectedIssue != null && !token.IsCancellationRequested)
+                if (_selectedIssue != null)
                 {
                     uploadButton.Enabled = true;
                 }
-            }
-            catch (OperationCanceledException)
-            {
-                // Expected when a new keystroke cancels the previous request
             }
             catch (Exception ex)
             {
--- a/src/Greenshot.Plugin.Jira/Forms/JiraForm.Designer.cs
+++ b/src/Greenshot.Plugin.Jira/Forms/JiraForm.Designer.cs
@@ -177,8 +177,7 @@ namespace Greenshot.Plugin.Jira.Forms;
 			this.jiraKey.Name = "jiraKey";
 			this.jiraKey.Size = new System.Drawing.Size(158, 20);
 			this.jiraKey.TabIndex = 1;
-			this.jiraKey.TextChanged += new System.EventHandler(this.JiraKeyTextChanged);
-			// 
+			//
 			// JiraForm
 			// 
 			this.AutoScaleDimensions = new System.Drawing.SizeF(96F, 96F);


danrhodes:
Hmmm, why the issues merging for plugins?

Lakritzator:
I don't know when you started this, but I didn't have a clue on why the conflict. 

That's one very impressive list of stability changes 🥰

Lakritzator:
What about the end result, I mean yes these changes mostly make sense, but in the end are they measurable?

danrhodes:
Fair question — and honestly, most of these won’t produce a dramatic before/after benchmark because they’re fixing edge cases and slow-burn issues rather than one big bottleneck. But some are definitely measurable.

The ones you’d see on a graph:

- The WindowDetails.Restore() busy-wait would genuinely pin a core at 100% for the duration of the wait — that’s trivially measurable and immediately noticeable on any system with limited cores

- The GC.Collect() removal in EditorDestination — forced full GC pauses on every export are measurable in UI latency, especially on lower-spec machines

- The Jira keystroke API spam — easy to see in network traces, and users on slow connections would feel it

The ones that matter over time but are harder to graph:

- The memory leaks (Font, Bitmap, event handlers, timers) — these won’t show up in a 5-minute test, but they compound over long sessions. Anyone who leaves Greenshot running all day (which is most users) would eventually see working set creep. The event handler leaks in particular prevent entire plugin object graphs from being collected

- The timer disposal fixes are more about correctness — leaked timers hold thread pool resources and prevent clean shutdown

The ones that are defensive:

- The ConfigureAwait(false) changes and timeout additions are deadlock/hang prevention — you can’t measure “didn’t hang” easily, but the absence of the fix is measurable when it happens (frozen UI, stuck threads in dumps)

So the short answer: the busy-wait fix and the GC.Collect removal are immediately measurable. The memory leaks are measurable over longer sessions with a simple working set monitor. The rest are correctness and resilience fixes that prevent hard-to-diagnose issues down the line rather than moving a benchmark needle.

If it’d help, I could put together a simple before/after test — run a loop of captures over 30 mins and compare working set + CPU profile between the two branches.

Happy to do that if you want hard numbers, but wanted to get the fixes up first while they were fresh.

In terms of the merge issues, I pulled all the fixes from code I had stashed directly into a fresh branch of the greenshot/main that was up to date just prior to publishing which is my general practice so I know I'm working on the most latest code. 🤷‍♂️

danrhodes:
The merge conflicts are likely due to changes on gs/main and the same base class or shared patterns across the plugin files since I created the branch. I'll resolve these tomorrow.

danrhodes:
What a nightmare that was to get rid of the merge issues. It turns out I forked from my main, which has changes that are not yet merged, and had to do a full rebase to get it clean. 👎🏻

danrhodes:
I have ran out of time to run two 30 minutes tests so will complete these tomorrow and post the results.

Lakritzator:
No stress, I will take it when it's ready...

danrhodes:
I have written a powershell script that will monitor the running exe and pull into a csv all the relevant performance statistics, running for 4 hours on the pre and 4 hours on the post so I can then graph these and determine the changes impact. Quite easy to test as I use GS all day long I'll just swap out my latest pr at dinner 👌

Lakritzator:
I can see that the changes will have some effect, so it's worth it anyway, I was just curious how much it's measurable. I didn't expect you to put hours in that.

danrhodes:
I'm curious to see myself if only for an exercise in my own head.

Don't worry, I love this kind of stuff. I'm used to having to provide proof of concept through work so it's no big deal and may help towards other issues/pr'a maybe.

danrhodes:
I haven't had a moment spare to run the benchmark test, I will look to do them tomorrow 👍

src/Greenshot.Base/Core/CredentialsHelper.cs
```diff
@@ -371,7 +371,8 @@ private CredUi.INFO GetInfo(IWin32Window owner)
             info.pszMessageText = Message;
             if (Banner != null)
             {
-                info.hbmBanner = new Bitmap(Banner, ValidBannerWidth, ValidBannerHeight).GetHbitmap();
+                using var bannerBitmap = new Bitmap(Banner, ValidBannerWidth, ValidBannerHeight);
+                info.hbmBanner = bannerBitmap.GetHbitmap();
```
Lakritzator, line 381: Makes me wonder if the HBitmap is ever disposed...

src/Greenshot.Plugin.Jira/Forms/JiraForm.cs
Lakritzator, line 1: Doesn't it make sense to use reactive extensions here for the debounce?

Lakritzator:
All good changes. Just making my skin crawl with the .ConfigureAwait(false) as this looks like it's on the correct places, but sometimes... 🤷‍♂️
Greenshot wasn't started with async await in mind, applying this to it causes some very weird hoops we need to jump through.

danrhodes:
I have the results of both pre and post tests so will collate and post later.

src/Greenshot.Base/Core/CredentialsHelper.cs
danrhodes: Good catch -  the HBITMAP from GetHbitmap() was leaking on any exception path; wrapped the cleanup in a try/finally to guarantee it's always freed. Will be in the next push.

src/Greenshot.Plugin.Jira/Forms/JiraForm.cs
danrhodes: Rx would be cleaner but adds a whole dependency just for one debounce — the Task.Delay + CTS approach does the same job without the overhead.

Lakritzator: We already use it, go for it.

Lakritzator: Jira Plugin, dapplo.windows

danrhodes:
Fanfare please, drumroll too ----

I ran a 60-minute perf capture before and after the changes.

The main targets were GDI and USER object usage — GDI dropped ~46% and USER objects ~56% in steady state, so the fixes have made a solid dent.

GC collections also fell off a cliff (Gen0 down 75%, Gen1 -87%, Gen2 -92%), which tracks with less object churn from the areas I've cleaned up.

Page faults were sitting around 4,200/s toward the end of the pre-fix capture and basically flatlined to zero post-fix once everything loaded in.

A couple of trade-offs to flag.

Working Set and Private Bytes crept up slightly (+12% and +7%), and the Large Object Heap went from ~8 MB to ~26.5 MB. 

This is expected — the changes consolidate a lot of small, frequently-created GDI/USER handles into fewer, larger managed allocations, so the memory footprint shifts from unmanaged handles into the managed heap. GC Time % ticked up as a result, but that's a fair trade for reducing handle consumption.

Net result: meaningful reduction in resource usage across the areas I've been able to address so far. There may still be more to find, but the captures show a clear improvement.

Pre and post data attached.

image
image
image
image
image

Apologies for going all work mode here........

[Greenshot-Perf-post-fix-20260218-090805.csv](https://github.com/user-attachments/files/25398908/Greenshot-Perf-post-fix-20260218-090805.csv)
[Greenshot-Perf-pre-fix-20260217-101711.csv](https://github.com/user-attachments/files/25398909/Greenshot-Perf-pre-fix-20260217-101711.csv)

src/Greenshot.Plugin.Jira/Forms/JiraForm.cs
danrhodes: Switched to Rx Throttle — System.Reactive was already a transitive dep via Dapplo.Windows, so no new dependency needed. CTS bookkeeping is gone entirely. Mind blown! 🙌

danrhodes:
FriendsFriendsTvGifGIF

I have to say im pretty happy with this PR as a whole 👍

Lakritzator:
I'm a bit surprised about the page faults, although I wouldn't be worried about the old behavior, I didn't expect such an improvement. Probably not something people will notice though. 

I'm fine with the trade of, we can always pass through the code to look for reductions, but we can't let those pesky bugs run loose.

Lakritzator (approved):

Lakritzator:
I do wonder if letting Greenshot run for a long time, we see things like in #472 ?

danrhodes:
> I do wonder if letting Greenshot run for a long time, we see things like in #472 ?

Quite possibly if left running long enough. Especially as Shutdown these days isn't a true shutdown in Windows.

danrhodes:
> I'm a bit surprised about the page faults, although I wouldn't be worried about the old behavior, I didn't expect such an improvement. Probably not something people will notice though. 
> 
> 
> 
> I'm fine with the trade of, we can always pass through the code to look for reductions, but we can't let those pesky bugs run loose.

Im happy to iterate over to look for any reductions. 🙏

PR #962, opened 2026-02-15
Fix the icon resize issues

Lakritzator:
This PR was created to fix the icon resize issues, after we were getting a lot of issues for this, we need to solve it.

I hope to have all issues fixed, but it does come at a bit of a trade-off, the application will use more memory.

### DIFF
--- a/src/Greenshot.Base/Core/MenuHelper.cs
+++ b/src/Greenshot.Base/Core/MenuHelper.cs
@@ -76,7 +76,7 @@ public static class MenuHelper
             return;
         }
         var iconClone = needsClone ? ImageHelper.Clone(icon) : icon;
-        if (iconClone != null)
+        if (iconClone == null)
         {
             toolstripItem.Image = null;
             return;
@@ -84,7 +84,7 @@ public static class MenuHelper
 
         toolstripItem.Image = iconClone;
         // Dispose the cloned icon when the menu item is disposed to prevent memory leaks
-        toolstripItem.Disposed += (sender, e) => iconClone.Dispose();
+        toolstripItem.Disposed += (sender, e) => iconClone?.Dispose();
     }
 
 }


Lakritzator:
@danrhodes Can you review this?

I still have some doubts about the PluginUtils.GetCachedExeIcon, probably it makes sense if this always returns a cloned version, so the one getting it needs to figure out how to dispose it.

danrhodes:
Ah sorry, I have logged off now, I've only just seen this, I will review tomorrow for you.

Lakritzator:
No worries, it can wait... I was just annoyed with the bugs coming in, so I had to move away from my other challenging topics (cursor, new capture and some other.... surprises) in Greenshot to do work on this.

danrhodes:
Hey @Lakritzator , had a look through this. The approach is solid — the root cause is clear, when IconSize changes the cache gets cleared and all the cached images get disposed, so any UI element still holding a reference ends up with a disposed Image. Cloning before assigning to UI elements and disposing the clone when the element is disposed is the right fix.

To answer your question about GetCachedExeIcon — yes, I think it makes sense for it to always return a clone. Right now it returns a shared reference to a cache entry that can be disposed at any time, which means every caller has to remember to clone it. If GetCachedExeIcon returned a clone directly, the ownership semantics become clear (caller owns it, caller disposes it), all the scattered ImageHelper.Clone calls at each call site go away, and future callers can't forget to clone. The memory trade-off is already accepted by this PR anyway; this would just centralise it. The cache still holds the master copy, callers get independent clones, and OnIconSizeChanged continues to dispose the masters on resize. If you do that, the explicit clone in WindowDetails.DisplayIcon becomes redundant and can be removed.

One bug I spotted — in MainForm.cs around line 1178, the Disposed handler is attached unconditionally but DisplayIcon can return null (it has explicit return null paths), so windowIcon.Dispose() would throw a NullReferenceException during disposal. Needs an if (windowIcon != null) guard to match the pattern used everywhere else in the PR.

Also worth noting the WindowDetails.DisplayIcon change subtly alters the ownership contract of that property. Previously, the first code path (appIcon.ToBitmap()) returned a new bitmap but the second path returned the shared cached reference. Now both paths return caller-owned images, which is good for consistency, but it means all callers of DisplayIcon must now dispose the result. The MainForm.cs change handles this (once the null bug is fixed), but worth double-checking there aren't other callers that don't dispose.

The old image disposal in ExternalCommandPlugin.cs around line 172 is well done — without that you'd leak clones on every icon size change. Only minor point there is confirming OnIconSizeChanged always fires on the UI thread to avoid a race between disposing the old image and the UI rendering it. Probably fine since PropertyChanged from settings usually marshals to the UI thread, but worth being sure.

One last thought — the clone + dispose pattern appears 6+ times across the PR. You could extract a small helper like SetClonedIcon(ToolStripItem item, Image sourceIcon) that does the clone, assigns it, and hooks up the Disposed handler. Each call site becomes a one-liner and you eliminate the risk of someone forgetting the Disposed handler in future. Up to you, though, not essential.

Looks good overall, just the null check fix needed.

Lakritzator:
@danrhodes I was working on the cursor issues in #966, will look at your comment "next".

Lakritzator:
Okay, if I have access and energy to sit behind my PC, I'll fix that and merge. Thanks for the very thorough assessment.

Lakritzator:
I'll merge it like then, and we handle GetCachedExeIcon in another PR.

danrhodes:
I have an edge case in native windows on the latest release my green cursor is HUGE! I'll send a screenshot over plus my settings.

Issue #898, opened 2026-01-31
Scaled mouse cursors appear too large and pixelated in Photoshop and Illustrator


meerderideen:
Hi,

I really like that the scaled mouse cursors from Windows 10 are now also captured well (I am currently using **Greenshot 1.4.68**). Unfortunately, however, this does not work in programs such as **Adobe Photoshop** or **Adobe Illustrator** (just two examples). In these programs, the system-wide settings in Windows do not affect the mouse cursors anyway, so they are displayed small (which is normal).

When taking a screenshot with Greenshot, however, these mouse cursors are suddenly displayed many times larger and therefore appear **extremely pixelated** and with a strange border:

<img width="168" height="163" alt="Image" src="https://github.com/user-attachments/assets/74796954-f3d5-4e8a-b7f2-b73fbd5ed225" />
<img width="224" height="255" alt="Image" src="https://github.com/user-attachments/assets/1c80cd0a-5e44-4cf0-a573-3103e61eb696" />
<img width="146" height="143" alt="Image" src="https://github.com/user-attachments/assets/094ebbfe-40df-492a-a8bf-3b8616e698d9" />
<img width="224" height="186" alt="Image" src="https://github.com/user-attachments/assets/c80f347e-9634-4116-949a-77391b94c68d" />
<img width="238" height="221" alt="Image" src="https://github.com/user-attachments/assets/4009dc59-1aa5-4b45-a998-f95b879d236f" />
<img width="173" height="148" alt="Image" src="https://github.com/user-attachments/assets/4b2c27e8-484f-4f15-93a2-da46e994ba09" />

Ideally, in such cases, the cursor should be captured "as is."

Windows 10
Greenshot 1.4.68

Lakritzator:
Can you name me some apps which also have this behavior, something that doesn't cost?

meerderideen:
Hello, yes... Affinity Photo or Affinity Designer for example (it´s for free: https://www.affinity.studio/de_de/photo-editing-software):

<img width="198" height="136" alt="Image" src="https://github.com/user-attachments/assets/d7eff37c-c617-4c7a-aa70-3632631eb37e" />
<img width="176" height="160" alt="Image" src="https://github.com/user-attachments/assets/d6fe4069-3d69-4a9d-b2d1-8ebb693914f2" />
<img width="162" height="127" alt="Image" src="https://github.com/user-attachments/assets/bf8aef3a-3e67-489e-baf5-f1d4f3ba85ea" />
[...]

PR #966, opened 2026-02-16
Improved cursor capturing

Lakritzator:
Changed the way the cursor is captured, this should solve most cursor issues and make things possible which weren't before.

Most of the complex logic I wrote in DapploWindows, see [this](https://github.com/dapplo/Dapplo.Windows/commit/a7484c61e8614945038c6221f5e543c500a99d04) commit

Tested with:

- Normal cursor, in different shapes and sizes
- Black cursor, in different shapes and sizes
- Inverted cursor in different shapes and sizes
- Colored cursor in different shapes and sizes
- Some custom cursors, like Paint.NET

I didn't try ALL shapes, but most, and not all sizes but most.
Somehow I do expect that we will have some applications which might cause issues, let's test this.

### DIFF
--- a/src/Greenshot.Editor/Drawing/CursorContainer.cs
+++ b/src/Greenshot.Editor/Drawing/CursorContainer.cs
@@ -1,30 +1,28 @@
 ﻿/*
- * Greenshot - a free and open source screenshot tool
- * Copyright (C) 2004-2026 Thomas Braun, Jens Klingen, Robin Krom
- * 
- * For more information see: https://getgreenshot.org/
- * The Greenshot project is hosted on GitHub https://github.com/greenshot/greenshot
- * 
- * This program is free software: you can redistribute it and/or modify
- * it under the terms of the GNU General Public License as published by
- * the Free Software Foundation, either version 1 of the License, or
- * (at your option) any later version.
- * 
- * This program is distributed in the hope that it will be useful,
- * but WITHOUT ANY WARRANTY; without even the implied warranty of
- * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
- * GNU General Public License for more details.
- * 
- * You should have received a copy of the GNU General Public License
- * along with this program.  If not, see <https://www.gnu.org/licenses/>.
- */
+* Greenshot - a free and open source screenshot tool
+* Copyright (C) 2004-2026 Thomas Braun, Jens Klingen, Robin Krom
+* 
+* For more information see: https://getgreenshot.org/
+* The Greenshot project is hosted on GitHub https://github.com/greenshot/greenshot
+* 
+* This program is free software: you can redistribute it and/or modify
+* it under the terms of the GNU General Public License as published by
+* the Free Software Foundation, either version 1 of the License, or
+* (at your option) any later version.
+* 
+* This program is distributed in the hope that it will be useful,
+* but WITHOUT ANY WARRANTY; without even the implied warranty of
+* MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+* GNU General Public License for more details.
+* 
+* You should have received a copy of the GNU General Public License
+* along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
 
 using System;
 using System.Drawing;
-using System.Drawing.Drawing2D;
 using System.IO;
 using System.Runtime.Serialization;
-using System.Windows.Forms;
 using Dapplo.Windows.Common.Structs;
 using Dapplo.Windows.Icons;
 using Greenshot.Base.Interfaces;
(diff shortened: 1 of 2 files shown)


Lakritzator:
This should solve most issues reported in e.g. #898 or #863

Lakritzator:
If one of you, @danrhodes and @Christian-Schulz, has some time I would love a second opinion.
This was quite complex, had many issues before I came to the current solution.

We also need to fix the loading of the previous file formats, I probably need to keep the old logic parallel so it's backwards compatible..

Lakritzator:
Crazy cursors:
image

image

image

Lakritzator:
Already found my first error, the inverted cursor no longer works, I did a small change before I quit, it seems to break that one. Those pesky changes😂

Ah, no... it's not a glitch, it's a bug that the export doesn't work correctly for the inverted cursor, this probably has to do with the way the image is used. Up to the editor things now work.

danrhodes:
I'll take a look later 👍

Christian-Schulz:
> We also need to fix the loading of the previous file formats, I probably need to keep the old logic parallel so it's backwards compatible..

I can take a closer look at the serialization tonight. I have some experience with it. 😉

Christian-Schulz:
I have a general question.
Why do we need a dedicated `DrawableContainer` for a Cursor? 
After capturing we could generate a PNG from `CapturedCursor` object and save it in an `ImageContainer`.
Am I missing something?

Christian-Schulz:
I don't think we need to consider backward compatibility here.
I can't find any old version (up to v1.2 ) that could create and save an `CursorContainer`. 

All old versions are creating an `IconContainer` if cursor was on the capture. Only the property name in `Surface` is `_cursorContainer` but the type was allways `IconContainer`.

I think the `CursorContainer` was never used.

Christian-Schulz:
Tested Affinity Studio.
The moving-tool has a hand as cursor. 
After starting the capture the hand is getting bigger.
Capture_cursor

All other custom cursor in this tool are ok.

Lakritzator:
@Christian-Schulz you still can walk away from this, before you get sucked into the rabbit hole 😅

Lakritzator:
Still there? Are you sure? Walk away 😂

Lakritzator:
I wish it was that easy, I wouldn't have wasted multiple evenings on this...

Lakritzator:
So the easiest reason for not having "just an image container" is the hotspot, there is information with the cursor which specifies where the active "hotspot" is. That what defines with what you select/click. It could be the point of an arrow or the middle of a cross.

Lakritzator:
But that is nothing... So a cursor could be an image, with alpha information. But it could also be a monochrome cursor... It contains a mask to clear pixels and information on what to draw. And it could also use xor to draw, invert the pixels below parts of the cursor, see the inverted cursor in the settings.
To store that we need more then just one bitmap. And the drawing routineisa bit weird.

Of course it can have different sizes, and non of that information is easily available, so we have to guess a bit...

Getting the information is in the bowels of GDi, so really fun stuff

Lakritzator:
With the fix following in a few minutes, the invert cursor should also work:

image

Lakritzator:
@Christian-Schulz 
> Tested Affinity Studio. The moving-tool has a hand as cursor. After starting the capture the hand is getting bigger.
> All other custom cursor in this tool are ok.
I would like to find something I have myself, so I can test it a bit...
I THINK, as some information isn't available and my code needs to guess, the size of the hand is exactly the base size and you have defined a larger mouse-cursor.

The code for that is here: https://github.com/dapplo/Dapplo.Windows/blob/master/src/Dapplo.Windows.Icons/CursorHelper.cs#L145

I need examples to solve the missing information...

Lakritzator:
I found some **possible** solution, or at least a part of it, that is to take the window below the cursor into account.

- WindowFromPoint(): Returns the handle (HWND) of the window at those coordinates.
- GetGUIThreadInfo(): If an application is dragging something or has temporarily locked the mouse to itself, it will use SetCapture(). You must check the hwndCapture field of the GUITHREADINFO struct to see if a window has captured the mouse; if so, use that HWND instead.
- GetWindowThreadProcessId(): Pass the HWND you found here to retrieve the exact Process ID (PID) of the application.

But I need to consider how I can use that information reliability.

Christian-Schulz:
These are understandable reasons why we need the CursorContainer. I surrender. 😊

I would not recommend serializing the CapturedCursor from Dapplo. 
The CursorContainer should define its own internal type for serialization and create a new CapturedCursor at OnDeserialized().
I can extend the class and show you what I mean. But not until the end of the week.

And yes, the Windows API is not my friend.

Lakritzator:
I agree, it's just a transition and not finalized, I needed some kind of container until I was sure about what I needed to "store" to pass it to the drawing method.

If it helps, I could change the Dapplo code to use out parameters for all the values, the drawing logic will just take the needed parameters.

But it helped me a lot to separate some generic code, and have test cases etc for it, in a project which supports this way of working. And Greenshot wasn't it... generalization of code does make the API design slightly better. (Not saying the APIs in Dapplo.Windows are great though)

I welcome your advice.

Lakritzator:
If most of the suggestions are missing spaces... then a review is sad. The point with the file is something I agree upon, we should use the same logic as the capture cursor, as this would support the same drawing capabilities.

Christian-Schulz:
Here are my example with a dedicated type for CaptureCursor for serialization. 
https://github.com/Christian-Schulz/greenshot/commit/fac2a80ca88363191cc4020370aafa02c1a16827

I tested save/open file and copy&paste between different editor windows.

Note there seems to be a bug in main branch. The editor does not open. 
That's why I go one commit to the past.

Christian-Schulz:
> Note there seems to be a bug in main branch. The editor does not open. That's why I go one commit to the past.

#981 fixes this

Lakritzator:
@Christian-Schulz I can fix the Affinity Studio issue some time later, or maybe not at all, if we can find only one cursor doing so. It was wrong most of the time before, now it's wrong in edge cases....

Christian-Schulz:
Yes, the solution is now 10 times better.

Issue #975, opened 2026-02-19
Slow to copy to clipboard


danrhodes:
**Describe the bug**
I often find when taking a screenshot and pasting straight after my screenshot is not yet on my clipboard. It can take 5+ seconds to appear. Sometimes it is straight there? I believe this is due to the destinations I have set -

1- save directly
2- open in editor 
3- save to clipboard

**To Reproduce**
Example steps to reproduce the behavior:
1. Take a screenshot with destinations set as above

**Expected behavior**
The screenshot should be immediately available on my clipboard to paste

**Additional context**
I suspect changing the order of the destinations will fix this as my assumption is anybody using it in the way would prefer the images on the clipboard immediately and having them saved and then opened in the editor can wait a few seconds.

Ver 1.4.112
Windows 11 25h2
Core ultra 7 / ssd / 32gb ram

PR #976, opened 2026-02-19
Avoid redundant surface rendering for exports

danrhodes:
The clipboard delay bug was caused by all capture destinations running sequentially on the UI thread, with the clipboard sitting behind file save and editor in the execution order — sometimes causing a 5+ second wait before the screenshot was available to paste, even on high spec devices.

The bottleneck isn't disk speed but CPU work: rendering the surface to a bitmap and compressing it is the expensive step, and it was happening multiple times. I've made three changes to fix this.

**First**, clipboard now runs at the highest priority, so it executes before anything else, making screenshots instantly available to paste.

**Second**, the file save, encode and write is now done on a background thread — we render the image once on the UI thread as required by GDI+, then hand the resulting bitmap off to a background thread for the slow PNG/JPEG compression and disk write, so it no longer blocks the clipboard or the editor from opening.

**Third, since both clipboard and file save previously rendered the surface independently, we now share a single pre-rendered bitmap** between them, cutting the number of expensive full-image render passes in half. As a further improvement, the HTML clipboard format was also re-rendering the surface a second time internally to write a temp file, which I've fixed to reuse the already-rendered bitmap instead.

https://github.com/greenshot/greenshot/issues/975

### DIFF
--- a/src/Greenshot/Destinations/ClipboardDestination.cs
+++ b/src/Greenshot/Destinations/ClipboardDestination.cs
@@ -43,7 +43,7 @@ namespace Greenshot.Destinations
 
         public override int Priority
         {
-            get { return -1; }
+            get { return 2; }
         }
 
         public override Keys EditorShortcutKeys
--- a/src/Greenshot/Destinations/EmailDestination.cs
+++ b/src/Greenshot/Destinations/EmailDestination.cs
@@ -34,7 +34,7 @@ namespace Greenshot.Destinations
     /// <summary>
     /// This is the EmailDestination, used for MAPI clients.
     /// </summary>
-    public class EmailDestination : AbstractDestination
+    public class EmailDestination : AbstractDestination, IAcceptsPreRenderedImage
     {
         private static readonly Image MailIcon = GreenshotResources.GetImage("Email.Image");
         private static bool _isActiveFlag;
@@ -96,13 +96,19 @@ namespace Greenshot.Destinations
 
         public override Image DisplayIcon => MailIcon;
 
-        // TODO: Implement IAcceptsPreRenderedImage to avoid a redundant surface render pass
-        // when a shared rendered bitmap is already available from the capture pipeline.
-        // SendImage would need an overload accepting a pre-rendered Image instead of ISurface.
         public override ExportInformation ExportCapture(bool manuallyInitiated, ISurface surface, ICaptureDetails captureDetails)
         {
             ExportInformation exportInformation = new ExportInformation(Designation, Description);
             MapiMailMessage.SendImage(surface, captureDetails);
+            exportInformation.ExportMade = true;
+            ProcessExport(exportInformation, surface);
+            return exportInformation;
+        }
+
+        public ExportInformation ExportCaptureWithRenderedImage(Image preRenderedImage, ISurface surface, ICaptureDetails captureDetails)
+        {
+            ExportInformation exportInformation = new ExportInformation(Designation, Description);
+            MapiMailMessage.SendImage(preRenderedImage, captureDetails);
             exportInformation.ExportMade = true;
             ProcessExport(exportInformation, surface);
             return exportInformation;
--- a/src/Greenshot/Forms/MainForm.cs
+++ b/src/Greenshot/Forms/MainForm.cs
@@ -329,6 +329,7 @@ namespace Greenshot.Forms
 
         public MainForm(CopyDataTransport dataTransport)
         {
+            SimpleServiceProvider.Current.AddService(SynchronizationContext.Current);
             var uiContext = TaskScheduler.FromCurrentSynchronizationContext();
             SimpleServiceProvider.Current.AddService(uiContext);
  
--- a/src/Greenshot/Helpers/CaptureHelper.cs
+++ b/src/Greenshot/Helpers/CaptureHelper.cs
@@ -768,9 +768,7 @@ namespace Greenshot.Helpers
             int destinationCount = captureDetails.CaptureDestinations.Count;
             if (destinationCount > 0)
             {
-                // Capture the UI SynchronizationContext now, before any background threads are started,
-                // so background tasks can marshal back to the UI thread when needed (e.g. clipboard).
-                var uiContext = SynchronizationContext.Current;
+                var uiContext = SimpleServiceProvider.Current.GetInstance<SynchronizationContext>();
 
                 // Pre-render the surface once on the UI thread if any destination that needs a bitmap
                 // is active. This shared bitmap is passed to both clipboard and file save, avoiding
@@ -778,8 +776,6 @@ namespace Greenshot.Helpers
                 bool hasFileDestination = captureDetails.CaptureDestinations.Exists(d =>
                     d.Designation == nameof(WellKnownDestinations.FileNoDialog) ||
                     d.Designation == nameof(WellKnownDestinations.FileDialog));
-                bool hasClipboardDestination = captureDetails.CaptureDestinations.Exists(d =>
-                    d.Designation == nameof(WellKnownDestinations.Clipboard));
 
                 // If quality prompt is enabled, show it once here for all file destinations rather
                 // than once per destination, so the user is only asked once and all file saves
@@ -791,9 +787,12 @@ namespace Greenshot.Helpers
                     qualityDialog.ShowDialog();
                 }
 
+                bool hasPreRenderDestination = hasFileDestination ||
+                    captureDetails.CaptureDestinations.Exists(d => d is IAcceptsPreRenderedImage);
+
                 Image sharedRenderedBitmap = null;
                 bool disposeSharedBitmap = false;
-                if (hasFileDestination || hasClipboardDestination)
+                if (hasPreRenderDestination)
                 {
                     disposeSharedBitmap = ImageIO.CreateImageFromSurface(surface, sharedFileOutputSettings, out sharedRenderedBitmap);
                 }
@@ -875,7 +874,9 @@ namespace Greenshot.Helpers
                             {
                                 Log.InfoFormat("Not overwriting: {0}", ex1.Message);
                                 // File already exists and overwrite is disallowed — fall back to save dialog on UI thread.
-                                uiContext?.Post(_ => ImageIO.SaveWithDialog(surface, captureDetails), null);
+                                // Use Send (not Post) so the background thread blocks until the dialog completes,
+                                // preventing surface disposal before SaveWithDialog has finished.
+                                uiContext?.Send(_ => ImageIO.SaveWithDialog(surface, captureDetails), null);
                             }
                             catch (Exception ex2)
                             {
@@ -907,34 +908,11 @@ namespace Greenshot.Helpers
                 }
                 finally
                 {
-                    // Schedule disposal of the shared rendered bitmap once all background file
-                    // saves complete, without blocking the UI thread. If there are no background
-                    // tasks, disposal is handled synchronously in the finally block below.
-                    if (backgroundTasks.Count > 0 && disposeSharedBitmap && sharedRenderedBitmap != null)
+                    if (backgroundTasks.Count > 0)
                     {
-                        var bitmapToDispose = sharedRenderedBitmap;
-                        disposeSharedBitmap = false;
-                        sharedRenderedBitmap = null;
-
-                        Task.WhenAll(backgroundTasks.ToArray())
-                            .ContinueWith(
-                                t =>
-                                {
-                                    try
-                                    {
-                                        bitmapToDispose.Dispose();
-                                    }
-                                    catch
-                                    {
-                                        // Ignore exceptions during bitmap disposal in background continuation.
-                                    }
-                                },
-                                CancellationToken.None,
-                                TaskContinuationOptions.None,
-                                TaskScheduler.Default);
+                        Task.WaitAll(backgroundTasks.ToArray());
                     }
 
-                    // If no asynchronous disposal was scheduled, dispose the shared rendered bitmap now.
                     if (disposeSharedBitmap)
                     {
                         sharedRenderedBitmap?.Dispose();
--- a/src/Greenshot.Base/Core/ImageIO.cs
+++ b/src/Greenshot.Base/Core/ImageIO.cs
@@ -440,6 +440,43 @@ namespace Greenshot.Base.Core
         }
 
         /// <summary>
+        /// Create a tmpfile from a pre-rendered bitmap using the name from the configured pattern.
+        /// Used e.g. by the email export when a shared rendered bitmap is already available.
+        /// </summary>
+        /// <param name="renderedImage">Pre-rendered bitmap; not disposed by this method.</param>
+        /// <param name="captureDetails"></param>
+        /// <param name="outputSettings"></param>
+        /// <returns>Path to image file</returns>
+        public static string SaveNamedTmpFile(Image renderedImage, ICaptureDetails captureDetails, SurfaceOutputSettings outputSettings)
+        {
+            string pattern = CoreConfig.OutputFileFilenamePattern;
+            if (string.IsNullOrEmpty(pattern?.Trim()))
+            {
+                pattern = "greenshot ${capturetime}";
+            }
+
+            string filename = FilenameHelper.GetFilenameFromPattern(pattern, outputSettings.Format, captureDetails);
+            filename = Regex.Replace(filename, @"[^\d\w\.]", "_");
+            filename = Regex.Replace(filename, @"_+", "_");
+            string tmpFile = Path.Combine(Path.GetTempPath(), filename);
+
+            Log.Debug("Creating TMP File: " + tmpFile);
+
+            try
+            {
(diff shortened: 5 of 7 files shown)


Lakritzator:
Wow... but I need to check this a bit in detail, very crucial path...
Just had an extremely long day at work, so need to check if I have the energy today, but why not test to see what copilot thinks of it 🤷‍♂️

danrhodes:
Indeed and totally recommend this is thoroughly checked. My initial testing was fine but appreciate a second and even third pair of eyes.

I've noticed a slowdown in images being available on the clipboard a lot recently. Even to the point of having pasted the previous image into a doc as the latest one was still being written to clipboard. 

This all on a high spec ultra core 7 and ultra core 9 machine so cannot be due to old slow hardware which sent me down this route.

Lakritzator:
Hmm, the points written down by copilot aren't complete nonsense, I guess it's worth considering them... I guess I will start using this feature more often, it is like a (not that bad) second opinion.

danrhodes:
I've read through all the comments and all valid points. I will address them later. I'm going to build this version and use it today too to find any bugs.

src/Greenshot/Helpers/CaptureHelper.cs
```diff
@@ -763,8 +767,31 @@ private void HandleCapture()
...
+                {
+                    var sharedOutputSettings = new SurfaceOutputSettings();
+                    disposeSharedBitmap = ImageIO.CreateImageFromSurface(surface, sharedOutputSettings, out sharedRenderedBitmap);
+                }
+
+                var backgroundTasks = new List<Task>();
+
+                try
+                {
                 foreach (IDestination destination in captureDetails.CaptureDestinations)
```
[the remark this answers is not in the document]
danrhodes: Explicitly sorting CaptureDestinations by Priority then Description at the loop site to guarantee clipboard (priority -1) always executes first, regardless of the order destinations were added to the list.

src/Greenshot/Helpers/CaptureHelper.cs
```diff
@@ -774,10 +801,105 @@ private void HandleCapture()
...
+                        {
+                            // User cancelled the filename dialog — skip this destination.
+                            continue;
+                        }
+
+                        if (CoreConfig.OutputFilePromptQuality)
+                        {
+                            var qualityDialog = new QualityDialog(outputSettings);
+                            qualityDialog.ShowDialog();
+                        }
```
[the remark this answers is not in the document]
danrhodes: Moved the quality dialogue before the destination loop so it shows once and its settings are shared across all file destinations, preventing multiple prompts and any inconsistency between them.

src/Greenshot/Helpers/CaptureHelper.cs
```diff
@@ -774,10 +801,105 @@ private void HandleCapture()
...
+                                    CoreConfig.OutputFileCopyPathToClipboard, uiContext);
+
+                                // Update the config path on completion — marshal to UI thread for thread-safety.
+                                uiContext?.Post(_ => CoreConfig.OutputFileAsFullpath = bgFullPath, null);
+                            }
+                            catch (ArgumentException ex1)
+                            {
+                                Log.InfoFormat("Not overwriting: {0}", ex1.Message);
+                                // File already exists and overwrite is disallowed — fall back to save dialog on UI thread.
+                                uiContext?.Post(_ => ImageIO.SaveWithDialog(surface, captureDetails), null);
```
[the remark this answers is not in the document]
danrhodes: Acknowledged — SaveWithDialog is already marshalled to the UI thread via uiContext.Post so there's no threading issue; changing the error UX from modal dialog to non-blocking notification is out of scope for this PR and can be addressed separately.

src/Greenshot/Helpers/CaptureHelper.cs
```diff
@@ -774,10 +801,105 @@ private void HandleCapture()
...
+                        });
+
+                        backgroundTasks.Add(task);
+                    }
+                    else if (destination.Designation == nameof(WellKnownDestinations.Clipboard) &&
+                             sharedRenderedBitmap != null &&
+                             destination is ClipboardDestination clipboardDest)
+                    {
+                        // Pass the shared pre-rendered bitmap — avoids a second render pass.
+                        clipboardDest.ExportCaptureWithRenderedImage(sharedRenderedBitmap, surface, captureDetails);
```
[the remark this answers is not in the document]
danrhodes: Good call — implemented IAcceptsPreRenderedImage in Greenshot.Base/Interfaces/ as a proper marker interface. ClipboardDestination now implements it publicly, CaptureHelper dispatches via the interface rather than a type check, and EmailDestination/PrinterDestination have TODO comments marking them as future candidates once their respective helpers (MailHelper.SendImage, PrintHelper) are updated to accept a pre-rendered Image. The Greenshot.Destinations using is also removed from CaptureHelper since there's no longer any direct type reference.

src/Greenshot/Helpers/CaptureHelper.cs
```diff
@@ -774,10 +801,105 @@ private void HandleCapture()
...
+                            canDisposeSurface = false;
+                        }
+                    }
+                }
+
+                // Wait for all background file saves to complete before allowing surface disposal.
+                if (backgroundTasks.Count > 0)
+                {
+                    Task.WaitAll(backgroundTasks.ToArray());
+                }
```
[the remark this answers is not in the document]
danrhodes: Moved Task.WaitAll into the finally block — if an exception is thrown mid-loop, tasks now always complete before surface disposal proceeds.

danrhodes:
@Lakritzator all comments from Co-Pilot have been addressed.

Lakritzator:
Great, more to review 😂

What do you think about delayed rendering? The means we first place the requested format to the clipboard after it's requested. Saves memory and time. I have support for this in my much better clipboard code in Dapplo.Windows

https://github.com/dapplo/Dapplo.Windows/blob/master/doc/articles/clipboard-usage.md#delayed-rendering

Lakritzator:
Looking at this, which is good, but it does show that we would have a better solution if we internally use ImageSharp. We would only need one image in the memory, as it's thread safe, unless we use it where it can only be a bitmap like the editor.

danrhodes:
I'll take a look later, two kids birthday parties this weekend to take the little ones to 🤣

danrhodes:
Managed to look over your suggestions @Lakritzator both great ideas, and Iagree with the direction, delayed rendering would eliminate the upfront format cost entirely, and moving to ImageSharp internally would remove the need for per-task clones. I am happy to look at both as follow-up PRs, as they're each a meaningful refactor in their own right.

danrhodes:
@Lakritzator should I look at the further suggests outside the real scope of this pr as part of this pr or would it better we merge this and look at image sharp and delayed rendering as they own independent pr's?

Lakritzator:
Let's not add ImageSharp and delayed rendering here, just wanted to point out that we are solving problems because we are inside a trouble bubble.

Anyway, before I merge / double check, did you test with the destination picker? I usually only use that, this is why I never noticed the issue.

danrhodes:
Hey, apologies, yes tested with selecting options to go direct to save, clipboard, etc and also tested with my new favourite method, destination picker!

danrhodes:
Completed additional testing with this pr version today, and no issues either using destination picker, direct save, copy to clipboard or open in image editor. All looks good to me :-)

Lakritzator:
First code pass looks okay, but I need to see it inside VS , as I'm missing some key hints in the code when doing the review in the browser...

danrhodes:
Absolutely 💯 recommend a full review. No rush 👍

src/Greenshot/Destinations/ClipboardDestination.cs
```diff
@@ -43,7 +43,7 @@ public override string Description
 
         public override int Priority
         {
-            get { return 2; }
+            get { return -1; }
```
Lakritzator: I'm not sure about this, the priority is used for setting the display order, you just  placed the clipboard entry somewhere else.

src/Greenshot/Destinations/EmailDestination.cs
```diff
@@ -96,6 +96,9 @@ public override bool IsActive
 
         public override Image DisplayIcon => MailIcon;
 
+        // TODO: Implement IAcceptsPreRenderedImage to avoid a redundant surface render pass
```
Lakritzator: Any particular reason not to supply this?

src/Greenshot/Helpers/CaptureHelper.cs
```diff
@@ -763,9 +768,41 @@ private void HandleCapture()
             int destinationCount = captureDetails.CaptureDestinations.Count;
             if (destinationCount > 0)
             {
-                // Flag to detect if we need to create a temp file for the email
-                // or use the file that was written
-                foreach (IDestination destination in captureDetails.CaptureDestinations)
+                // Capture the UI SynchronizationContext now, before any background threads are started,
+                // so background tasks can marshal back to the UI thread when needed (e.g. clipboard).
+                var uiContext = SynchronizationContext.Current;
```
Lakritzator: You should actually be able to use the context from here: https://github.com/greenshot/greenshot/blob/764ffa7896103fb7d17dd68a3b4642b955204a33/src/Greenshot/Forms/MainForm.cs#L333

src/Greenshot/Helpers/CaptureHelper.cs
```diff
@@ -774,10 +811,133 @@ private void HandleCapture()
...
+                                }
+
+                                // Update the config path on completion — marshal to UI thread for thread-safety.
+                                uiContext?.Post(_ => CoreConfig.OutputFileAsFullpath = bgFullPath, null);
+                            }
+                            catch (ArgumentException ex1)
+                            {
+                                Log.InfoFormat("Not overwriting: {0}", ex1.Message);
+                                // File already exists and overwrite is disallowed — fall back to save dialog on UI thread.
+                                uiContext?.Post(_ => ImageIO.SaveWithDialog(surface, captureDetails), null);
```
Lakritzator: I'm not sure here,  but post suggests async to me, meaning the code will just continue, I guess to the finally... Would need to check though, I am not sure and my internet connection in the train is a bit dodgy at ~200KM/h

src/Greenshot/Helpers/CaptureHelper.cs
```diff
@@ -774,10 +811,133 @@ private void HandleCapture()
...
+                    }
+                }
+
+                }
+                finally
+                {
+                    // Schedule disposal of the shared rendered bitmap once all background file
+                    // saves complete, without blocking the UI thread. If there are no background
+                    // tasks, disposal is handled synchronously in the finally block below.
+                    if (backgroundTasks.Count > 0 && disposeSharedBitmap && sharedRenderedBitmap != null)
```
Lakritzator: Is there a possibility that there are background tasks and the rest of the expression says false? Not awaiting background tasks is not a good practise.

src/Greenshot/Helpers/CaptureHelper.cs
```diff
@@ -774,10 +811,133 @@ private void HandleCapture()
...
+                                        bgFullPath,
+                                        bgOverwrite,
+                                        bgOutputSettings,
+                                        CoreConfig.OutputFileCopyPathToClipboard,
+                                        uiContext);
+                                }
+
+                                // Update the config path on completion — marshal to UI thread for thread-safety.
+                                uiContext?.Post(_ => CoreConfig.OutputFileAsFullpath = bgFullPath, null);
+                            }
```
Lakritzator, line 872: Is this needed? Maybe we need to look at handling the changes in such a way that we do not need a uicontext...

src/Greenshot/Helpers/CaptureHelper.cs
```diff
@@ -763,9 +768,41 @@ private void HandleCapture()
             int destinationCount = captureDetails.CaptureDestinations.Count;
             if (destinationCount > 0)
             {
-                // Flag to detect if we need to create a temp file for the email
-                // or use the file that was written
-                foreach (IDestination destination in captureDetails.CaptureDestinations)
+                // Capture the UI SynchronizationContext now, before any background threads are started,
+                // so background tasks can marshal back to the UI thread when needed (e.g. clipboard).
```
Lakritzator: I learned there is a way to prevent the need for a UI (STA) thread when using the clipboard, might look into that another day... And we need  good serialization to json of our objects

Lakritzator:
If we also modify the other destinations to use IAcceptsPreRenderedImage, do we need to modify the CaptureHelper?

danrhodes:
I'll take a look later thanks for the feedback 👍

src/Greenshot/Destinations/EmailDestination.cs
danrhodes: None tbh. Good catch, no particular reason not to — fixed. Added a SendImage(Image, ICaptureDetails) overload to MapiMailMessage and EmailDestination now implements IAcceptsPreRenderedImage directly, TODO removed. Will be in the next push.

src/Greenshot/Helpers/CaptureHelper.cs
danrhodes: MainForm now registers the SynchronizationContext into SimpleServiceProvider at startup, and CaptureHelper retrieves it from there instead of relying on SynchronizationContext.Current at the call site.

src/Greenshot/Helpers/CaptureHelper.cs
danrhodes: Good point — OutputFileAsFullpath doesn't actually need marshalling (FileDestination sets it directly), and the clipboard copy in SaveRenderedImage could similarly drop the uiContext dependency. The two dialog fallbacks (SaveWithDialog, MessageBox) do genuinely need the UI thread. Happy to simplify this in a follow-up PR — or if you'd prefer it addressed here, let me know.

src/Greenshot/Helpers/CaptureHelper.cs
danrhodes: You're right — Post is async/fire-and-forget, so on the SaveWithDialog fallback path (line 876) the background task completes immediately after posting, the finally runs, and surface may be disposed before the dialogue actually executes on the UI thread. We should use Send instead of Post there to block the background thread until the UI thread finishes the dialogue — or rethink that fallback path entirely. Fixed

src/Greenshot/Helpers/CaptureHelper.cs
danrhodes: You're right on both counts. This is why we do code reviews! We now Task.WaitAll unconditionally before disposal, removing the buggy conditional and the fire-and-forget ContinueWith chain entirely.

src/Greenshot/Destinations/ClipboardDestination.cs
danrhodes: The -1 was a leftover from an earlier version where execution order mattered for the pre-render logic. Now that clipboard uses IAcceptsPreRenderedImage, and file saves run on background tasks, the ordering between them is irrelevant. Reverted to the original 2.

danrhodes:
> If we also modify the other destinations to use IAcceptsPreRenderedImage, do we need to modify the CaptureHelper?

Yes — the pre-render gate was hardcoded to check for file and clipboard destinations by name, so any new IAcceptsPreRenderedImage implementor would have silently fallen back to ExportCapture. Fixed: the gate now checks d is IAcceptsPreRenderedImage directly, so adding the interface to any destination is all that's needed going forward.

src/Greenshot/Forms/MainForm.cs
```diff
@@ -329,6 +329,7 @@ private static void FreeMutex()
 
         public MainForm(CopyDataTransport dataTransport)
         {
+            SimpleServiceProvider.Current.AddService(SynchronizationContext.Current);
```
Lakritzator, line 332: This is not what I had in mind, and maybe my mind was wrong... but why do we now need 2 types which so similar things, maybe we can merge that somehow?
I need to check what this means, but I have my head full of work related topics currently...
https://devblogs.microsoft.com/dotnet/taskscheduler-fromcurrentsynchronizationcontext/

danrhodes: I added the SynchronizationContext registration alongside the existing TaskScheduler because CaptureHelper needs both fire-and-forget marshalling (.Post()) and blocking marshalling (.Send()) back to the UI thread — the blocking .Send() is critical on the SaveWithDialog fallback path to prevent the surface being disposed before the dialogue finishes. That said, I knowyou could achieve the same blocking behaviour with Task.Factory.StartNew(..., uiTaskScheduler).Wait(), which would make the SynchronizationContext registration redundant. I went with SynchronizationContext because .Send() expresses the intent more directly, but I'm happy to consolidate onto just the TaskScheduler if you'd prefer a single registered type, let me know your thoughts :-)

Lakritzator: Let's look at that another day...

Lakritzator:
Good for those people who are crazy enough to have multiple destinations active...

danrhodes:
Yes and originally that was me, I had three but have since moved to destination picker.

Issue #946, opened 2026-02-11
I do not get an error message if the file I want to overwrite cannot be written to


marcOcram:
**Describe the bug**
If I want to save to a file and that file is in use, I get the error window.

**To Reproduce**
Example steps to reproduce the behavior:
1. Capture '....'
2. Save as file
3. Lock that file (e.g. use `File.Open()` and keep the handle alive)
4. Capture '....'
5. Save as same file and overwrite

**Expected behavior**
I expect to get a window that tells me that the file cannot be saved to currently and get the possibility to save to another file.

**Versions (please complete the following information):**
 - Greenshot version: 1.3.312 - 5a60aff83e (64 bit)
 - Windows version: Windows 11 24H2 (26100.7623)

**Additional context**
```
Software version: 1.3.312 - 5a60aff83e (64 bit)
.NET runtime version: 4.0.30319.42000+
Time: 2026-02-11 16:17:39 +01:00
OS: Windows 10 2009 x64 build 26100
GDI object count: 156
User object count: 69

Exception: System.IO.IOException
Message: The process cannot access the file 'xyz.png' because it is being used by another process.

Stack:
   at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
   at System.IO.FileStream.Init(String path, FileMode mode, FileAccess access, Int32 rights, Boolean useRights, FileShare share, Int32 bufferSize, FileOptions options, SECURITY_ATTRIBUTES secAttrs, String msgPath, Boolean bFromProxy, Boolean useLongPath, Boolean checkHost)
   at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access)
   at Greenshot.Base.Core.ImageIO.Save(ISurface surface, String fullPath, Boolean allowOverwrite, SurfaceOutputSettings outputSettings, Boolean copyPathToClipboard)
   at Greenshot.Base.Core.ImageIO.SaveWithDialog(ISurface surface, ICaptureDetails captureDetails)
   at Greenshot.Destinations.FileWithDialogDestination.ExportCapture(Boolean manuallyInitiated, ISurface surface, ICaptureDetails captureDetails)
   at Greenshot.Base.Core.AbstractDestination.<>c__DisplayClass28_0.<ShowPickerMenu>b__4(Object sender, EventArgs e)
   at System.Windows.Forms.ToolStripItem.RaiseEvent(Object key, EventArgs e)
[...]

PR #986, opened 2026-02-21
Handle IO exceptions when saving image

danrhodes:
Root cause: System.IO.IOException is thrown when the file is locked by another process. It is not a subclass of ExternalException (which is COM/interop-related), so the original catch (ExternalException) block never fired, and the exception propagated all the way up, causing the crash report.

Fix: Changed to a when filter pattern — catch (Exception e) when (e is ExternalException || e is IOException) — so both cases are caught and the user sees the existing "Cannot save file to {0}" error dialogue instead of a crash.

Behaviour after the fix: The user gets the error dialogue, dismisses it, and SaveWithDialog returns null. The caller (FileWithDialogDestination.cs:60) already checks for null and skips updating the path — so the user can simply invoke Save As again and choose a different filename.

Fixes https://github.com/greenshot/greenshot/issues/946

### DIFF
--- a/src/Greenshot.Base/Core/ImageIO.cs
+++ b/src/Greenshot.Base/Core/ImageIO.cs
@@ -380,7 +380,7 @@ namespace Greenshot.Base.Core
                     returnValue = fileNameWithExtension;
                     IniConfig.Save();
                 }
-                catch (ExternalException)
+                catch (Exception e) when (e is ExternalException || e is IOException)
                 {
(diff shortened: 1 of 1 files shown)


Lakritzator:
Great 👍 

I was also going for this but, only I took the long route:
https://github.com/dapplo/Dapplo.Windows/tree/master/src/Dapplo.Windows.Dialogs

This should solve the max path issues, made it yesterday/this morning.

Will create a PR where all dialogs move towards the more modern versions… and while doing so, I planned to fix the exceptions, like this, but I’ll take it now

Issue #939, opened 2026-02-10
Exception: System.ObjectDisposedException


octaeder:
Make a screenshot, a panel opens (option) for target selection, then press escape (or choose clipboard and immediately switch to another app (alt+tab)).

Software version: 1.4.93 - 9a0b16a7dd (64 bit)
.NET runtime version: 4.0.30319.42000+
Time: 2026-02-10 12:22:55 +01:00
OS: Windows 11 x64 build 26200
GDI object count: 111
User object count: 46

Exception: System.ObjectDisposedException
Message: Cannot access a disposed object.
Object name: 'ContextMenuStrip'.

Stack:
   at System.Windows.Forms.Control.CreateHandle()
   at System.Windows.Forms.ToolStripDropDown.CreateHandle()
   at System.Windows.Forms.Control.get_Handle()
   at System.Windows.Forms.ToolStripDropDown.ReparentToDropDownOwnerWindow()
   at System.Windows.Forms.ToolStripDropDown.SetVisibleCore(Boolean visible)
   at System.Windows.Forms.ToolStripDropDown.set_Visible(Boolean value)
   at System.Windows.Forms.ToolStripDropDown.ProcessDialogKey(Keys keyData)
   at System.Windows.Forms.Control.PreProcessMessage(Message& msg)
   at System.Windows.Forms.Control.PreProcessControlMessageInternal(Control target, Message& msg)
   at System.Windows.Forms.Application.ThreadContext.PreTranslateMessage(MSG& msg)

Configuration dump:

PR #987, opened 2026-02-21
Update AbstractDestination.cs

danrhodes:
Fixes an ObjectDisposedException crash that occurred when pressing Escape or switching away (Alt+Tab) immediately after selecting a destination from the capture picker menu. The root cause was that ContextMenuStrip.Dispose() was being called synchronously inside event handlers that Windows Forms was still executing, causing subsequent internal calls to SetVisibleCore and ProcessDialogKey to hit an already-disposed object. All disposal calls are now deferred via BeginInvoke with an IsDisposed guard to prevent double-disposal, and an additional IsDisposed check was added before the AppFocusChange close call to guard against the race condition where an alt-tab fires after an item-click has already begun teardown.

Fixes https://github.com/greenshot/greenshot/issues/939

### DIFF
--- a/src/Greenshot.Base/Core/AbstractDestination.cs
+++ b/src/Greenshot.Base/Core/AbstractDestination.cs
@@ -212,7 +212,10 @@ namespace Greenshot.Base.Core
                         else
                         {
                             Log.DebugFormat("Letting the menu 'close' as the tag is set to '{0}'", menu.Tag);
-                            menu.Close();
+                            if (!menu.IsDisposed)
+                            {
+                                menu.Close();
+                            }
                         }
 
                         break;
@@ -230,7 +233,10 @@ namespace Greenshot.Base.Core
                         // We might already be in the disposing process, so queue the disposal to avoid re-entrancy
                         menu.BeginInvoke(new Action(() =>
                         {
-                            menu.Dispose();
+                            if (!menu.IsDisposed)
+                            {
+                                menu.Dispose();
+                            }
                         }));
                         break;
                     default:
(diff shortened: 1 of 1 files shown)


src/Greenshot.Base/Core/AbstractDestination.cs
```diff
@@ -230,7 +233,10 @@ public ExportInformation ShowPickerMenu(bool addDynamics, ISurface surface, ICap
                         // We might already be in the disposing process, so queue the disposal to avoid re-entrancy
                         menu.BeginInvoke(new Action(() =>
                         {
-                            menu.Dispose();
+                            if (!menu.IsDisposed)
```
Lakritzator, line 236: Shouldn't dispose be reentrant?

danrhodes: Hey,

Great question! In an ideal world, yes — Dispose() should be reentrant per the IDisposable contract.

However, WinForms controls (including ContextMenuStrip) don't reliably guarantee this, particularly in async scenarios involving BeginInvoke where the control handle may have already been destroyed by the time the invoked action runs.

The existing code already had a comment acknowledging re-entrancy as a concern: "We might already be in the disposing process, so queue the disposal to avoid re-entrancy". This PR applies that same defensive pattern consistently to the other disposal sites that were missing it.

The IsDisposed guard is a belt-and-braces approach that prevents a potential ObjectDisposedException at the cost of a single property check — a negligible overhead. If Greenshot's WinForms components were confirmed to implement fully reentrant disposal, I'd agree the guard is redundant, but given the uncertainty here I think it's the safer choice.

Lakritzator: I hate it when it's not doing what it is supposed to do, even doing an if check, will not guard 100% of the time...

Issue #925, opened 2026-02-06
greenshot crashes using the crop tool in editing screen


Ergotroyano:
**Describe the bug**
when I finish the area selection for cropping the image and release de mouse button an error screen pops up.

**To Reproduce**
Example steps to reproduce the behavior:
1. Capture '....'
2. Open Editor (set to automatically open after taking screenshot)
3. select de crop tool
4. drag selection
5. release the mouse button

**Expected behavior**
I'd except the program to show the area selected and remove any part of the image outside the selected area and be able to keep editing the image.

**Screenshots**

<img width="3071" height="1831" alt="Image" src="https://github.com/user-attachments/assets/7d99a9bf-5675-4494-874d-d5cfd57e6580" />

**Versions (please complete the following information):**
 - Greenshot version: Greenshot 1.3.312 - 5a60aff83e (64) bit)
 - Windows version: Windows 11 Home 24H2 26100.7462

**Additional context**
If I close de error screen the instance never recovers

<img width="3070" height="1831" alt="Image" src="https://github.com/user-attachments/assets/95961bbc-4ac9-430b-9307-c0681d958360" />

this is the message contained in the error screen:

Software version: 1.3.312 - 5a60aff83e (64 bit)
.NET runtime version: 4.0.30319.42000+
Time: 2026-02-06 16:00:08 -03:00
OS: Windows 10 2009 x64 build 26100
GDI object count: 433
User object count: 360

Exception: System.NullReferenceException
Message: Object reference not set to an instance of an object.

Stack:
   at Greenshot.Editor.Drawing.Adorners.AbstractAdorner.get_BoundsOnSurface()
   at Greenshot.Editor.Drawing.Adorners.AbstractAdorner.Paint(PaintEventArgs paintEventArgs)
   at Greenshot.Editor.Drawing.Surface.SurfacePaint(Object sender, PaintEventArgs paintEventArgs)
   at System.Windows.Forms.Control.OnPaint(PaintEventArgs e)
   at System.Windows.Forms.Control.PaintWithErrorHandling(PaintEventArgs e, Int16 layer)
   at System.Windows.Forms.Control.WmPaint(Message& m)
   at System.Windows.Forms.Control.WndProc(Message& m)
   at System.Windows.Forms.NativeWindow.Callback(IntPtr hWnd, Int32 msg, IntPtr wparam, IntPtr lparam)

Configuration dump:

danrhodes:
@danrhodes reminder.

[...]

PR #1003, opened 2026-02-23
Update AbstractAdorner.cs to fix issues/925

danrhodes:
I couldnt replicate either but found why this caused the crash - Moving var bounds = BoundsOnSurface; inside the existing try/catch block in AbstractAdorner.Paint fixes a NullReferenceException that crashed the editor whenever the user released the mouse after drawing a crop selection. The property BoundsOnSurface calls Owner.Parent.ToSurfaceCoordinates(...), but Parent is set to null by RemoveElement during crop teardown before the pending WM_PAINT message is processed — meaning a repaint could fire against a container that was already mid-disposal. That call was sitting above the try/catch that already existed to suppress paint-time errors (noted as BUG-2065), so it was completely unprotected. Moving it one line down into the same try block means any NullReferenceException from a partially-torn-down adorner is silently swallowed during that paint cycle — the adorner simply draws nothing, which is the correct behaviour since the container is already being removed anyway.

Fixes https://github.com/greenshot/greenshot/issues/925

### DIFF
--- a/src/Greenshot.Editor/Drawing/Adorners/AbstractAdorner.cs
+++ b/src/Greenshot.Editor/Drawing/Adorners/AbstractAdorner.cs
@@ -154,13 +154,13 @@ namespace Greenshot.Editor.Drawing.Adorners
         {
             Graphics targetGraphics = paintEventArgs.Graphics;
 
-            var bounds = BoundsOnSurface;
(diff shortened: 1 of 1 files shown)


Lakritzator:
Hard tracking, sensible explanation, "easy" fix. 👍

Issue #1007, opened 2026-02-25
Greenshot crashes when clicking tray icon after changing icon size


Nippius:
**Describe the bug**
Changing icon size crashes the app. 

**To Reproduce**
Example steps to reproduce the behavior:
1. Right click tray icon
2. Click "Preferences"
3. Select "General" tab
4. Change the icon size (I tried with 16 and 32)
5. Click "ok"
6. Click the tray icon again and application crashes

**Expected behavior**
The app shouldn't crash, it should show the tray menu with the new icon size or, at least, the original one until restarted

**Versions (please complete the following information):**
Software version: 1.3.312 - 5a60aff83e (64 bit)
.NET runtime version: 4.0.30319.42000+
Time: 2026-02-25 00:29:56 +00:00
OS: Windows 10 2009 x64 build 19045
GDI object count: 103
User object count: 47

Exception: System.ArgumentException
Message: Parameter is not valid.

Stack:
   at System.Drawing.Image.get_Width()
   at System.Drawing.Image.get_Size()
   at System.Windows.Forms.ToolStripRenderer.CreateDisabledImage(Image normalImage, ImageAttributes imgAttrib)
   at System.Windows.Forms.ToolStripRenderer.OnRenderItemImage(ToolStripItemImageRenderEventArgs e)
   at System.Windows.Forms.ToolStripProfessionalRenderer.OnRenderItemImage(ToolStripItemImageRenderEventArgs e)
   at System.Windows.Forms.ToolStripRenderer.DrawItemImage(ToolStripItemImageRenderEventArgs e)
   at System.Windows.Forms.ToolStripMenuItem.OnPaint(PaintEventArgs e)
   at System.Windows.Forms.ToolStripItem.HandlePaint(PaintEventArgs e)
   at System.Windows.Forms.ToolStrip.OnPaint(PaintEventArgs e)
   at System.Windows.Forms.Control.PaintWithErrorHandling(PaintEventArgs e, Int16 layer)
   at System.Windows.Forms.Control.WmPaint(Message& m)
   at System.Windows.Forms.Control.WndProc(Message& m)
   at System.Windows.Forms.ToolStrip.WndProc(Message& m)
   at System.Windows.Forms.ToolStripDropDown.WndProc(Message& m)
   at System.Windows.Forms.NativeWindow.Callback(IntPtr hWnd, Int32 msg, IntPtr wparam, IntPtr lparam)

Configuration dump:

PR #1010, opened 2026-02-25
Fix crash when changing icon size in preferences /issues/1007

danrhodes:
Changing the icon size in Preferences was causing the app to crash the next time you clicked the tray icon. This happened because the placeholder images used in the quick-settings menu were stored in a single shared variable — so when the settings were rebuilt with the new icon size, that shared image got destroyed, but the old menu items were still pointing at it. When Windows tried to paint them, it hit a dead image and threw an exception. The fix gives each menu instance its own image so they can't step on each other, and also corrects a typo that was causing the wrong (unvalidated) icon size value to actually get saved.

Fixes https://github.com/greenshot/greenshot/issues/1007

### DIFF
--- a/src/Greenshot.Base/Core/CoreConfiguration.cs
+++ b/src/Greenshot.Base/Core/CoreConfiguration.cs
@@ -343,7 +343,7 @@ namespace Greenshot.Base.Core
                     {
                         newSize.Height = 16;
                     }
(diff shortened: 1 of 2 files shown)


Lakritzator (approved):

Lakritzator:
😨

Issue #1021, opened 2026-03-02
Error capturing a Windows Toast notification


danrhodes:
When capturing a Windows Toast notification re: an Adobe Update (may not be related), I got the error below.

Software version: 1.4.112 - acf799517f (64 bit)
.NET runtime version: 4.0.30319.42000+
Time: 2026-03-02 19:11:05 +00:00
OS: Windows 11 x64 build 26200
GDI object count: 102
User object count: 47
Plugin list: 

Exception: System.NullReferenceException
Message: Object reference not set to an instance of an object.

Stack:
   at Greenshot.Editor.Forms.ImageEditorForm.get_CaptureDetails()
   at Greenshot.Editor.Destinations.EditorDestination.get_Description()
   at Greenshot.Base.Core.AbstractDestination.<>c__DisplayClass30_0.<GetMenuItem>b__0(Object <p0>, EventArgs <p1>)
   at System.Windows.Forms.ToolStripDropDownItem.OnDropDownShow(EventArgs e)
   at System.Windows.Forms.ToolStripDropDownItem.ShowDropDownInternal()
   at System.Windows.Forms.ToolStripMenuItem.OnMouseButtonStateChange(MouseEventArgs e, Boolean isMouseDown)
   at System.Windows.Forms.ToolStripItem.HandleMouseDown(MouseEventArgs e)
   at System.Windows.Forms.ToolStrip.OnMouseDown(MouseEventArgs mea)
   at System.Windows.Forms.Control.WmMouseDown(Message& m, MouseButtons button, Int32 clicks)
   at System.Windows.Forms.Control.WndProc(Message& m)
   at System.Windows.Forms.ToolStrip.WndProc(Message& m)
   at System.Windows.Forms.ToolStripDropDown.WndProc(Message& m)
   at System.Windows.Forms.NativeWindow.Callback(IntPtr hWnd, Int32 msg, IntPtr wparam, IntPtr lparam)

Configuration dump:

<img width="674" height="212" alt="Image" src="https://github.com/user-attachments/assets/b24a5ddd-3a93-4bbc-8a09-d04bc994968c" />

Windows 11 25H2

---
**Stack Trace Hash:** [2d75921ad74b]

Lakritzator:
I'm wondering if we should make debug builds for the unrelated versions, saves time looking for the error in the code

PR #1022, opened 2026-03-02
Guard null CaptureDetails in EditorDestination

danrhodes:
Fixes a NullReferenceException crash in EditorDestination.Description that occurred when hovering the "Send to Editor" menu item after capturing a Windows Toast notification. When an editor is opened via the blank editor path or the ReuseEditor path, its Surface.CaptureDetails is never populated from a capture and remains null. The original code only guarded against Title being null (with ?.) but assumed CaptureDetails itself was always present — causing the crash. The fix extracts CaptureDetails?.Title into a local variable first, safely short-circuiting to null if CaptureDetails is absent, and reuses that local for the Substring call to avoid the double property access that was there before.

Fixes - https://github.com/greenshot/greenshot/issues/1021

### DIFF
--- a/src/Greenshot.Base/Core/CoreConfiguration.cs
+++ b/src/Greenshot.Base/Core/CoreConfiguration.cs
@@ -343,7 +343,7 @@ namespace Greenshot.Base.Core
                     {
                         newSize.Height = 16;
                     }
-                    else if (IconSize.Height > 256)
+                    else if (newSize.Height > 256)
                     {
                         newSize.Height = 256;
                     }
@@ -353,7 +353,7 @@ namespace Greenshot.Base.Core
 
                 if (_iconSize != newSize)
                 {
-                    _iconSize = value;
+                    _iconSize = newSize;
                     PropertyChanged?.Invoke(this, new PropertyChangedEventArgs("IconSize"));
                 }
             }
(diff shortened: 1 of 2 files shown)


Lakritzator:
Hmmm, feels like it's only solving the symptom, why didn't I pass a capture details for that? There should be some meta data even in that case, which might be good to have later when I add the history.

danrhodes:
Ha, fair point — on reflection I had a look again and you're absolutely right, the symptom fix was the wrong approach. The actual fix is a single line in the base Surface() constructor ([Surface.cs:492](vscode-webview://0fav3fnffb3cg8flvuls3d9best9s1cefs12gvg5k87ol3brc37l/src/Greenshot.Editor/Drawing/Surface.cs#L492)) — initialising CaptureDetails = new CaptureDetails() so it is never null regardless of which constructor path is used. Surface(ICapture) overwrites it with the real capture metadata as it always did, so nothing changes for normal captures. The blank/reused editor case now always has a CaptureDetails with at minimum a timestamp, which should also help with the history feature you mentioned. EditorDestination has been restored to its original form. On that note I'm off to bed — it's been a looooong day! 😅

Lakritzator:
Good night, I know how that is, had a hard weekend studying for my ArchiMate exam which I had today...

danrhodes:
I hope you passed - good luck! 🤞

Lakritzator:
Yes yes, I have my archimate foundation certification.
Not sure if I am going to ever use it for Greenshot😂

danrhodes:
Not sure I'll ever repeat my day of recovering 4tb of corrupted data from a server on the greenshot project either, hopefully 🤣🤷‍♂️

Issue #1015, opened 2026-02-27
Weird filename issues when using destination picker and dated folders


danrhodes:
When using the destination picker -> open in editor -> Save as, I get the below, as the folder is not auto-created as it would be when auto saving to a folder via Save As 

My filename is ${YYYY}-${MM}-${DD}\${NUM} - ${title}

<img width="762" height="95" alt="Image" src="https://github.com/user-attachments/assets/092dbe83-5b2d-404f-8e6e-fbbfdcbc7bc8" />

<img width="732" height="251" alt="Image" src="https://github.com/user-attachments/assets/29e6e695-0dd0-4d09-8412-85098dde91db" />

<img width="543" height="317" alt="Image" src="https://github.com/user-attachments/assets/74e9e48e-93aa-4cb4-8b66-6d4a170e02a2" />

<img width="722" height="254" alt="Image" src="https://github.com/user-attachments/assets/de0d71da-a84a-487b-ae23-6a4d7f48d92e" />

PR #1023, opened 2026-03-02
Auto-create initial directory for SaveFileDialog

danrhodes:
Fixes a "Path does not exist" error shown by the native Windows Save dialog when the filename pattern includes a date-based subdirectory component

(e.g. ${YYYY}-${MM}-${DD}\${NUM} - ${title}).

Note - not sure why but the \ isn't showing in the example path above it is after the ${DD} it shows in edit description 🤷‍♂️

The pattern expands to a relative path like 2026-03-02\000002 - title, but ApplySuggestedValues was passing the whole string as the filename, leaving the dialog unaware that the 2026-03-02\ portion represented a subdirectory — and since that directory doesn't exist yet on the first save of a given day, the native SaveFileDialog rejected it before the save could proceed.

This only affects the dialog save path (destination picker → Save As, or destination picker → open in editor → Save As); the standard auto-save path works correctly because ImageIO.Save() always calls Directory.CreateDirectory() before writing.

Trickier than it first appeared and took considerable testing to get right — an early attempt fixed the wrong layer (Init() rather than ApplySuggestedValues()), and a second subtle issue caused the date subdirectory to be combined twice on subsequent saves because InitialDirectory was used as the base rather than conf.OutputFilePath, resulting in paths like 2026-03-02\2026-03-02\000018.

The final fix splits the expanded pattern into its directory and filename components, anchors the directory part against conf.OutputFilePath (the configured base output path) to avoid the double-combine, creates the subdirectory if it doesn't yet exist, and sets InitialDirectory and FileName separately.

The newly created directory is stored in the existing _eagerlyCreatedDirectory field so CleanUp() automatically removes it if the user cancels the dialog and the folder is empty — a safety mechanism already present in the code for exactly this purpose.

Fixes https://github.com/greenshot/greenshot/issues/1015

### DIFF
--- a/src/Greenshot.Base/Core/CoreConfiguration.cs
+++ b/src/Greenshot.Base/Core/CoreConfiguration.cs
@@ -343,7 +343,7 @@ namespace Greenshot.Base.Core
                     {
                         newSize.Height = 16;
                     }
-                    else if (IconSize.Height > 256)
+                    else if (newSize.Height > 256)
                     {
                         newSize.Height = 256;
                     }
@@ -353,7 +353,7 @@ namespace Greenshot.Base.Core
 
                 if (_iconSize != newSize)
                 {
-                    _iconSize = value;
+                    _iconSize = newSize;
                     PropertyChanged?.Invoke(this, new PropertyChangedEventArgs("IconSize"));
                 }
             }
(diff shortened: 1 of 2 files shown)


Lakritzator:
🤯

PR #1024, opened 2026-03-02
[Fix] Use action SaveToStream as filter in TrySaveToStream()

Christian-Schulz:
I've stumbled across this so many times. Now I turn it into a PR.

This change has no effect on many FileFormatHandlers because `LoadFromStream` and `SaveToStream` mostly contain the same file extensions.


Lakritzator (approved):

Lakritzator:
I guess I really need glasses, good catch.

PR #1027, opened 2026-03-03
Add safety checks, locking, and sync fixes

danrhodes:
These fixes have been sitting in my notes for a while — a mixture of things spotted during general code sessions, digging through edge cases while working on other fixes, and a few that came up from just reading through areas of the codebase I don't normally touch day to day.

The ScaleHelper one in particular had been nagging at me for some time, knowing the centred scale behaviour felt off but not getting around to properly tracing through the geometry until now. Figured it was time to get them all tidied up and submitted rather than letting them sit any longer.

Full breakdown below, none of these fix actual reported issue, but do improve the code quality somewhat. A full review on this please if possible to ensure all is ship shape 👌🤞

**MainForm.cs — OneDrive settings null guard**
Added explicit null check on FirstOrDefault() before passing to File.Exists(), and replaced the .Skip(1).Take(1).First() chain with a bounds-checked array index.

**ImageHelper.cs — Unsafe as cast dereference**
Null-checked the as Bitmap result before calling .Clone() so a non-Bitmap image type throws a meaningful exception instead of a silent NullReferenceException.

**ImageHelper.cs — EXIF bounds check**
Added a Value == null || Value.Length == 0 guard before indexing EXIF orientation data to prevent IndexOutOfRangeException on malformed images.

**Confluence.cs — Async deadlock risk**
Replaced all .Wait() and .Result blocking calls (7 occurrences) with .GetAwaiter().GetResult() to eliminate SynchronizationContext deadlock potential on the WinForms UI thread.

**NetworkHelper.cs** — Stream position not reset on retry. Added memoryStream.Seek(0, SeekOrigin.Begin) before the fallback re-read in both catch blocks so the stream isn't read from mid-position after a failed image load attempt.

**WordExporter.cs — Blanket exception swallowing**
Split the catch-all catch (Exception) into a targeted catch (COMException) for the expected "Word not running" case, with an explicit LOG. Warn catch for anything unexpected.

**Fraction.cs — Integer overflow in CompareTo**
Cross-multiplied comparison now casts to long before multiplying to prevent silent uint overflow wrapping that would invert sort order for large fraction values.

**ImageEditorForm.cs — Unsynchronised static list**
Added a _editorListLock object and wrapped all three EditorList accesses (Add, Remove, Sort) in lock blocks to prevent list corruption when multiple editor windows open or close concurrently.

**ScaleHelper.cs** — Centred scale completely non-functional. This was a really interesting one to work on — the centred scale block was computing result.Width - result.Width and result.Height - result.Height, both of which are always zero, meaning Inflate was always called with no-op deltas and Ctrl+resize centring had silently never worked. The fix computes the diff between the scaled result and the original boundsBeforeResize, divides by 2, and passes that to Inflate — which grows each edge outward symmetrically, giving you the correct anchor-at-centre behaviour.

### DIFF
--- a/src/Greenshot.Base/Core/Fraction.cs
+++ b/src/Greenshot.Base/Core/Fraction.cs
@@ -82,7 +82,7 @@ namespace Greenshot.Base.Core
         }
 
         public int CompareTo(Fraction other)
-            => (int) (Numerator * other.Denominator) - (int) (other.Numerator * Denominator);
+            => ((long) Numerator * other.Denominator).CompareTo((long) other.Numerator * Denominator);
 
         #endregion
 
--- a/src/Greenshot.Base/Core/ImageHelper.cs
+++ b/src/Greenshot.Base/Core/ImageHelper.cs
@@ -74,6 +74,10 @@ namespace Greenshot.Base.Core
                 }
 
                 PropertyItem item = image.GetPropertyItem(ExifOrientationId);
+                if (item.Value == null || item.Value.Length == 0)
+                {
+                    return;
+                }
 
                 ExifOrientations orientation = (ExifOrientations) item.Value[0];
                 // Orient the image.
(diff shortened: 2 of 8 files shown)


Lakritzator:
Impressive, there is one thing of which I wasn't sure if it would change the behavior, I think it was the ImageHelper, but it doesn't matter in that on fictional case.

PR #1028, opened 2026-03-03
Fix typos

mwtoews:
This fixes a handful of typos, identified and corrected with codespell.


src/Greenshot/Helpers/MailHelper.cs
```diff
@@ -402,7 +402,7 @@ internal class MapiHelperInterop
             /// </summary>
             private MapiHelperInterop()
             {
-                // Intenationally blank
+                // Internationally blank
```
jklingen:
```suggestion
                // Intentionally blank
```

Issue #1037, opened 2026-03-07
Adding a ExternalCommand entry that has a duplicated name gives an exception


alexlehm:
### Please confirm the following before submitting a Bug Report

- [x] I have searched existing open and closed issues and did not find a duplicate

### Greenshot version

1.3.132

### Describe the bug

I added a ExternalCommand entry where I accidentally wrote the same name as one I had earlier, this gives an Exception message instead of a concise error message.

Software version: 1.3.312 - 5a60aff83e (64 bit)
.NET runtime version: 4.0.30319.42000+
Time: 2026-03-07 13:13:19 +01:00
OS: Windows 10 2009 x64 build 19045
GDI object count: 168
User object count: 125

````Exception: System.ArgumentException
Message: An item with the same key has already been added.

Stack:
   at System.ThrowHelper.ThrowArgumentException(ExceptionResource resource)
   at System.Collections.Generic.Dictionary`2.Insert(TKey key, TValue value, Boolean add)
   at Greenshot.Plugin.ExternalCommand.SettingsFormDetail.ButtonOkClick(Object sender, EventArgs e)
   at System.Windows.Forms.Control.OnClick(EventArgs e)
   at System.Windows.Forms.Button.OnMouseUp(MouseEventArgs mevent)
   at System.Windows.Forms.Control.WmMouseUp(Message& m, MouseButtons button, Int32 clicks)
   at System.Windows.Forms.Control.WndProc(Message& m)
   at System.Windows.Forms.ButtonBase.WndProc(Message& m)
   at System.Windows.Forms.Button.WndProc(Message& m)
   at System.Windows.Forms.NativeWindow.Callback(IntPtr hWnd, Int32 msg, IntPtr wparam, IntPtr lparam)

Configuration dump:
````

This is mostly cosmetic, but a error message like duplicate title would be nice.

---
**Stack Trace Hash:** [9e54e81d2aa7]

PR #1045, opened 2026-03-10
Update SettingsFormDetail.cs

danrhodes:
Two bugs fixed, one UX improvement added:

**Bug fix (edit case)**: The original duplicate check at line 147 only ran when _commando == null (adding a new entry). When editing an existing entry and renaming it to a duplicate name, no check was done — leading to the unhandled ArgumentException. The new check always runs but excludes the entry's own original name (textBox_name.Text != _commando).

**UX improvement (tooltip)**: Instead of a popup dialog, hovering over the red name field now shows a tooltip: "A command with this name already exists. Please choose a different name." This is non-intrusive and consistent with the existing pattern (red background = something wrong). The tooltip clears when the name is valid again.

**Minor**: Changed Color.Red → Color.LightCoral for the duplicate name highlight, to be less alarming than the hard error red used for missing/invalid command paths.

000034 - Greenshot - Report20260310-1932 diagsession - Microsoft Visual Studio

Fixes https://github.com/greenshot/greenshot/issues/1037

### DIFF
--- a/src/Greenshot.Base/Core/ImageIO.cs
+++ b/src/Greenshot.Base/Core/ImageIO.cs
@@ -430,7 +430,7 @@ namespace Greenshot.Base.Core
                     returnValue = fileNameWithExtension;
                     IniConfig.Save();
                 }
-                catch (Exception e) when (e is ExternalException || e is IOException)
+                catch (Exception e) when (e is ExternalException || e is IOException || e is UnauthorizedAccessException)
                 {
                     MessageBox.Show(Language.GetFormattedString("error_nowriteaccess", saveImageFileDialog.FileName).Replace(@"\\", @"\"), Language.GetString("error"));
                 }
--- a/src/Greenshot.Base/Core/PluginUtils.cs
+++ b/src/Greenshot.Base/Core/PluginUtils.cs
@@ -170,6 +170,21 @@ namespace Greenshot.Base.Core
                 Log.Error("error retrieving icon: ", exIcon);
             }
 
+            // Fallback: use the Windows shell-associated icon (handles exes with no embedded icon, e.g. Windows curl.exe)
+            try
+            {
+                var shellIcon = Icon.ExtractAssociatedIcon(path);
+                if (shellIcon != null)
+                {
+                    Log.DebugFormat("Loaded shell icon for {0}", path);
+                    return shellIcon.ToBitmap();
+                }
+            }
+            catch (Exception exShell)
+            {
+                Log.Warn("error retrieving shell icon: ", exShell);
+            }
+
             return null;
         }
 


src/Greenshot.Plugin.ExternalCommand/SettingsFormDetail.cs
```diff
@@ -144,10 +145,18 @@ private void OkButtonState()
...
-        if (_commando == null && !string.IsNullOrEmpty(textBox_name.Text) && ExternalCommandConfig.Commands.Contains(textBox_name.Text))
+        bool isDuplicate = !string.IsNullOrEmpty(textBox_name.Text) &&
+            ExternalCommandConfig.Commands.Contains(textBox_name.Text) &&
+            textBox_name.Text != _commando;
+        if (isDuplicate)
         {
             buttonOk.Enabled = false;
-            textBox_name.BackColor = Color.Red;
+            textBox_name.BackColor = Color.LightCoral;
+            _nameToolTip.SetToolTip(textBox_name, "A command with this name already exists. Please choose a different name.");
```
Lakritzator: We would need a translation for this, add it to the English file and I believe our translation agent will do the rest. 

Usage is like the line here: https://github.com/greenshot/greenshot/blob/main/src/Greenshot.Plugin.ExternalCommand/ExternalCommandPlugin.cs#L192

Lakritzator:
Looks pretty good, but could you add the translation entry and code?

danrhodes:
> Looks pretty good, but could you add the translation entry and code?

Argh, yes, my bad 🙈 I will do this tomorrow!

Thanks for the catch! 👍

src/Greenshot.Plugin.ExternalCommand/SettingsFormDetail.cs
danrhodes: Leave it with me 👍

danrhodes:
Done 👍

Lakritzator:
@jklingen what did we do wrong, or rather forget, as the language agent didn't trigger?
https://github.com/greenshot/greenshot/tree/main/src/Greenshot.Plugin.ExternalCommand/Languages

jklingen:
@Lakritzator nothing done wrong - the custom I18N agent only does their job if Copilot is involved. AFAIK there is no straightforward way (yet) to have an agent take action automatically (e.g. in resonse to a PR event).
I think for human changes to UI messages, we currently need to explicitly ask Copilot to take care of them.

Thought: if we activated automated Copilot reviews on every PR, it would probably notice the language update and at least leave a comment (if not we could a explicit instructions for them to do so), but the reviewing copilot also cannot initiate updates to the branch.

danrhodes:
Can we ask it retrospectively on this pr to create a new or for the language update?

jklingen:
@danrhodes yes, a single sentence is enough ☺️ see https://github.com/greenshot/greenshot/tasks/194bd054-2251-4d9e-8668-b9c88b9e308b?author=*

I did a reverse check for a small sample of the translations, looks good (as far as I can tell).

danrhodes:
> @danrhodes yes, a single sentence is enough ☺️ see https://github.com/greenshot/greenshot/tasks/194bd054-2251-4d9e-8668-b9c88b9e308b?author=*
> 
> 
> 
> I did a reverse check for a small sample of the translations, looks good (as far as I can tell).
> 
> 

HTTP 404

jklingen:
> HTTP 404

oh good to know - I assumed the agent sessions in a public project were public as well. Obviously not...

I just needed to ask Copilot to: "Have a look at the changes done in #1045 and take care of the translations." But it seems that it needs a repository owner to do this.

danrhodes:
Makes sense I suppose. 👍

Issue #1036, opened 2026-03-07
Curl.exe does not show up in the tool bar when it does not have a Windows icon


alexlehm:
### Please confirm the following before submitting a Bug Report

- [x] I have searched existing open and closed issues and did not find a duplicate

### Greenshot version

1.3.312

### Describe the bug

I am using a few External Commands plugin entires to upload images to nullpointer-type sites and use curl.exe for that. When using the standard curl from Windows 10, the application has no Windows icon, this means that the entry does not show up in the tool bar, however when I use a curl.exe from another open source build that has an icon, it shows up. The same happens when using bash.exe which is either from Cygwin or Git-bash (the latter one has an icon). The same entries work when calling them from the list of plugins in the menu.

### Additional context

Define a ExternalCommand plugin entry that uses curl.exe to upload a file somewhere, e.g. this:

<img width="362" height="204" alt="Image" src="https://github.com/user-attachments/assets/41e47c6a-ad44-43a3-a74b-1fb390737895" />

then define the same entry with a path to curl.exe which has an icon, e.g.

<img width="362" height="204" alt="Image" src="https://github.com/user-attachments/assets/3b4de9cc-30e6-4a24-bee5-7fb8d8ad1d6e" />

The tool bar looks like this, the system curl.exe entry is missing (also an entry I use which uses bash.exe)

<img width="389" height="113" alt="Image" src="https://github.com/user-attachments/assets/3b47e0dc-b181-4175-a7d1-41b03cc5bacf" />

alexlehm:
I should say that this worked a while ago but I cannot exactly say when it broke, likely with the most recent update.

danrhodes:
For clarity you would assume it would use the icon of the exe file?

alexlehm:
Yes, e.g. when I use cmd.exe to run the specific command, it shows the cmd icon that looks like this:

<img width="57" height="50" alt="Image" src="https://github.com/user-attachments/assets/8e062c1b-dd66-43ea-b45d-107496144df7" />

danrhodes:
Hmmmm this really is a puzzle!

PR #1046, opened 2026-03-10
Update PluginUtils.cs to fix icons

danrhodes:
Fixes an issue where External Command plugin entries using executables with no embedded icon resource (such as the Windows built-in curl.exe, tar.exe, and ssh.exe) were silently excluded from the editor toolbar. The toolbar's AddDestinations() method skips any destination whose DisplayIcon is null, and PluginUtils.GetExeIcon() was only using Dapplo.Windows.Icons.IconHelper.ExtractAssociatedIcon, which returns null for executables with no embedded icon. A fallback to Icon.ExtractAssociatedIcon() has been added, which goes through the Windows shell and always returns an icon (typically the generic console application icon for iconless executables).

image

Executables that already have an embedded icon are unaffected — the fallback is only reached when IconHelper returns null. The affected entries were still visible in the context menu since that code path has no null icon guard, making the omission from the toolbar inconsistent and confusing.

Tested with various random icons

image

image

Fixes https://github.com/greenshot/greenshot/issues/1036

### DIFF
--- a/src/Greenshot.Base/Core/PluginUtils.cs
+++ b/src/Greenshot.Base/Core/PluginUtils.cs
@@ -170,6 +170,21 @@ namespace Greenshot.Base.Core
                 Log.Error("error retrieving icon: ", exIcon);
             }
 
+            // Fallback: use the Windows shell-associated icon (handles exes with no embedded icon, e.g. Windows curl.exe)
+            try
+            {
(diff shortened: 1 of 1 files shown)


Lakritzator:
Good idea for a stupid bug 👍

Issue #910, opened 2026-02-03
In window 11 Screen Capture ( a selection) show this error for Word 365


grojas123:
**Describe the bug**
Software version: 1.3.312 - 5a60aff83e (64 bit)
.NET runtime version: 4.0.30319.42000+
Time: 2026-02-03 15:12:37 -06:00
OS: Windows 10 2009 x64 build 26200
GDI object count: 81
User object count: 37

Exception: System.InvalidCastException
Message: Unable to cast COM object of type 'Microsoft.Office.Interop.Word.ApplicationClass' to interface type 'Microsoft.Office.Interop.Word._Application'. This operation failed because the QueryInterface call on the COM component for the interface with IID '{00020970-0000-0000-C000-000000000046}' failed due to the following error: Element not found. (Exception from HRESULT: 0x8002802B (TYPE_E_ELEMENTNOTFOUND)).

Stack:
   at System.StubHelpers.StubHelpers.GetCOMIPFromRCW(Object objSrc, IntPtr pCPCMD, IntPtr& ppTarget, Boolean& pfNeedsRelease)
   at Microsoft.Office.Interop.Word.ApplicationClass.get_Version()
   at Greenshot.Plugin.Office.OfficeExport.WordExporter.InitializeVariables(IDisposableCom`1 wordApplication)
   at Greenshot.Plugin.Office.OfficeExport.WordExporter.GetWordApplication()
   at Greenshot.Plugin.Office.OfficeExport.WordExporter.<GetWordDocuments>d__6.MoveNext()
   at System.Collections.Generic.List`1..ctor(IEnumerable`1 collection)
   at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
   at Greenshot.Plugin.Office.Destinations.WordDestination.ExportCapture(Boolean manuallyInitiated, ISurface surface, ICaptureDetails captureDetails)
   at Greenshot.Helpers.CaptureHelper.HandleCapture()
   at Greenshot.Helpers.CaptureHelper.CaptureWithFeedback()
   at Greenshot.Helpers.CaptureHelper.MakeCapture()
   at Greenshot.Helpers.CaptureHelper.CaptureRegion(Boolean captureMouse)
   at Greenshot.Base.Controls.HotkeyControl.HandleMessages(Message& m)
   at Greenshot.Forms.MainForm.WndProc(Message& m)
[...]

PR #1049, opened 2026-03-11
Handle InitializeVariables failure, default to Office97

danrhodes:
Wrap InitializeVariables calls in try/catch to prevent an unhandled InvalidCastException when Word's COM type library is in a degraded state (TYPE_E_ELEMENTNOTFOUND), falling back to Word 97 compatibility instead of crashing the capture.

Saw this as an issue off the back of the last one re Excel...

Fixes https://github.com/greenshot/greenshot/issues/910


Lakritzator:
I'm not sure if it really solves the issue, com interop started to be a pain in the....
Anyway... again, it does improve the general code.

Issue #901, opened 2026-02-01
Improve the window position of the color picker


Dmdwt236:
**Describe the bug**
When annotating screenshots in Greenshot using highlights or rectangles, the color picker window opens at an arbitrary position on the screen instead of near the screenshot editor window. This behavior makes color selection unintuitive and inconvenient. 
Especially on ultra-wide monitors. 

**To Reproduce**
Take a screenshot with Greenshot.
Open the screenshot in the Greenshot image editor.
Select an annotation tool such as highlight or rectangle.
Click on the color picker to change the annotation color.
Observe where the color picker window appears.

**Expected behavior**
**The color picker should appear close to the screenshot editor window. Ideally, it should be integrated directly into the editor UI, for example as an embedded color palette similar to the one used in Microsoft Paint.**

**Screenshots**
xx

**Versions (please complete the following information):**
Greenshot version: **ALL** 
Windows version: Windows 10 + 11

danrhodes:
Hmmm having a think about this....

PR #1050, opened 2026-03-11
Update ColorDialog.cs

danrhodes:
Fixes the colour picker dialog appearing at an arbitrary screen position, which was particularly disruptive on ultra-wide monitors. The dialog now positions itself near the top-left of the editor window (just below the toolbar area) when opened, and is clamped to the working area of whichever screen the editor is on, ensuring it always appears close to the editor regardless of monitor configuration.

image

Fixes https://github.com/greenshot/greenshot/issues/901 hopefully

### DIFF
--- a/src/Greenshot.Editor/Forms/ColorDialog.cs
+++ b/src/Greenshot.Editor/Forms/ColorDialog.cs
@@ -293,38 +293,17 @@ namespace Greenshot.Editor.Forms
 
         public new DialogResult ShowDialog(IWin32Window owner)
         {
-            PositionNearOwner(owner);
-            return base.ShowDialog(owner);
-        }
-
-        private void PositionNearOwner(IWin32Window owner)
-        {
-            if (owner is not Control ownerControl) return;
-
-            // Find the top-level form
-            var ownerForm = ownerControl as Form ?? ownerControl.FindForm();
-            if (ownerForm == null) return;
-
-            // Place the dialog just below the top-left of the owner window (below the toolbar area)
-            var ownerBounds = ownerForm.Bounds;
-            int x = ownerBounds.Left + 8;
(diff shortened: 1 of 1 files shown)


Lakritzator:
Ah yes, I actually thought a couple of times about looking at this one, fully agree to it being annoying. 👍

danrhodes:
Agreed 👍🤣

Lakritzator:
My solution would have been to have the window appear near the mouse cursor... I'm not 100% sure if we regret the location in the future.

danrhodes:
Happy to change. Top left was just easy to make work.

I did consider bringing it directly into the GUI but then pushed the mad thoughts aside once I thought of available pixels space.

I can bring it up near the cursor but given the colour picker icon is generally too left???

danrhodes:
Ah do you mean where the cursor is on my screen shot? So it's not covering the toolbar?

I get that, however the background tool bar is not able to be clicked on when the colour picker is active so might aid in the reduction of users getting confused if the CP UI is front and centre?

As I say I'm happy t change 👍

Lakritzator:
I mean doing something like Dialog.Location = Mouse.Location (- width /2, - height / 2)
That it just opens pretty much under the mouse cursor, and we do not need to calculate the somewhat complex thing you are doing.

Lakritzator:
Btw... it really needs a redesign.... 😂

danrhodes:
I've gone with your suggestion :-)

image

Lakritzator:
I expected the code to be much simpler, why do we need that kind of calculation, need to check

danrhodes:
The calculation centers the dialog on the cursor while clamping it within the working area of the active screen. Without it, a simple Location = mouse position breaks near screen edges, on secondary monitors, and when the taskbar isn’t at the bottom — the extra lines are just bounds checks 👍

Lakritzator:
Yeah I saw you are right 🤷‍♂️

PR #1051, opened 2026-03-11
Update EditorDestination.cs

danrhodes:
While using the editor, I encountered a NullReferenceException crash that occurred when opening the Destinations picker menu. The crash originated in EditorDestination.get_Description, which accessed editor.CaptureDetails.Title without guarding against a null CaptureDetails. This can happen when the Destination picker is opened while another editor window is present but not yet fully initialized — DynamicDestinations() enumerates all open editors, including partially-loaded ones. The fix adds a null check on CaptureDetails so the description gracefully falls back to the base label instead of throwing.

Software version: 1.4.112 - acf799517f (64 bit)
Time: 2026-03-11 19:07:07 +00:00
OS: Windows 11 x64 build 26200
GDI object count: 109
User object count: 85
Plugin list: 

Exception: System.NullReferenceException
Message: Object reference not set to an instance of an object.

Stack:
   at Greenshot.Editor.Forms.ImageEditorForm.get_CaptureDetails()
   at Greenshot.Editor.Destinations.EditorDestination.get_Description()
   at Greenshot.Base.Core.AbstractDestination.<>c__DisplayClass30_0.<GetMenuItem>b__0(Object <p0>, EventArgs <p1>)
   at System.Windows.Forms.ToolStripDropDownItem.OnDropDownShow(EventArgs e)
   at System.Windows.Forms.ToolStripDropDownItem.ShowDropDownInternal()
   at System.Windows.Forms.ToolStripMenuItem.OnMouseButtonStateChange(MouseEventArgs e, Boolean isMouseDown)
   at System.Windows.Forms.ToolStripItem.HandleMouseDown(MouseEventArgs e)
   at System.Windows.Forms.ToolStrip.OnMouseDown(MouseEventArgs mea)
   at System.Windows.Forms.Control.WmMouseDown(Message& m, MouseButtons button, Int32 clicks)

Configuration dump:

### DIFF
--- a/src/Greenshot.Editor/Forms/ImageEditorForm.cs
+++ b/src/Greenshot.Editor/Forms/ImageEditorForm.cs
@@ -152,11 +152,6 @@ namespace Greenshot.Editor.Forms
             // Compute emojis in background
             EmojiData.Load();
 
-            lock (_editorListLock)
-            {
-                EditorList.Add(this);
-            }
-
             //
             // The InitializeComponent() call is required for Windows Forms designer support.
             //
@@ -191,6 +186,13 @@ namespace Greenshot.Editor.Forms
             Surface = surface;
             // Initial "saved" flag for asking if the image needs to be save
             _surface.Modified = !outputMade;
+
+            // Only register in the global editor list after the surface is fully assigned,
+            // so background threads iterating Editors never see a partially-initialized editor.
+            lock (_editorListLock)
+            {
+                EditorList.Add(this);
+            }
 
             UpdateUi();
 
@@ -240,6 +242,13 @@ namespace Greenshot.Editor.Forms
                 throw new ApplicationException("Surface modified");
             }
 
+            // Remove from the global list while _surface is being swapped, so background threads
+            // iterating Editors never observe this editor in a null-surface state.
+            lock (_editorListLock)
+            {
+                EditorList.Remove(this);
+            }
+
             RemoveSurface();
 
             panel1.Height = 10;
@@ -273,6 +282,12 @@ namespace Greenshot.Editor.Forms
                 {
                     Text = _surface.CaptureDetails.Title + " - " + Language.GetString(LangKey.editor_title);
                 }
+            }
+
+            // Re-register in the global list now that the new surface is fully assigned.
+            lock (_editorListLock)
+            {
+                EditorList.Add(this);
             }
 
             Activate();


Lakritzator:
I am not sure what "Send to" you mean, and how you even managed to do something while the window is not yet initialized?

danrhodes:
Apologies, I meant "Destination Picker", it's often wrongly referred to as "sent to" at work. 


I had a screenshot open in editor, took another screenshot and opened that in Editor and it threw that exception. From memory, as I only took the error into notepad++ the screenshot had no title as it was a screenshot of the Greenshot settings so unsure why it had no title 🤔

Lakritzator:
At work I try to learn my colleagues to think about the cause, not the symptom, solving something with a null check, hides the fact that the thing shouldn't be null... I was just curious why this happens, and fixing it might be okay but it might hide a deeper issue. 

This is what Gemini-cli told me after ~5 Minutes:

✦ The NullReferenceException in EditorDestination.Description when accessing editor.CaptureDetails.Title is primarily caused by a race condition and unsafe exposure of partially initialized ImageEditorForm instances in a global static list.

  Analysis of the Cause


   1. Premature Global Exposure:
      In ImageEditorForm.cs, the constructor calls Initialize(ISurface surface, bool outputMade). At the very beginning of Initialize, the form adds itself to the static EditorList:
   1     // src\Greenshot.Editor\Forms\ImageEditorForm.cs
   2     lock (_editorListLock)
   3     {
   4         EditorList.Add(this);
   5     }
      At this point, the _surface field of the form is still null. It isn't assigned until much later in the same method (Surface = surface;).


   2. Concurrency via Background Threads:
      Greenshot frequently calls DestinationHelper.GetAllDestinations() from background threads (e.g., when a new capture is completed or when building UI menus). GetAllDestinations() performs a global sort on all active destinations by their Description:


   1     // src\Greenshot.Base\Core\DestinationHelper.cs
   2     public static IEnumerable<IDestination> GetAllDestinations() {
   3         return ... .OrderBy(p => p.Priority).ThenBy(p => p.Description);
   4     }


   3. Triggering the NRE:
      The EditorDestination.DynamicDestinations() generator iterates over the global ImageEditorForm.Editors list. If an editor is currently being initialized (or having its surface replaced during a "reuse editor" operation), it is already in the list but its _surface is null.
      When GetAllDestinations() calls Description on the corresponding EditorDestination:


   1     // src\Greenshot.Editor\Destinations\EditorDestination.cs
   2     return ... + editor.CaptureDetails.Title?.Substring(0, Math.Min(20, editor.CaptureDetails.Title.Length));
      The editor.CaptureDetails property in ImageEditorForm is implemented as:


   1     public ICaptureDetails CaptureDetails => _surface.CaptureDetails;
      If _surface is null, accessing editor.CaptureDetails throws the NullReferenceException. Even if editor.CaptureDetails were made null-safe (e.g., _surface?.CaptureDetails), the subsequent access to .Title would still throw if the result was null.

  Why "shouldn't all editors have a non-null CaptureDetails?"


  You are correct that they should. However, the current architecture has two main weaknesses that break this assumption:
   * Initialization Window: There is a significant gap between when an editor becomes "visible" to the rest of the system (via EditorList) and when its mandatory data (the Surface and CaptureDetails) is actually attached.
   * Surface Lifecycle: In SetSurface, the current surface is disposed and set to null before a new one is assigned. Because these operations are not synchronized with the _editorListLock, background threads can "catch" the editor in this intermediate null state.

  Conclusion


  The NRE isn't just a missing null check in the Description property; it's a symptom of exposing the ImageEditorForm to global destination managers before it is fully ready. While adding a null check in EditorDestination would stop the crash, the underlying issue is that Editors
  can return instances that are not yet functional.

Lakritzator:
It could just have said, wow that is crappy code, why do you even expect it to work? 🤷‍♂️

danrhodes:
I'll review you Gemini suggestion more thoroughly later and update the pr 👍

danrhodes:
Fixed the actual crime, not just the evidence. Editors now join EditorList only once fully initialized, and temporarily leave during surface replacement.

Over to you @Lakritzator :-)

Lakritzator:
Like a true detective 🫆

Issue #1060, opened 2026-03-13
Destination picker issues showing multiple editors


danrhodes:
### Please confirm the following before submitting a Bug Report

- [x] I have searched existing open and closed issues and did not find a duplicate

### Greenshot version

1.4.138

### Describe the bug

<img width="820" height="302" alt="Image" src="https://github.com/user-attachments/assets/b4cd1974-19b7-413d-91b1-de7a597fe3ea" />

<img width="778" height="215" alt="Image" src="https://github.com/user-attachments/assets/2e49aacd-41f9-4ec6-892e-927d37408593" />

Each time i open a screenshot in editor it appears in this list until i close and re-open GS.

danrhodes:
Interestingly, there is no title on the editor it opens.

<img width="503" height="111" alt="Image" src="https://github.com/user-attachments/assets/762854e4-7f5e-4b53-952f-77e18eb70dfb" />

Yet the title shows in the editor drop down which shouldnt be there.

<img width="816" height="292" alt="Image" src="https://github.com/user-attachments/assets/b147b838-cfa9-401d-9934-b04ae94ba8cb" />

The issue did not exist in 1.4.134 the previous release.

Lakritzator:
I’m not seeing anything out of the ordinary, it’s made so you can add an screenshot to an existing

danrhodes:
None of these editors are open though and when you click one it does nothing? Surely it should only drop down to add to an existing editor of one is open?

Also when you click to just open in editor the first level option there's no file title, yet when you click save the title is there.

Lakritzator:
Might this have to do with #1051

danrhodes:
> Might this have to do with #1051 

I suspect so yes. I am about to fix  🙈

danrhodes:
Ok, this looks to be fixed!

Titles fixed - 
![Image](https://github.com/user-attachments/assets/ad168962-6135-4264-8a41-9aabcd78b3d2)

Only open editors show in the dropdown, and no editors that have been closed
![Image](https://github.com/user-attachments/assets/c13d4ebf-4539-415c-89a7-bf8c7a737927)

I've also made a small adjustment to the drop-down, which I hope you are ok with. Instead of opening in the image editor, it now says Add to image editor
![Image](https://github.com/user-attachments/assets/e4b2cad6-935f-46d8-a7b2-b94f88161f5b)

I will push shortly.

PR #1062, opened 2026-03-13
Fixing my mistakes

danrhodes:
This fix addresses two bugs introduced with #1051 (not #1027 that I also suspected) that affected the image editor.

The first was a double-registration issue in ImageEditorForm.Initialize() — when a new editor opened, it was being added to the global EditorList twice (once by SetSurface() and again explicitly in Initialize()). Since List.Remove() only removes the first occurrence; closing an editor left a ghost entry behind, causing the "Open in image editor" destination picker submenu to accumulate stale closed-editor entries indefinitely.

The second bug was that UpdateUi() → ApplyLanguage() was overwriting the editor window title (set correctly by SetSurface()) with just the bare "Greenshot editor" string on every new editor open — this has been fixed by re-applying the capture title after UpdateUi() completes.

Additionally, the submenu entries targeting a specific open editor have been renamed from "Open in image editor" to "Add to editor....." for clarity, since selecting one of those entries adds the new screenshot as an image object into an already-open editor rather than opening a fresh one.

Apologies for the disruption — both root causes are now properly resolved! 🙈

Fixes https://github.com/greenshot/greenshot/issues/1060

### DIFF
--- a/src/Greenshot.Editor/Configuration/LanguageKeys.cs
+++ b/src/Greenshot.Editor/Configuration/LanguageKeys.cs
@@ -61,6 +61,7 @@ namespace Greenshot.Editor.Configuration
         settings_destination,
         settings_destination_clipboard,
         settings_destination_editor,
+        settings_destination_editor_add,
         settings_destination_fileas,
         settings_destination_printer,
         settings_destination_picker,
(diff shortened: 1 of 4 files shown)


danrhodes:
Update: After going back over my fixes, I can confirm the bugs addressed here were introduced by #1051, not #1027 as I originally stated in the pr. The restructuring of ImageEditorForm.cs in #1051 caused both the double-registration and the title overwrite — apologies for the incorrect attribution.​​​​​​​​​​​​​​​​

Lakritzator:
No worries, that's why we dog feed.
I expect a few bugs in my own changes too, the more we touch the higher the risk 🤷‍♀️

Issue #1072, opened 2026-03-16
tried to Copy an area that had an image


David98230:
### Please confirm the following before submitting a Bug Report

- [x] I have searched existing open and closed issues and did not find a duplicate

### Greenshot version

1.3.312

### Describe the bug

Software version: 1.3.312 - 5a60aff83e (64 bit)
.NET runtime version: 4.0.30319.42000+
Time: 2026-03-15 18:27:38 -07:00
OS: Windows 10 2009 x64 build 26200
GDI object count: 84
User object count: 42

Exception: System.NullReferenceException
Message: Object reference not set to an instance of an object.

Stack:
   at Greenshot.Plugin.Office.OfficeExport.Entities.OneNotePage.get_DisplayName()
   at Greenshot.Base.Core.AbstractDestination.<>c__DisplayClass30_0.<GetMenuItem>b__0(Object <p0>, EventArgs <p1>)
   at System.Windows.Forms.ToolStripDropDownItem.OnDropDownShow(EventArgs e)
   at System.Windows.Forms.ToolStripDropDownItem.ShowDropDownInternal()
   at System.Windows.Forms.ToolStripMenuItem.OnMenuAutoExpand()
   at System.Windows.Forms.MenuTimer.OnTick(Object sender, EventArgs e)
   at System.Windows.Forms.Timer.OnTick(EventArgs e)
   at System.Windows.Forms.Timer.TimerNativeWindow.WndProc(Message& m)
   at System.Windows.Forms.NativeWindow.Callback(IntPtr hWnd, Int32 msg, IntPtr wparam, IntPtr lparam)

Configuration dump:

### Additional context

I clicked the box that said, I checked for duplicates, but I didn't. Sorry.  It also froze the menu on the screen when it crashed

David98230:
Thank You.

PR #1076, opened 2026-03-17
Update OneNoteExporter.cs issues/1072

danrhodes:
Pages were being added to the list without verifying that their parent hierarchy was fully populated. If a page's section or notebook was null, DisplayName would crash accessing Parent.Parent.Name when the menu expanded. The fix skips any pages with an incomplete parent chain during parsing.

Fixes https://github.com/greenshot/greenshot/issues/1072


danrhodes:
Waaaaayy this is new 👆

Lakritzator:
Do you mean the Gemini reviews? Yeah, I enabled it for something else, but I didn't find the fine-tuning settings 😂

Lakritzator (approved):

danrhodes:
> Do you mean the Gemini reviews? Yeah, I enabled it for something else, but I didn't find the fine-tuning settings 😂

It's not a bad thing in my opinion we run most change requests through AI as well as peer review at work its a safety net for human error.

Issue #1061, opened 2026-03-13
greenshot not opening windows microsoft office 2019 and error displayed


nirmalt2003:
### Please confirm the following before submitting a Bug Report

- [x] I have searched existing open and closed issues and did not find a duplicate

### Greenshot version

1.3.312

### Describe the bug

Software version: 1.3.312 - 5a60aff83e (64 bit)
.NET runtime version: 4.0.30319.42000+
Time: 2026-03-13 23:57:53 +05:30
OS: Windows 10 2009 x64 build 19045
GDI object count: 86
User object count: 38

Exception: System.InvalidCastException
Message: Unable to cast COM object of type 'Microsoft.Office.Interop.Word.ApplicationClass' to interface type 'Microsoft.Office.Interop.Word._Application'. This operation failed because the QueryInterface call on the COM component for the interface with IID '{00020970-0000-0000-C000-000000000046}' failed due to the following error: Error loading type library/DLL. (Exception from HRESULT: 0x80029C4A (TYPE_E_CANTLOADLIBRARY)).

Stack:
   at System.StubHelpers.StubHelpers.GetCOMIPFromRCW(Object objSrc, IntPtr pCPCMD, IntPtr& ppTarget, Boolean& pfNeedsRelease)
   at Microsoft.Office.Interop.Word.ApplicationClass.get_Version()
   at Greenshot.Plugin.Office.OfficeExport.WordExporter.InitializeVariables(IDisposableCom`1 wordApplication)
   at Greenshot.Plugin.Office.OfficeExport.WordExporter.GetWordApplication()
   at Greenshot.Plugin.Office.OfficeExport.WordExporter.<GetWordDocuments>d__6.MoveNext()
   at System.Collections.Generic.List`1..ctor(IEnumerable`1 collection)
   at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
   at Greenshot.Plugin.Office.Destinations.WordDestination.ExportCapture(Boolean manuallyInitiated, ISurface surface, ICaptureDetails captureDetails)
   at Greenshot.Helpers.CaptureHelper.HandleCapture()
   at Greenshot.Helpers.CaptureHelper.CaptureWithFeedback()
   at Greenshot.Helpers.CaptureHelper.MakeCapture()
   at Greenshot.Helpers.CaptureHelper.CaptureRegion(Boolean captureMouse)
   at Greenshot.Base.Controls.HotkeyControl.HandleMessages(Message& m)
[...]

PR #1077, opened 2026-03-17
Fixing a root cause not a symptom in Office

danrhodes:
The previous fix caught InvalidCastException in InitializeVariables and fell back to a default version — treating a broken COM object as if it were a usable one. The root cause is TYPE_E_CANTLOADLIBRARY: when the Office type library can't be loaded, every subsequent COM call on that object will fail the same way, not just .Version. The correct fix is to re-throw InvalidCastException from InitializeVariables and handle it in the callers (GetWordApplication, etc.) by disposing the broken COM object and returning null — so Greenshot treats the Office application as unavailable rather than attempting to use an object it fundamentally cannot interface with.

This came from and fixes https://github.com/greenshot/greenshot/issues/1061

### DIFF
--- a/src/Greenshot.Plugin.Office/OfficeExport/ExcelExporter.cs
+++ b/src/Greenshot.Plugin.Office/OfficeExport/ExcelExporter.cs
@@ -56,7 +56,16 @@ namespace Greenshot.Plugin.Office.OfficeExport
 
             if (excelApplication?.ComObject != null)
             {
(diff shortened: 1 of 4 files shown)


Lakritzator:
"This came from and fixes https://github.com/greenshot/greenshot/issues/1061"
I will merge it, but as far as I understood, it might fix the error message, it doesn't fix that the user wanted to export to word?

PR #1078, opened 2026-03-17
[Fix] Paste image from editor

Christian-Schulz:
This fixes 2 Issues.

1.  It was not possible to export the capture from the editor to clipboard and paste it in the editor back.

The DibFileFormatHandler now supports LoadDrawableFromStream again.
_(This was a 3h debug session for one line of code)_

2. 🙏Unfortunately, my commit from yesterday caused all pasted images to become the size 0x0.

Added `DrawableContainer.ResetToDefaultSize()`, which must be called whenever the Image is set. 

Applied the same behavior to IconContainer and CursorContainer.
Changing the Image, CaptureCursor, or Icon properties no longer affects the container size. 
All existing usages have been updated.

The context menu action "Reset Size" now uses ResetToDefaultSize() internally too.

### DIFF
--- a/src/Greenshot.Editor/Drawing/CursorContainer.cs
+++ b/src/Greenshot.Editor/Drawing/CursorContainer.cs
@@ -115,8 +115,6 @@ namespace Greenshot.Editor.Drawing
                 }
 
                 // Clone cursor (is this correct??)
-                Width = value.Size.Width;
-                Height = value.Size.Height;
                 cursor = value;
                 savedCursor = new CaptureCursorSerializationWrapper(value);
             }
@@ -144,6 +142,8 @@ namespace Greenshot.Editor.Drawing
 
         public void Load(string filename)
         {
+            throw new NotImplementedException("Loading cursor from file is not implemented yet.");
+
             if (!File.Exists(filename))
             {
                 return;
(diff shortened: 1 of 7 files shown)


src/Greenshot.Editor/Drawing/IconContainer.cs
```diff
@@ -73,8 +73,6 @@ public Icon Icon
             {
                 icon?.Dispose();
                 icon = (Icon) value.Clone();
-                Width = value.Width;
```
Lakritzator, line 76: Why don't we want the with and height from the icon?
Ah, I guess you moved the logic to the ResetToDefaultSize, which is called in the Surface.AddIconContainer.
This feels slightly off, I would be okay if we have "stupid" containers, which only contain the needed values and everything is managed from the outside. This does mean that the behavior changes, which for instance is important to realize for the CaptureBar plugin of PR 953
I also assume we deserialize the width/height anyway... so leave it...

Ah yes, due to the change, more knowledge was needed in the AbstractFileHandler...

src/Greenshot.Editor/Drawing/CursorContainer.cs
```diff
@@ -144,6 +142,8 @@ protected override void Dispose(bool disposing)
 
         public void Load(string filename)
         {
+            throw new NotImplementedException("Loading cursor from file is not implemented yet.");
```
Lakritzator, line 145: I guess we could also just remove the ICursorContainer.Load method...

src/Greenshot.Editor/Drawing/IconContainer.cs
Christian-Schulz: I think I'll add some more documentation to the porperties and a helper function `SetImageAndResetSize()`,`SetIconeAndResetSize()` ... .

Lakritzator: Probably not worth it for now, it would be better if we have behavior guidelines for the future implementations.

PR #1083, opened 2026-03-18
Fix uninstall old versions

Christian-Schulz:
@Lakritzator  I think i found the reason why old plugins (Win10, ...) are still present after an update.

It happens since v1.4.143.

With commit https://github.com/greenshot/greenshot/pull/1000/changes/57692e3b982f323dde53dce7bfefafdb328f6f35#diff-b1adc65b08ec5ee452d941e818386b35250dbeb90572ed5ac4f93f05f69914bc   the function  `UnInstallOldVersions()` was deleted in `setup.iss`.

Comment there:

> This fixes some of the issues in the installer, although Restart Manager was used, the installer checkedvia the Mutex if the application was running. 
> 
> Also it was running the uninstall before the application was closed. 
> 
> Fixed issues with triggering the TryClose from the wrong thread. Still having issues that the  ApplicationRestartManager.ListenForEndSession().Subscribe doesn't seem to work.

It looks like it was temporarily removed and then forgetten to be re-added.

⚠️I just copied the old function back and haven't tested it yet. 

I'm not very familiar with the installer, but I thought you should at least have this information

### DIFF
--- a/src/Greenshot-Installer/setup.iss
+++ b/src/Greenshot-Installer/setup.iss
@@ -597,6 +597,37 @@ begin
 	Result := asUninstallStrings;
 end;
 
+/////////////////////////////////////////////////////////////////////
+procedure UnInstallOldVersions();
+var
+	sUnInstallString: String;
+	index: Integer;
+	isUninstallMade: Boolean;
+	iResultCode : Integer;
+	asUninstallStrings : array of String;
+begin
+	isUninstallMade := false;
+	asUninstallStrings := GetUninstallStrings();
+	for index := 0 to (GetArrayLength(asUninstallStrings) -1) do
+	begin
+		sUnInstallString := RemoveQuotes(asUninstallStrings[index]);
+		if Exec(sUnInstallString, '/SILENT /NORESTART /SUPPRESSMSGBOXES','', SW_HIDE, ewWaitUntilTerminated, iResultCode) then
+			isUninstallMade := true;
+	end;
+
+	// Wait a few seconds to prevent installation issues, otherwise files are removed in one process while the other tries to link to them
+	if (isUninstallMade) then
+		Sleep(2000);
+end;
+
+/////////////////////////////////////////////////////////////////////
+procedure CurStepChanged(CurStep: TSetupStep);
+begin
+	if (CurStep=ssInstall) then
+	begin
+		UnInstallOldVersions();
+	end;
+end;
 
 /////////////////////////////////////////////////////////////////////
 // End of unstall code


Lakritzator:
I agree that there is some weird issue with the installation, but this shouldn't be it!

I think you are mixing causality and correlation, you correlate the fact that old plugins are staying behind on the system and that I removed the uninstall call from the pascal script in our Inno-setup.

Unfortunately the issue was already there in 1.3, and me remove the call to the uninstaller, is due to the fact that the uninstall is default behavior of Inno-setup, it would be a bad installer if this wasn't done automatically.

In fact, calling this manually is totally stupid and **might** actually BE the CAUSE of the leftovers... I'm only 30% sure, but it's a fact it shouldn't be called like that. I did find some hint which might be interesting, we use:

`Uninstallable=true`

But according to the documentation, it should be yes/no, or an expression which evaluates to true (which might be correct).
I'll play around with that... and do some more testing

Lakritzator:
I did a test:

Install Greenshot with the dropbox plugin.

What is interesting, if I run the uninstaller manually, the following message arrives:
image

Afterwards, the following files are left over:
Plugins\DropBox\Greenshot.Plugin.DropBox.dll
Dapplo.HttpExtensions.dll
Dapplo.HttpExtensions.JsonNet.dll
Dapplo.Log.dll
Dapplo.Windows.AppRestartManager.dll
Dapplo.Windows.Common.dll
Dapplo.Windows.Dpi.dll
Dapplo.Windows.Input.dll
Dapplo.Windows.Kernel32.dll
Dapplo.Windows.Messages.dll
Dapplo.Windows.User32.dll
Greenshot.exe
Greenshot.Base.dll
Greenshot.Editor.dll
HtmlAgilityPack.dll
log4net.dll
Microsoft.IO.RecyclableMemoryStream.dll
Microsoft.Toolkit.Uwp.Notifications.dll
Newtonsoft.Json.dll
System.Buffers.dll
System.CommandLine.dll
System.Reactive.dll
System.Resources.Extensions.dll

After installing a new version over it, these files stay.

Now, I had a hunch, and I asked Gemini, the following is what I got from Gemini:
"If Inno Setup skips installing a file because it is already present (and determined to be the same or newer), it does not "claim" ownership of that file. Consequently, when you run the uninstaller, Inno Setup will leave that file untouched."

Lakritzator:
So changing the install flags to "overwritereadonly ignoreversion" works much better, this means if the files are already there, the installer overwrites them and "takes" the ownership, so it uninstalls them too.

This doesn't solve that the WIN10 plugin files are there from an earlier installation, I need to find another solution for this.

Christian-Schulz:
> This doesn't solve that the WIN10 plugin files are there from an earlier installation, I need to find another solution for this.

There is an [InstallDelete] section
```
[InstallDelete]
Type: files; Name: "{app}\Plugins\Win10\Greenshot.Plugin.Win10.dll"
Type: files; Name: "{app}\Plugins\Win10\Microsoft.Toolkit.Uwp.Notifications.dll"
Type: dirifempty; Name: "{app}\Plugins\Win10"
```

or 

```
[InstallDelete]
Type: filesandordirs; Name: "{app}\Plugins\Win10"
```

Inno Setup should delete this files after the app is closed.

Lakritzator:
Yeah, that's what I am currently testing:

```
[InstallDelete]
Type: filesandordirs; Name: "{app}\Plugins"
```

It's not the best behavior, but to be honest...
I could also name all the plugins WE ever provided, maybe that prevents that we uninstall other plugins, if things like that exist. On the other side, they probably do not work anyway after an upgrade.

Lakritzator:
FYI, when I wrote the other comment, I already saw that it works in the "good" case, I am not sure yet about what happens if Greenshot is still running. Maybe I need to reintroduce the AppMutex again, checking the behavior now.

I will also provide a PR for the installer for 1.3, as we are releasing some fixes for after 1.3.312
This would at least solve some issues...

Lakritzator:
See #1085

PR #1084, opened 2026-03-18
Reducing issues with leftovers of installation issues

Lakritzator:
This removes the plugins from 1.2 and 1.3 for the cases that an uninstall had issues.
It was specifically made in such a way that only the plugins on the system that WE provide are deleted before installing the new files.


Lakritzator:
Here is another fix for 1.3, this should reduce the issues with leftovers causing crashes

PR #1085, opened 2026-03-18
Fix issues with leftover files due to bad install/uninstall

Lakritzator:
Similar to the 1.3 fix, this will remove all leftover plugins, which cause issues when starting the application.

Additionally the "ignoreversion" flag was missing, causing leftovers there before an installation to be outside of the uninstall, and staying there.

This isn't finished yet, still needs some testing if the uninstall works when the application is running while starting the installation.

### DIFF
--- a/src/Greenshot-Installer/setup.iss
+++ b/src/Greenshot-Installer/setup.iss
@@ -10,97 +10,97 @@
 #define ReleaseDir "..\Greenshot\bin\Release\net480"
 #define PluginDir "..\Greenshot\bin\Release\net480\Plugins"
 #define CertumThumbprint GetEnv('CertumThumbprint')
+#define DefaultInstallFlags "overwritereadonly ignoreversion"
 
 [Files]
-Source: {#ReleaseDir}\Greenshot.exe; DestDir: {app}; Components: greenshot; Flags: overwritereadonly
-Source: {#ReleaseDir}\Greenshot.Base.dll; DestDir: {app}; Components: greenshot; Flags: overwritereadonly
-Source: {#ReleaseDir}\Greenshot.Editor.dll; DestDir: {app}; Components: greenshot; Flags: overwritereadonly
-Source: {#ReleaseDir}\Greenshot.exe.config; DestDir: {app}; Components: greenshot; Flags: overwritereadonly
-Source: {#ReleaseDir}\log4net.dll; DestDir: {app}; Components: greenshot; Flags: overwritereadonly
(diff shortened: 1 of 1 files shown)


Lakritzator:
One thing which is still open, is if we might need to get the AppMutex check in the installer back, this will be addressed in a different issue/PR.

src/Greenshot-Installer/setup.iss
```diff
@@ -701,4 +645,36 @@ Filename: "{app}\{#ExeName}.exe"; Description: "{cm:startgreenshot}"; Parameters
...
+Type: filesandordirs; Name: "{app}\Plugins\Greenshot.Plugin.Dropbox"
+Type: filesandordirs; Name: "{app}\Plugins\Greenshot.Plugin.ExternalCommand"
+Type: filesandordirs; Name: "{app}\Plugins\Greenshot.Plugin.Flickr"
+Type: filesandordirs; Name: "{app}\Plugins\Greenshot.Plugin.GooglePhotos"
+Type: filesandordirs; Name: "{app}\Plugins\Greenshot.Plugin.Imgur"
+Type: filesandordirs; Name: "{app}\Plugins\Greenshot.Plugin.Jira"
+Type: filesandordirs; Name: "{app}\Plugins\Greenshot.Plugin.Office"
+Type: filesandordirs; Name: "{app}\Plugins\Greenshot.Plugin.Photobucket"
+Type: filesandordirs; Name: "{app}\Plugins\Greenshot.Plugin.Win10"
+
```
Christian-Schulz, line 674: @Lakritzator I think the folder names are wrong.
I think it should be 
`"{app}\Plugins\Win10"`

Lakritzator: It's worse, we are both right... those directories are the plugin directories for the portable version...

Christian-Schulz: They are different?? 🤦I know the person who defined the folder names in the portable version. 

This makes mixing postable version and install version an even bigger problem.
Should we rename it there?

Lakritzator: Just download the zip files and you will see... It makes sense to keep them the same yeah...

Christian-Schulz: I think I've installed Greenshot at least 20 times this week in my Test-VM.

Lakritzator: That's nothing, I have installed Greenshot probably >1000 times. 😝

PR #1087, opened 2026-03-19
Update ScaleHelper.cs to fix #1055

### LINKED ISSUE #1055: Draw ellipses and rectangles from center with CTRL. (This used to be a feature?)
### Please confirm the following before submitting a Feature Request

### Is your feature request related to a problem? Please describe.

I am pretty sure this used to be a feature up until 2 or 3 versions ago.. 
When drawing an ecclipse or rectangle:

CTRL + click (and then drag) : draw ellipse or rectangle from center point
SHIFT + click (and then drag) : draw square or circle (already a feature!)
CTRL + SHIFT click (and then drag) : draw square or circle from center point

### Describe the solution you'd like

This is the behaviour i am looking for. I am 85% confident that this exact behaviour was already included up until 2 or 3 versions ago

Image

### Additional context

I am 85% confident that this exact behaviour was already included up until 2 or 3 versions ago.

Also I love your greenshot software and the fact that is is freeware, your effort is so much appriciated!
Please let me know the donation amount required to get this feature request to the top of the list so I can organise a fundraise among my colleagues, thank you!

danrhodes:
Replaced the broken Inflate-based centering with correct geometry: use result.Width/Height as the half-size from the click origin, then build {clickX - halfW, clickY - halfH, 2*halfW, 2*halfH}.

The old code computed wdiff = (result.Width - boundsBeforeResize.Width) / 2 which is always 0 when SHIFT isn't also held, making CTRL a no-op. The CTRL+SHIFT weirdness was the same bug manifesting differently.

Fixes https://github.com/greenshot/greenshot/issues/1055

### DIFF
--- a/src/Greenshot.Editor/Helpers/ScaleHelper.cs
+++ b/src/Greenshot.Editor/Helpers/ScaleHelper.cs
@@ -243,9 +243,12 @@ namespace Greenshot.Editor.Helpers
 
             if (centeredScale)
             {
-                float wdiff = (result.Width - boundsBeforeResize.Width) / 2;
-                float hdiff = (result.Height - boundsBeforeResize.Height) / 2;
-                result = result.Inflate(wdiff, hdiff);
+                // The click point is the center: expand the result symmetrically around it.
+                // boundsBeforeResize.X/Y is the click origin (set in HandleMouseDown).
+                // result.Width/Height is the half-size (cursor distance from origin).
+                float halfW = result.Width;
+                float halfH = result.Height;
+                result = new NativeRectFloat(boundsBeforeResize.X - halfW, boundsBeforeResize.Y - halfH, halfW * 2, halfH * 2);
             }
 
             return result;


danrhodes:
Gemini is wrong — the result already has the correct size before centeredScale runs.

**Here's wh**y: Before Scale is even called, line 560 calculates _boundsAfterResize with Width = cursorX - startX and Height = cursorY - startY. That already-sized rectangle is what gets passed into Scale. So when result = boundsBeforeResize runs on line 224, result is not zero — it already has the right dimensions.

Gemini seems to have confused _boundsBeforeResize (the zero-size rect stored at mouse-down) with _boundsAfterResize (the already-sized rect calculated just before Scale is called). They're two different fields with very similar names.

Gemini's suggested fix would also break CTRL-only drawing. The angle/dist/cos/sin math it adds is polar coordinate conversion — it turns the cursor position into a single distance and angle from the origin, then derives width and height from that.

The result is not a free-form rectangle; the width and height end up coupled through the angle, so you lose the ability to draw a non-square centred shape. That's the same constraint SHIFT already applies. CTRL-only should give you a free ellipse or rectangle centred on your click point — not a circle or square.

Lakritzator:
Nobody said they are perfect 🤷‍♂️

And whatever Gemini said, I tried your solution and it works great.

Issue #1090, opened 2026-03-20
Output setting


Shahab1111:
### Please confirm the following before submitting a Bug Report

- [x] I have searched existing open and closed issues and did not find a duplicate

### Greenshot version

1.3.315

### Describe the bug

Software version: 1.3.315 - 420b08b2a9 (64 bit)
.NET runtime version: 4.0.30319.42000+
Time: 2026-03-20 15:13:14 +01:00
OS: Windows 11 x64 build 26200
GDI object count: 113
User object count: 138

Exception: System.InvalidOperationException
Message: Unable to retrieve the root folder.

Stack:
   at System.Windows.Forms.FolderBrowserDialog.RunDialog(IntPtr hWndOwner)
   at System.Windows.Forms.CommonDialog.ShowDialog(IWin32Window owner)
   at Greenshot.Forms.SettingsForm.BrowseClick(Object sender, EventArgs e)
   at System.Windows.Forms.Control.OnClick(EventArgs e)
   at System.Windows.Forms.Button.OnMouseUp(MouseEventArgs mevent)
   at System.Windows.Forms.Control.WmMouseUp(Message& m, MouseButtons button, Int32 clicks)
   at System.Windows.Forms.Control.WndProc(Message& m)
   at System.Windows.Forms.ButtonBase.WndProc(Message& m)
   at System.Windows.Forms.Button.WndProc(Message& m)
   at System.Windows.Forms.NativeWindow.Callback(IntPtr hWnd, Int32 msg, IntPtr wparam, IntPtr lparam)

Configuration dump:

---
**Stack Trace Hash:** `15bb161808d5` [stack-trace-hash: 15bb161808d5]

danrhodes:
This has been reported many times before, I assume your IT Deptartment installed GS for you? In the red box enter a path you can access such as C:\ and then click the dots to change the path.

danrhodes:
Same as #1040 #894 #785

Issue #1040, opened 2026-03-09
can't update any preferences


luclarson01-jpg:
### Please confirm the following before submitting a Bug Report

- [x] I have searched existing open and closed issues and did not find a duplicate

### Greenshot version

1.3.312

### Describe the bug

Software version: 1.3.312 - 5a60aff83e (64 bit)
.NET runtime version: 4.0.30319.42000+
Time: 2026-03-09 08:25:23 -04:00
OS: Windows 10 2009 x64 build 26200
GDI object count: 76
User object count: 90

Exception: System.InvalidOperationException
Message: Unable to retrieve the root folder.

Stack:
   at System.Windows.Forms.FolderBrowserDialog.RunDialog(IntPtr hWndOwner)
   at System.Windows.Forms.CommonDialog.ShowDialog(IWin32Window owner)
   at Greenshot.Forms.SettingsForm.BrowseClick(Object sender, EventArgs e)
   at System.Windows.Forms.Control.OnClick(EventArgs e)
   at System.Windows.Forms.Button.OnMouseUp(MouseEventArgs mevent)
   at System.Windows.Forms.Control.WmMouseUp(Message& m, MouseButtons button, Int32 clicks)
   at System.Windows.Forms.Control.WndProc(Message& m)
   at System.Windows.Forms.ButtonBase.WndProc(Message& m)
   at System.Windows.Forms.Button.WndProc(Message& m)
   at System.Windows.Forms.NativeWindow.Callback(IntPtr hWnd, Int32 msg, IntPtr wparam, IntPtr lparam)

Configuration dump:

---
**Stack Trace Hash:** [15bb161808d5]

danrhodes:
Is the text box where you enter the output location red? If so try entering a known folder you can access and then Ok out and go back in this should solve the issue. Is GS installed by you or yout IT Dept?

danrhodes:
Same as #894 #785 #1090

Issue #894, opened 2026-01-30
Cannot select path for storing files - with this cannot change any settings


chrisch82-sketch:
Software version: 1.3.312 - 5a60aff83e (64 bit)
.NET runtime version: 4.0.30319.42000+
Time: 2026-01-30 11:00:13 +01:00
OS: Windows 10 2009 x64 build 26100
GDI object count: 85
User object count: 115

Exception: System.InvalidOperationException
Message: Unable to retrieve the root folder.

Stack:
   at System.Windows.Forms.FolderBrowserDialog.RunDialog(IntPtr hWndOwner)
   at System.Windows.Forms.CommonDialog.ShowDialog(IWin32Window owner)
   at Greenshot.Forms.SettingsForm.BrowseClick(Object sender, EventArgs e)
   at System.Windows.Forms.Control.OnClick(EventArgs e)
   at System.Windows.Forms.Button.OnMouseUp(MouseEventArgs mevent)
   at System.Windows.Forms.Control.WmMouseUp(Message& m, MouseButtons button, Int32 clicks)
   at System.Windows.Forms.Control.WndProc(Message& m)
   at System.Windows.Forms.ButtonBase.WndProc(Message& m)
   at System.Windows.Forms.Button.WndProc(Message& m)
   at System.Windows.Forms.NativeWindow.Callback(IntPtr hWnd, Int32 msg, IntPtr wparam, IntPtr lparam)

Configuration dump:

<img width="406" height="437" alt="Image" src="https://github.com/user-attachments/assets/c945dd0b-b3ee-44b5-aa19-46808e7d0d5d" />

---
**Stack Trace Hash:** [15bb161808d5]

Christian-Schulz:
It looks pretty much the same like in #785, but in a different situation.

danrhodes:
This happens when installed via system user. Did your work install this for you?

danrhodes:
FYI you can type in a path and it works again.

chrisch82-sketch:
Yes it was work installed.
I cannot enter a path here. That's exactly my problem.
So an admin has to fix this on my work?

danrhodes:
If you type in c:\ and then press the button to the right it should work.

danrhodes:
Or also what worked for me was closing and re opening GS.

chrisch82-sketch:
> If you type in c:\ and then press the button to the right it should work.

Perfect, that work for me

danrhodes:
[...]

Issue #785, opened 2026-01-06
Error


atlantis1021:
I have received the following error in Greenshot after having to force reboot my workstation

Greenshot was frozen and reboot was required. Now I am trying to program capture using the F1 button and select the output, but get the following error.

**To Reproduce**
Example steps to reproduce the behavior:
1. Preferences
2. General tab
3. Change Capture Region to F1
4. Press OK
5. Select Storage Location under Output tab by clicking on the ellipses ...
6. Error appears

**Expected behavior**
I expect to be able to program my F1 button to capture the region of interest and output at my selected destination.

**Screenshots**

**Versions (please complete the following information):**
 - Greenshot version -  1.3.304 - f46c1b7f31 (64 bit)
 - Windows version - OS: Windows 10 2009 x64 build 26100 
 - I am a Win 11 device. IS tried to push an update to the workstation and has been unsuccessful.

**Additional context**
The error

Software version: 1.3.304 - f46c1b7f31 (64 bit)
.NET runtime version: 4.0.30319.42000+
Time: 2026-01-06 16:12:03 -05:00
OS: Windows 10 2009 x64 build 26100
GDI object count: 81
User object count: 110

Exception: System.InvalidOperationException
Message: Unable to retrieve the root folder.

Stack:
   at System.Windows.Forms.FolderBrowserDialog.RunDialog(IntPtr hWndOwner)
   at System.Windows.Forms.CommonDialog.ShowDialog(IWin32Window owner)
   at Greenshot.Forms.SettingsForm.BrowseClick(Object sender, EventArgs e)
   at System.Windows.Forms.Control.OnClick(EventArgs e)
   at System.Windows.Forms.Button.OnMouseUp(MouseEventArgs mevent)
   at System.Windows.Forms.Control.WmMouseUp(Message& m, MouseButtons button, Int32 clicks)
   at System.Windows.Forms.Control.WndProc(Message& m)
   at System.Windows.Forms.ButtonBase.WndProc(Message& m)
   at System.Windows.Forms.Button.WndProc(Message& m)
   at System.Windows.Forms.NativeWindow.Callback(IntPtr hWnd, Int32 msg, IntPtr wparam, IntPtr lparam)

Configuration dump:

Lakritzator:
Doesn't sound like an issue with the hotkey, but somehow Greenshot has an issue with the folder browser and the path where it tries to go to. Almost like you removed an USB stick or something like that.

[...]

PR #1097, opened 2026-03-20
Fix for - Unable to retrieve the root folder

danrhodes:
This PR addresses a real-world pain point reported by users in IT-managed environments where Greenshot is silently installed or updated under a system/service account. When the end user subsequently opens Settings and clicks the browse button for the storage location, the Windows shell fails to resolve the root folder for the system account context, throwing System.InvalidOperationException: Unable to retrieve the root folder and leaving the user with a red invalid-path box and no way forward.

Rather than attempting to fix the deeper process-context issue (which is out of scope), this fix hardens the BrowseClick handler with a graceful degradation strategy: on the first failure it retries the dialog rooted at MyComputer — which doesn't require a user profile to resolve — seeded with the current interactive user's MyDocuments path via Environment.GetFolderPath, which correctly resolves against the logged-in user's token regardless of the account Greenshot started under. If the dialog still cannot open, rather than leaving the user stuck, it populates the storage location textbox with the first accessible user folder found (Documents → Desktop → Temp) and shows a localised informational message explaining what happened. The fix is entirely contained within the catch path so zero behaviour changes for normal users, and the two new language keys (settings_storagelocation_folder_error / settings_storagelocation_folder_error_title) follow the existing localisation pattern so translations can be contributed in the usual way.

Fixes a massive pain point - #1040 #894 #785 #1090

### DIFF
--- a/src/Greenshot/Configuration/LanguageKeys.cs
+++ b/src/Greenshot/Configuration/LanguageKeys.cs
@@ -70,6 +70,8 @@ namespace Greenshot.Configuration
         settings_tooltip_language,
         settings_tooltip_primaryimageformat,
         settings_tooltip_storagelocation,
+        settings_storagelocation_folder_error,
+        settings_storagelocation_folder_error_title,
         settings_visualization,
         settings_window_capture_mode,
         tooltip_firststart,
--- a/src/Greenshot/Forms/SettingsForm.cs
+++ b/src/Greenshot/Forms/SettingsForm.cs
@@ -666,9 +666,14 @@ namespace Greenshot.Forms
             }
             catch (InvalidOperationException ex)
             {
-                Log.Warn("Problem opening folder browser dialog, retrying with empty path: ", ex);
-                // Retry with an empty SelectedPath so the user can still browse and correct the path
-                folderBrowserDialog1.SelectedPath = string.Empty;
+                // This can happen when Greenshot was launched under a system/service account (e.g. after an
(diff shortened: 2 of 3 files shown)


Lakritzator:
Question: Shouldn’t this be solved with the updated installer?

danrhodes:
> Question: Shouldn’t this be solved with the updated installer?

Absolutely it should, is, I did think this at the time, and indeed having tested the new installer it works and opens back in under the user context 👌

Might still be worth including to cover any edge cases where the user is unable to access the location, the code is there with regression if you think it would be useful?

Different cause same symptom? 🤷‍♂️

Lakritzator:
@danrhodes I merged it for now but wanted to place the following remarks:

- I think I wrote the code in such a way, that if nothing was set it will automatically take a sensible value. BUT I had issues with users who didn't mount (or it was delayed) their "USB Stick" or "Network drive", and Greenshot while starting tried to validate the value and changing it to the default was causing major conflicts with usability.
- I am trying to finally migrate Greenshots .ini file reading to an external library I wrote ages ago, and recently updated (better API & documentation): https://github.com/dapplo/Dapplo.Ini/wiki this ("Validate on configuration reading") will again be an issue I need to consider.

danrhodes:
Thanks for merging, and good shout on the validation timing issue — that’s a nasty edge case with removable/network drives.

Happy to look at that as a follow-up PR once you’re further along with the Dapplo.Ini migration.

If it’d help to have an extra pair of hands on the validation behaviour when you get there, just shout. 👍

PR #1108, opened 2026-03-24
Update ToastNotificationService.cs to fix #758

### LINKED ISSUE #758: Greenshot hangs in Microsoft.Toolkit.Uwp.Notifications code
**Describe the bug**
Greenshot hangs when it tries to display a Toast-Notification

**To Reproduce**
1. Capture '....'
2. Open Editor
-> Editor already hangs

**Expected behavior**
No hang

 - Main branch
 - Windows version 11 Pro OS Build 26200.7462

**Additional context**
I also noticed this for the current released version, it might have to do with my PC.

The line of code where it hangs, visible if I pause the application, is here:
https://github.com/greenshot/greenshot/blob/dc36b6dcba9cf5fba119a66eefa85f28c8ab0597/src/Greenshot.Plugin.Win10/ToastNotificationService.cs#L113

danrhodes:
Here goes..... - Accessing toastNotifier.Setting performs a synchronous WinRT COM interop call that marshals into the Windows notification subsystem.

On certain Windows 11 24H2 builds (observed around 26200.7462–26200.7840), a regression in the notification platform — sometimes exacerbated by security software intercepting COM calls — causes this synchronous call to block indefinitely, deadlocking the UI thread and making the entire Greenshot window unresponsive.

Because the call is inherently synchronous in the WinRT projection, we can't make it truly async, so instead we run it on a ThreadPool thread via Task.Run and impose a 500ms timeout.

On a healthy system, the call returns in single-digit milliseconds, so 500ms is a generous safety margin, but it means a user on a broken notification subsystem waits at most half a second before Greenshot's UI is usable again rather than hanging indefinitely.

If the call completes in time, we act on the result normally; if it times out or throws, we log a warning and skip the setting check, falling through to attempt the toast anyway.

The fix is not specific to Windows 11 24H2 — because it offloads the blocking WinRT call off the UI thread entirely, it will protect against any future OS regression or security software interaction that causes the same hang, regardless of Windows version.

Fixes #758

### DIFF
--- a/src/Greenshot/Native/ToastNotificationService.cs
+++ b/src/Greenshot/Native/ToastNotificationService.cs
@@ -22,6 +22,7 @@
 using System;
 using System.Drawing.Imaging;
 using System.IO;
+using System.Threading.Tasks;
 using Windows.Foundation.Collections;
 using Windows.Foundation.Metadata;
 using Windows.UI.Notifications;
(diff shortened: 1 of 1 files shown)


Lakritzator:
On the line with var settingTask = Task.Run(() => toastNotifier.Setting);
I see the following when debugging:

System.Exception
  HResult=0x8001010E
  Message=The application called an interface that was marshalled for a different thread. (Exception from HRESULT: 0x8001010E (RPC_E_WRONG_THREAD))
  Source=<Cannot evaluate the exception source>
  StackTrace:
<Cannot evaluate the exception stack trace>

Might have to do with changes I am making for #1116

PR #1109, opened 2026-03-24
Ability to close v1.3 with v1.4 installer

Christian-Schulz:
This is a PR only for v1.3.
If you have v1.3 installed and use the installer from v1.4.163+ Inno Setup tries to close Greenshot. 
But it closes not all. The UI thread and system tray icon is gone and the Hotkey is unregistered. Some threads are left.

At first I thought it was the UpdateService and I spent a lot of time tweaking the service, but that wasn't necessary. It wasn't the only thread. I couldn't identify the last other one.

I'll create a separate PR for my changes to the UpdateService in the dev branch.

Inno Setup uses the Windows Restart Manager. Since Greenshot v1.3 is not registered, Windows sends a `WM_DESTROY` message.

I reacted to this and call `Exit()` manually.

I'am not 100% sure if this is the right way, so i marked it as draft. 

- It works if no Editor is open. v1.3 is terminated and the last step of the Installer starts v1.4.163+
- It works if all open Editors have no modified content.
- It works if an Editor has modified content, you see the save question dialog and you save it.
- it don't work if an Editor has modified content and you are not fast enough to save it. (there is a timeout in the Installer)
- it don't work if an Editor has modified content and you cancel the save question dialog.

### DIFF
--- a/src/Greenshot.Editor/Forms/ImageEditorForm.cs
+++ b/src/Greenshot.Editor/Forms/ImageEditorForm.cs
@@ -928,16 +928,34 @@ namespace Greenshot.Editor.Forms
             UpdateUndoRedoSurfaceDependencies();
         }
 
+        private static bool IsExternalShutdownCloseReason(CloseReason closeReason)
+        {
+            return closeReason == CloseReason.ApplicationExitCall || closeReason == CloseReason.WindowsShutDown || closeReason == CloseReason.TaskManagerClosing;
+        }
+
+        // only for workaround for the installer from 1.3 to 1.4.
+        private bool _closeWithoutSavePrompt;
+
         private void ImageEditorFormFormClosing(object sender, FormClosingEventArgs e)
         {
-            if (_surface.Modified && !EditorConfiguration.SuppressSaveDialogAtClose)
+            void ForceCloseWithoutPrompt () 
+            {
+                // only for workaround for the installer from 1.3 to 1.4.
+                // When the installer tries to close the editor and the closing takes too long, the installer cancels the outer closing process
+                // So we force the closing without prompt again and the editor will not be in a wierd state
+                e.Cancel = true;
+                _closeWithoutSavePrompt = true;
+                BeginInvoke((MethodInvoker)delegate { Close(); });
+            };
+
+            if (!_closeWithoutSavePrompt && _surface.Modified && !EditorConfiguration.SuppressSaveDialogAtClose)
             {
                 // Make sure the editor is visible
                 WindowDetails.ToForeground(Handle);
 
                 MessageBoxButtons buttons = MessageBoxButtons.YesNoCancel;
                 // Dissallow "CANCEL" if the application needs to shutdown
-                if (e.CloseReason == CloseReason.ApplicationExitCall || e.CloseReason == CloseReason.WindowsShutDown || e.CloseReason == CloseReason.TaskManagerClosing)
+                if (IsExternalShutdownCloseReason(e.CloseReason))
                 {
                     buttons = MessageBoxButtons.YesNo;
                 }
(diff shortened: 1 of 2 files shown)


Lakritzator:
I would expect the message to be send out to every open top level window, so all the editors. And I believe that WinForms has a default handling for this, which is why most editor windows suddenly "ask" to save.

But isn't first a WM_CLOSE send by the installer, I'm pretty sure about that... maybe the WM_DESTROY comes after.

Christian-Schulz:
Maybe the open ImageEditorForms gets WM_CLOSE from the Installer before my changes. I didn't test it.

The MainFrame only gets WM_DESTROY and WM_NCDESTROY. 

Since I react to WM_DESTROY and call Exit() directly. Exit() calls Close() on every editor.

My motivation for these changes was: The installer was unable to completely terminate Greenshot , but the systemtray icon was gone. So the user had to use the task manger to kill Greenshot manually.

Question: Should we simply close Greenshot without prompting the user in the specific case where someone updates from v1.3 to 1.4?
Or should the user actively close Greenshot before running the installer?
Or should we use my solution, which works in most cases and could theoretically fail only if editors are open?

Lakritzator:
I think it's good to prepare ourselves in the next bugfix, we probably have some stuff we can backport.

We should just inform the installer that we aren't ready (return false to the ENDSESSION messages) if there are still modifications. The installer will then say something the application still running and provide the user to force an update, which is on them.

This can be done with something like this (pseudo code, didn't try) in the editor WndProc override:

```
protected override void WndProc(ref Message m)
{
     switch (m.Msg)
     {
         case WindowsMessages.WM_QUERYENDSESSION:
         case WindowsMessages.WM_ENDSESSION:
             message.Result = _surface.Modified?(IntPtr)HResult.S_FALSE:(IntPtr)HResult.S_TRUE; 
             return;
     };
     // old code and call base.WndProc
     base.WndProc(ref m);
} 

```

If then the app is forced, we should try to do this as gracefully as possible, which you already do with handling the WM_DESTROY.

Christian-Schulz:
FYI: 
I tried your Dapplo InstallerExample and reproduced the behavior.
The current Inno Setup from v1.4.163 exactly calls the Shutdown() without any action flag. 
like so:
```
 ...
private void TryRestart()
        {
            using var session = InstallerRestartManager.CreateSession();
            session.RegisterFile(ExeToInstall);
            var processes = session.GetProcessesUsingResources();
            
            foreach (var process in processes)
            {
                AddLine($"Process {process.strAppName} (PID: {process.Process.dwProcessId}) is using the file, status: {process.AppStatus}");
            }
            try
            {
                //session.Shutdown(Kernel32.Enums.RmShutdownType.RmForceShutdown, (progress) => //  force shutdown, if setup.iss.CloseApplications=force
                session.Shutdown(0, (progress) => // shutdown without specific action flag, if setup.iss.CloseApplications=yes
                {
                    this.Dispatcher.BeginInvoke(() =>
                    {
                        AddLine($"Shutdown progress {progress}");
                    });
                });
            }
```

Lakritzator:
I'm not 100% clear on what you wanted to say with the last comment...

Christian-Schulz:
Just for your information.
I took a closer look at InnoSetup and figured out exactly how the process is terminated.

If the installer is configured with `CloseApplications=yes` like we do in v1.4 it calls [RmShutdown](https://learn.microsoft.com/en-us/windows/win32/api/restartmanager/nf-restartmanager-rmshutdown) with `lActionFlags=0`
If the installer is configured with `CloseApplications=force` it calls RmShutdown with `lActionFlags=0x1`

Lakritzator:
Interesting, as 0x00 is not really a valid value according to the documentation

Christian-Schulz:
I thought the same thing.
The example in the documentation does it too. https://learn.microsoft.com/en-us/windows/win32/rstmgr/using-restart-manager-with-a-secondary-installer

Lakritzator:
Besides the "exit" remark in the review, and I think that is fair we need to be idempotent, do you think we can finalize this? I'm not sure what we do with the 0x00 fact... if we even have to do anything, as you already tested against 1.4
I'd like to get this into 1.3 and we are planning to release an update...

Christian-Schulz:
I tested it again with 1.4 installer. 

With these changes, it is no longer necessary to use Task Manager to terminate the remaining process. The icon in the system tray is available to the user at any time.
In most cases, the installer can completely terminate the Greenshot process. No thread remains hidden in the background.

There is one situation that, in my opinion, cannot be resolved: when the user has unsaved changes in the editor and takes too long (appr. 30 seconds) to save the file.
The installer has an internal timeout. In that case, the “Retry” option won't work either. The user must cancel the installer and restart it.

Lakritzator (approved):

Issue #1116, opened 2026-03-26
Capture screen region is copying the region plus all screens?


danrhodes:
### Please confirm the following before submitting a Bug Report

- [x] I have searched existing open and closed issues and did not find a duplicate

### Greenshot version

1.40167

### Describe the bug

<img width="438" height="403" alt="Image" src="https://github.com/user-attachments/assets/b6f0f743-4861-467b-9e71-1a34a00d9419" /> 

I think it might be picking up the button press in Windows, even though my button is disabled. If I right-click GS and click Capture region, this doesn't happen.

Didn't we change the way we capture the PrtSc button recently?

<img width="1275" height="123" alt="Image" src="https://github.com/user-attachments/assets/85863395-ba8e-4afc-90f1-72a4470bf7e3" />

Lakritzator:
I'm not 100% sure what you mean.

But I changed the complete hotkey handling, we are now working with raw-input, instead of https://learn.microsoft.com/en-us/windows/win32/inputdev/wm-hotkey

Can you try to describe what you mean differently, or maybe my brain doesn't work today...

danrhodes:
Sure, I'll add more info later, it was more of a reminder for me to check the code later 👍

danrhodes:
I've just checked, something must have changed in Windows....I have the button in Windows Settings set to Off as below and Greenshot is not running.

<img width="1275" height="123" alt="Image" src="https://github.com/user-attachments/assets/dcb3e639-ffe2-4dee-8197-49a865d26ba2" />

When I click PrtSc I get a full screenshot of all of my monitors on my clipboard.

<img width="445" height="324" alt="Image" src="https://github.com/user-attachments/assets/7b1849cb-dc00-4d17-bd60-8a19579b3338" />

When I have Greenshot running (installed with the disable Windows PrtSc option ticked in setup) I get both the full screen capture of all monitors AND the region I captured.

<img width="438" height="403" alt="Image" src="https://github.com/user-attachments/assets/3268ee7d-acdf-427b-aeb9-e1a647677fa9" />

If I then close GS and press PrtSc again I get just a full screen capture on my clipboard of all my displays.

<img width="445" height="324" alt="Image" src="https://github.com/user-attachments/assets/5b85810b-c41f-448d-b80e-753adb8895e6" />

This has only just started happening in the last week.

Lakritzator:
Let me think about this

Lakritzator:
[...]

PR #1119, opened 2026-03-27
Post the handling of the hotkey outside of the current Low-Level-Keyboard hook thread, as in many cases it takes longer then ~1000ms.

Lakritzator:
As described in #1116 the hotkey key press wasn't suppressed, this was due to the screenshot logic taking to long (doing stuff /showing things)

This fix posts the call to the UI thread for "later", placing it outside of the Low-Level Keyboard hook.

### DIFF
--- a/src/Greenshot/Helpers/HotkeyHelper.cs
+++ b/src/Greenshot/Helpers/HotkeyHelper.cs
@@ -23,6 +23,7 @@ using System;
 using System.IO;
 using System.Linq;
 using System.Text;
+using System.Threading;
 using System.Windows.Forms;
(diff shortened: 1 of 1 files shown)


Lakritzator:
@danrhodes This should fix it... waiting for reviews from Gemini 😂

Lakritzator:
@danrhodes merging this, maybe not the pretties solution, but we can make this conceptually more beautiful another time.
I am still wishing for "Greenshot recipes" to be one of the features in the future (making it possible to define a more specific flow, using some standardized building blocks. which triggers on x)

PR #1121, opened 2026-03-27
Collection of backports of issues fixed in 1.4

Lakritzator:
This PR will combine a couple of fixes which were made for 1.4

First one would be a backport of the fix in #1010

More to follow...


Lakritzator:
Wrongly created...

PR #1122, opened 2026-03-27
Collection of backports for 1.3 with fixes which were made for 1.4

Lakritzator:
This PR will be used to collect fixes for different issues which were made for 1.4

First: Backport of #1010, solving issues with resizing our icons, which throws: System.ArgumentException

### DIFF
--- a/src/Greenshot.Base/Core/AbstractDestination.cs
+++ b/src/Greenshot.Base/Core/AbstractDestination.cs
@@ -181,7 +181,8 @@ namespace Greenshot.Base.Core
             {
                 ImageScalingSize = CoreConfig.IconSize,
                 Tag = null,
-                TopLevel = true
+                TopLevel = true,
+                Font = new Font(FontFamily.GenericSansSerif, 9) // set new default font, so we are allowed to dispose it later, we will scale it later on the Opening event
             };
 
             menu.Opening += (sender, args) =>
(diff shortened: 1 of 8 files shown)


Lakritzator:
@danrhodes and @Christian-Schulz if you have any candidates for a backport for 1.3 which make a lot of sense to include, issues we see a lot of times, comment here and I will consider them.

danrhodes:
How about all the memory and performance improvements that have been included in 1.4?

Lakritzator:
> How about all the memory and performance improvements that have been included in 1.4?

Good point, I didn't select them as I prefer to keep the changes to a minimum, so short fixes without potential side effects.
Besides we should have something left for 1.4 😉

Christian-Schulz:
I would say #1109.
That wasn't actually reported. I only noticed it during my installation tests, but it makes the later upgrade from 1.3 to 1.4 run a bit more smoothly.

Lakritzator:
@Christian-Schulz To that one I agree. as soon as we finalize it.

Christian-Schulz:
@Lakritzator I'll look into it tomorrow. But maybe that could wait until the next backport.

Lakritzator:
Suggestion is to place #1170 on the list, take "IsOneDriveBlockingHotkey" from 1.4 where it moved to HotkeyHelper.

Lakritzator:
@jklingen I think this is the final list of backported bug & security fixes for 1.3, additionally take #1092 and #1109 (which is almost ready).

I can finalize #1109 (to make 1.3 future proof and behave better when an 1.4 update comes) today/tomorrow.

Issue #1129, opened 2026-03-29
selected arrow tip not marked in drop down list


octaeder:
### Please confirm the following before submitting a Bug Report

- [x] I have searched existing open and closed issues and did not find a duplicate

### Greenshot version

1.4.93 - 9a0b16a7dd (Win11, 64bit installer)

### Describe the bug

open an image/screenshot. Select arrow drawing tool. Open arrow tip drop down list. There is no item selected or it is not noticable, even after selecting a tip.

<img width="113" height="115" alt="Image" src="https://github.com/user-attachments/assets/33e55053-873f-4c65-bea9-2bf556078fae" />

It could also be helpful if the icon in the top-left corner of the dropdown reflected the current selection, since it remains visible even when the list is closed.

kamilla11:
Hello! I have implemented a fix for this issue.

- Now the selected item is displayed in the list
- The button icon is also updated depending on the selection.

The changes are minimal and affect only the UI.

The PR has been sent, please take a look

PR #1136, opened 2026-03-31
Fix arrow head selection visibility and improve dropdown button

kamilla11:
**Issue selected arrow tip not marked in drop down list** #1129
### The problem
The arrow tip selection drop-down list does not display the currently selected option. The user does not see which item is active.

### Solution
- Added visual highlighting of the selected item via `Checked`
- The dropdown button now displays the current selection via the icon

### Result
- The user sees the selected item
- The button reflects the current selection with the desired icon

### Note
- The changes relate only to the UI
- The logic of the application has not been changed

I also attach a screenshot of the current implementation.
image
image
image
image
image


Lakritzator:
Looking good, definitively the right direction.

Reviewing this I found a few issues:

1. I think something is still missing,  what about representing the current arrow head if an arrow is selected? So before someone changes it, it should also reflect the correct arrowhead.
2. I need to look at the icons, they are bigger than the ones we used, and I hope they are from the fugue set? https://p.yusukekamiyamane.com/

kamilla11:
Thanks for the feedback. I'll take it for revision. 
Icons are not from this set, I'll fix them too.

octaeder:
Hey, thanks for implementing this. The images look exactly as I expected. Now I can even figure out what the words mean, even though I’m not familiar with the language in use 😉

kamilla11:
Hello! Sorry for the long absence. I have finalized the solution based on your comments:
1. Now the screen immediately displays the correct arrowhead before someone's changes. And instead of a checkbox, an arrow image is displayed.
2. The previous icons have been replaced with icons from the collection https://p.yusukekamiyamane.com and are in the same format as arrowHeadsDropDownButton.Image (embedded bitmap). The matching of the images is as follows:

- arrowHeadBothMenuItem - arrow-resize.png
- arrowStartEndMenuItem - arrow-180-medium.png
- arrowHeadEndMenuItem - arrow-000-medium.png
- arrowHeadNoneMenuItem - control-pause.png (edited)

I couldn't find the blue "minus", so I edited the pause image.

Screenshots of the new version are attached below

kamilla11:
Hello! Can you check the fixes, please?

Lakritzator:
Great work, sorry for the delay, had some time off due to some private reasons.

Issue #1153, opened 2026-04-08
Crashes when trying to capture a region


AnadarProSvcs:
### Please confirm the following before submitting a Bug Report

- [x] I have searched existing open and closed issues and did not find a duplicate

### Greenshot version

1.3.315

### Describe the bug

When I try to capture a region either via the keyboard shortcut or the menu, it starts normally, freezes momentarily and then Greenshot crashes with no visible error messages. Thank you.

### Additional context

1. Make sure Greenshot is running
2. Right click Greenshot icon and select Capture Region
3. Start to capture a portion of the screen (browser window and Word in my case)
4. App crashes after freeze

danrhodes:
@AnadarProSvcs should be merged to a prerelease for testing soon.

Issue #1019, opened 2026-03-02
Crash on Capture Region


bt1002:
**Describe the bug**
crash on capture region

**To Reproduce**
Example steps to reproduce the behavior:
1. Capture region
2. try to select open in image editor
3. GUI hangs
4. randomly crashes (inconsistent)

**Expected behavior**
allow me to select how to open/edit

**Screenshots**

<img width="608" height="607" alt="Image" src="https://github.com/user-attachments/assets/f5510feb-90e5-40ba-a321-638a74f8d70b" />

**Versions (please complete the following information):**
 - 1.3.312
 - Microsoft Windows 11 Pro
 - Version	10.0.26200 Build 26200

**Additional context**

- Very inconsistent
- have to restart greenshot and will occasionally work again for a screenshot or two
- reboot does not help

danrhodes:
Are you able to test a pre-release to see if this issue is already resolved, as the community have made many changes since? Download links here - https://github.com/greenshot/greenshot/releases

Lakritzator:
I think this is the issue with Windows notifications, try disabling the notify in the capture settings

bt1002:
> I think this is the issue with Windows notifications, try disabling the notify in the capture settings

Thanks,  notifications seem to do it.  haven had crashes  since I turned it off

<img width="393" height="434" alt="Image" src="https://github.com/user-attachments/assets/fe8a38c3-53c1-4e7e-b393-74256adc232e" />

Lakritzator:
It's still an annoying bug, and it seems not to be on our side... I'm keeping an eye out for solutions.

Lakritzator:
I tried this today as I think I got an windows update, and seems to work "again".... will check the Windows version when I have access to my PC later.

ronald-cr:
Tryed current release `Greenshot 1.3.312` and pre-release `Greenshot 1.4.134-gd47f990ac7 (continuous build)`.
Using Windows 11 (25H2, OS Build 26200.7840).

I am getting same error. I can make one screenshot (with shortkey or mouse with context menu). After than greenshot crashes and needs to be killed via task manager.

danrhodes:
> Tryed current release `Greenshot 1.3.312` and pre-release `Greenshot 1.4.134-gd47f990ac7 (continuous build)`.
> Using Windows 11 (25H2, OS Build 26200.7840).
> 
[...]

PR #1154, opened 2026-04-08
Fixes for region capture issues

### LINKED ISSUE #1031: Crash when triggering a "capture region" with hotkey
**Describe the bug**
Crash and subusequent error window when triggering a "capture region" with hotkey

**To Reproduce**
1. press "capture region" hotkey (CTRL + SHIFT + S in my case)
2. See Error window appear with "Sorry, an unexpected error occurred..."

**Expected behavior**
No crash

**Screenshots**

Software version: 1.3.300 - c1ad1d4a93 (64 bit)
Time: 2026-03-04 14:04:15 -08:00
OS: Windows 10 2009 x64 build 19045
GDI object count: 96
User object count: 47

Exception: System.ArgumentException
Message: Parameter is not valid.

Stack:
   at System.Drawing.Bitmap..ctor(Int32 width, Int32 height, PixelFormat format)
   at Greenshot.Base.Core.ImageHelper.CreateEmpty(Int32 width, Int32 height, PixelFormat format, Color backgroundColor, Single horizontalResolution, Single verticalResolution)
   at Greenshot.Base.Core.WindowCapture.CaptureRectangle(NativeRect captureBounds)
   at Greenshot.Base.Core.WindowCapture.CaptureRectangle(ICapture capture, NativeRect captureBounds)
   at Greenshot.Base.Core.WindowCapture.CaptureScreen(ICapture capture)
   at Greenshot.Helpers.CaptureHelper.MakeCapture()
   at Greenshot.Helpers.CaptureHelper.CaptureRegion(Boolean captureMouse)
   at Greenshot.Base.Controls.HotkeyControl.HandleMessages(Message& m)
   at Greenshot.Forms.MainForm.WndProc(Message& m)

Configuration dump:

**Additional context**

---
**Stack Trace Hash:** [443fe1b66ea6]

danrhodes:
Fixes #1031 #1153 #1019, which I have grouped together as they are all "region capture" issues.

**Fix 1 — ArgumentException in Bitmap..ctor: degenerate DIB dimensions**
**File: WindowCapture.cs:357-363**

**What was wrong**: Inside the offscreenContent branch, Image.FromHbitmap(safeDibSectionHandle.DangerousGetHandle()) was called to get tmpBitmap, then its .Width and .Height were passed directly to ImageHelper.CreateEmpty. Under two known conditions — a multi-monitor setup where the virtual desktop rectangle spans monitors at mixed DPI scales, or a transient GDI resource exhaustion (session lock/unlock, high GDI object counts) — the DIB section handle can produce a bitmap with zero or invalid dimensions. new Bitmap(0, height, format) throws ArgumentException: Parameter is not valid, which was unhandled and crashed the capture.

**What was fixed**: A guard was added immediately after tmpBitmap is loaded. If either dimension is <= 0, we log the degenerate state, fall back to returning the raw DIB bitmap (same path as the non-offscreen branch), and break out successfully. The transparent compositing is skipped in this edge case — a minor visual trade-off (no transparency in inter-monitor gaps) versus a hard crash.

Additionally, the retry loop's catch block was widened to also catch ArgumentException. Previously, only ExternalException was retried; any ArgumentException escaping the offscreenContent block went straight to the unhandled exception handler. Now it's wrapped and treated as a retryable failure, giving the loop its three attempts before giving up.

**Fix 2 — ArgumentNullException in Monitor.Enter: race between Dispose and background thread
CaptureHelper.cs:96**

**What was wrong**: MakeCapture starts a background thread (RetrieveWindowDetails) that uses lock (_windows) to safely add entries to the list. Dispose — called via the using block when MakeCapture returns or throws — was setting _windows = null. If Dispose ran between the moment the background thread read _windows as a non-null reference and the moment it entered the lock, Monitor.Enter received null and threw ArgumentNullException. This is a classic TOCTOU race: the thread sees a non-null _windows, then Dispose nulls it, then the thread calls lock on the now-null reference.

The Join() call at the end of MakeCapture was supposed to prevent this — but if MakeCapture exited via an exception (e.g., Crash A happening first), the Join() still ran, but an exception from CaptureWithFeedback or CaptureScreen could skip execution flow in ways that cause Dispose to fire before the join completes cleanly. The two crash entries from the same user 33 seconds apart (23:50:42 and 23:51:15) confirm this sequence: Crash A fires first, then Crash B follows as the teardown races the background thread.

**What was fixed**: _windows = null was removed from Dispose. The List<WindowDetails> is not an unmanaged resource — it doesn't need explicit nulling for correctness. The background thread's lifetime is already managed by the Join() in MakeCapture. Leaving _windows alive until the GC collects CaptureHelper is entirely safe, and the comment explains why to prevent the null-on-dispose pattern from being reintroduced.

**Fix 3 — ArgumentException defensive guard in ImageHelper.CreateEmpty
ImageHelper.cs:1360-1363**

**What was wrong**: CreateEmpty passed width and height straight to new Bitmap(...) with no validation. System.Drawing.Bitmap accepts only positive dimensions; anything <= 0 throws ArgumentException: Parameter is not valid with no indication of what the invalid values actually were. Since CreateEmpty is called from multiple sites across the codebase (torn edges, shadows, borders, resize, rotation), any caller passing bad geometry produced the same opaque exception with no diagnostic information.

**What was fixed**: An explicit guard was added at the top of CreateEmpty. If either dimension is non-positive, it throws ArgumentException with the actual values in the message (Cannot create a bitmap with non-positive dimensions (0x480)). This means:

Any future call site that passes bad dimensions gets a clear, diagnosable error message in the log instead of a cryptic GDI+ failure.

**Fix 1's path now benefits doubly**: if the degenerate-dimension guard in WindowCapture were ever bypassed (e.g. a new call site added in future), the error would still surface with a useful message and be caught by the widened retry loop rather than crashing the process.


src/Greenshot.Base/Core/WindowCapture.cs
```diff
@@ -354,6 +354,13 @@ public static Bitmap CaptureRectangle(NativeRect captureBounds)
                         if (offscreenContent)
                         {
                             using Bitmap tmpBitmap = Image.FromHbitmap(safeDibSectionHandle.DangerousGetHandle());
+                            if (tmpBitmap.Width <= 0 || tmpBitmap.Height <= 0)
+                            {
+                                Log.Warn($"DIB section returned degenerate bitmap dimensions ({tmpBitmap.Width}x{tmpBitmap.Height}), skipping offscreen capture.");
+                                returnBitmap = Image.FromHbitmap(safeDibSectionHandle.DangerousGetHandle());
```
Lakritzator: Why not take the tmpBitmap created on line 356 to return (remove the using keyword) and after the if/break use a using block? Prevents 2x Image.FromHbitmap?
So:

```
using (tmpBitmap) {
    returnBitmap = ImageHelper.CreateEmpty(tmpBitmap.Width, tmpBitmap.Height, PixelFormat.Format32bppArgb, Color.Transparent, tmpBitmap.HorizontalResolution, tmpBitmap.VerticalResolution);
}
```

Lakritzator:
Again very impressing work ❤️

danrhodes:
> Again very impressing work ❤️

Thanks,

Happy to help where I can 👍

danrhodes:
@Lakritzator - Good catch on the double Image.FromHbitmap call — updated to remove the using from the declaration and reuse tmpBitmap directly as returnBitmap in the early-exit path, with the using block wrapping the CreateEmpty + draw loop instead.

While looking at this, I noticed the else branch (non-offscreen path, line 384) was calling Image.FromHbitmap with no dimension check at all — so a degenerate DIB there would silently return a 0×0 bitmap to the caller and crash outside the retry loop. The if branch got the guard, but the else was missed, so I've added a matching check there too.

Lakritzator (approved):

PR #1162, opened 2026-04-13
Added a new bug report with hashing and github searching

Lakritzator:
This is a new bug report, looking more modern and it uses the hashing I introduced with https://github.com/greenshot/greenshot/blob/main/.github/workflows/issue-hash.yml

Something like this (minus the wpf visual tree debugger):
image


danrhodes:
You are a legend hashing will make bug identification SO much easier! ⭐️⭐️⭐️⭐️⭐️

Can it check against git and auto submit a new bug if it doesn't exist or is that just going to cause too many submissions? 🙈

What's method is generating the hash can we break out the code into a hash generator and go over the existing issues and add hashes to the comments? Could maybe do this automagically.

Christian-Schulz:
Love it ❤️,

I saw you added a hash on all Issues.  This helps to find duplicates like : https://github.com/search?q=repo%3Agreenshot%2Fgreenshot+c4ed667e628c&type=issues

How are GitHub Actions supposed to work, and when do they run?

danrhodes:
image

Very impressive 🚀❤️

danrhodes:
Could it auto link to other issues given the stack traces this changes this dramatically 🤯👀

Lakritzator:
> Could it auto link to other issues given the stack traces this changes this dramatically 🤯👀

Yes, but I implemented this only for new issues, it would have been to many API calls to do so for all current issues, although I might recheck that in the future. This still needs a bit of testing...

Lakritzator:
> How are GitHub Actions supposed to work, and when do they run?

This one runs on every open / edit: https://github.com/greenshot/greenshot/blob/main/.github/workflows/issue-hash.yml
It's the more complex one, as it checks against others.

The other was made to do a batch of ~100 issues, and changes the initial comment to have the hash & adds a label: https://github.com/greenshot/greenshot/blob/main/.github/workflows/issue-hash-batch.yml

Christian-Schulz:
The old dialog displayed a detailed message for the user. The new dialog should do the same.
> "Sorry, an unexpected error occurred.
> 
> The good news is: you can help us getting rid of it by filing a bug report.
> Please visit the URL below, create a new bug report and paste the contents from the text area into the description.
> 
> Please add a meaningful summary and enclose any information you consider to be helpful for reproducing the issue.
> Also, we would highly appreciate if you checked whether a tracker item already exists for this bug. (You can use the search to find those quickly.) Thank you :)"

Lakritzator:
Reminder to also look at what the hashing is doing, why does this happen: #1186

Lakritzator:
Actually.. I had a very quick peek, although not much time, I didn't finish the code, and seem to have overseen what Gemini created as a skeleton. It takes the "Exception: " line and adds the "at Greenshot" and "at Dapplo" lines... so doesn't add the rest of the stacktrace, like "at System.Windows.Forms.ToolStripDropDown". It even doesn't consider if someone adds 2 exception stack traces, so I would assume it takes stuff double.

I was so focused on getting the base functionality working, so trigger on new issues, the correct tagging, having a bulk process etc, I should have tested the exception extraction.

But I designed it to be such a way that we can replace the current hashes, so I will fix that.

Christian-Schulz:
It's my fault too; I did read the YAML, but I didn't review the normalization of the exception text either.

Maybe make it a bit simpler an only us the last `^Exception:` and the last `^Configuration dump:` as stop words.

Lakritzator:
Or just add two hashes for both of they differ

jklingen:
Wow, just noticed this - awesome idea @Lakritzator 

> The old dialog displayed a detailed message for the user. The new dialog should do the same.

To be honest, I am not sure about that. I had already been thinking (before I saw this) about whether we should make the text "less inviting" to just copy/paste the stack trace, or simply remove text and link entirely. (Hoping to reduce the number of duplicates and tickets missingadditional information like reproduction steps. It seems that many people don't read the 2nd paragraph of the message - I assume that these also wouldn't try the search button.)

Lakritzator:
Closing this in favor of work in #784

Issue #1157, opened 2026-04-13
Error System IO FileNotFundException


lindseynickerson:
### Please confirm the following before submitting a Bug Report

- [x] I have searched existing open and closed issues and did not find a duplicate

### Greenshot version

5.0

### Describe the bug

Exception: System.IO.FileNotFoundException
Message: Could not load file or assembly 'System.Reactive, Version=5.0.0.0, Culture=neutral, PublicKeyToken=94bc3704cddfc263' or one of its dependencies. The system cannot find the file specified.

Stack:
   at Dapplo.Windows.Messages.WinProcHandler.Subscribe(WinProcHandlerHook winProcHandlerHook)
   at Dapplo.Windows.User32.DisplayInfo.get_AllDisplayInfos() in D:\a\1\s\src\Dapplo.Windows.User32\DisplayInfo.cs:line 127
   at Dapplo.Windows.User32.DisplayInfo.get_ScreenBounds() in D:\a\1\s\src\Dapplo.Windows.User32\DisplayInfo.cs:line 77
   at Greenshot.Base.Core.Capture..ctor()
   at Greenshot.Helpers.CaptureHelper..ctor(CaptureMode captureMode)
   at Greenshot.Helpers.CaptureHelper.CaptureRegion(Boolean captureMouse)
   at Greenshot.Base.Controls.HotkeyControl.HandleMessages(Message& m)
   at Greenshot.Forms.MainForm.WndProc(Message& m)
   at System.Windows.Forms.NativeWindow.Callback(IntPtr hWnd, Int32 msg, IntPtr wparam, IntPtr lparam)

Configuration dump:

---
**Stack Trace Hash:** `b78c88a7027d` [stack-trace-hash: b78c88a7027d]

danrhodes:
Can you tell us how you installed GS where did you get the installer from?

Issue #1100, opened 2026-03-23
unexpected error when pressing Print key


RobertoAmazonas:
### Please confirm the following before submitting a Bug Report

- [x] I have searched existing open and closed issues and did not find a duplicate

### Greenshot version

1.3.312

### Describe the bug

Software version: 1.3.312 - 5a60aff83e (64 bit)
.NET runtime version: 4.0.30319.42000+
Time: 2026-03-23 08:15:11 +01:00
OS: Windows 10 2009 x64 build 26200
GDI object count: 56
User object count: 18

Exception: System.IO.FileNotFoundException
Message: Die Datei oder Assembly "System.Reactive, Version=5.0.0.0, Culture=neutral, PublicKeyToken=94bc3704cddfc263" oder eine Abhängigkeit davon wurde nicht gefunden. Das System kann die angegebene Datei nicht finden.

Stack:
   at Dapplo.Windows.Messages.WinProcHandler.Subscribe(WinProcHandlerHook winProcHandlerHook)
   at Dapplo.Windows.User32.DisplayInfo.get_AllDisplayInfos() in D:\a\1\s\src\Dapplo.Windows.User32\DisplayInfo.cs:line 127
   at Dapplo.Windows.User32.DisplayInfo.get_ScreenBounds() in D:\a\1\s\src\Dapplo.Windows.User32\DisplayInfo.cs:line 77
   at Greenshot.Base.Core.Capture..ctor()
   at Greenshot.Helpers.CaptureHelper..ctor(CaptureMode captureMode)
   at Greenshot.Helpers.CaptureHelper.CaptureRegion(Boolean captureMouse)
   at Greenshot.Base.Controls.HotkeyControl.HandleMessages(Message& m)
   at Greenshot.Forms.MainForm.WndProc(Message& m)
   at System.Windows.Forms.NativeWindow.Callback(IntPtr hWnd, Int32 msg, IntPtr wparam, IntPtr lparam)

Configuration dump:

---
**Stack Trace Hash:** `b78c88a7027d` [stack-trace-hash: b78c88a7027d]

Issue #836, opened 2026-01-19
Green shot is not working. Unable to take a screen shot


JPolavarapu:
**Describe the bug**
Green shot is not working. Unable to take a screen shot

**To Reproduce**
Example steps to reproduce the behavior:
1. Capture '....'
2. Open Editor
3. Add a line
4. Resize

**Expected behavior**
Software version: 1.3.304 - f46c1b7f31 (64 bit)
.NET runtime version: 4.0.30319.42000+
Time: 2026-01-19 08:21:55 -05:00
OS: Windows 10 2009 x64 build 22631
GDI object count: 59
User object count: 21

Exception: System.IO.FileNotFoundException
Message: Could not load file or assembly 'System.Reactive, Version=5.0.0.0, Culture=neutral, PublicKeyToken=94bc3704cddfc263' or one of its dependencies. The system cannot find the file specified.

Stack:
   at Dapplo.Windows.Messages.WinProcHandler.Subscribe(WinProcHandlerHook winProcHandlerHook)
   at Dapplo.Windows.User32.DisplayInfo.get_AllDisplayInfos() in D:\a\1\s\src\Dapplo.Windows.User32\DisplayInfo.cs:line 127
   at Dapplo.Windows.User32.DisplayInfo.get_ScreenBounds() in D:\a\1\s\src\Dapplo.Windows.User32\DisplayInfo.cs:line 77
   at Greenshot.Base.Core.Capture..ctor()
   at Greenshot.Helpers.CaptureHelper..ctor(CaptureMode captureMode)
   at Greenshot.Helpers.CaptureHelper.CaptureRegion(Boolean captureMouse)
   at Greenshot.Base.Controls.HotkeyControl.HandleMessages(Message& m)
   at Greenshot.Forms.MainForm.WndProc(Message& m)
   at System.Windows.Forms.NativeWindow.Callback(IntPtr hWnd, Int32 msg, IntPtr wparam, IntPtr lparam)

Configuration dump:

**Screenshots**

**Versions (please complete the following information):**
 - Greenshot version
 - Windows version

**Additional context**

---
**Stack Trace Hash:** [b78c88a7027d]

Lakritzator:
It looks like an installation which was screwed up. We see many companies use their own way of installing, which is their right, but sometimes the ones packaging forget to take the changes. We change the contents of our installation over time, which means some files, which represent functionality we use, are added others are removed.

How was Greenshot installed, did you download it from our website and install it via the installer on your own PC?
[...]

PR #1164, opened 2026-04-14
An idea to improve bug tracking

danrhodes:
Users who encounter a missing dependency error — typically because they extracted Greenshot from a zip or copied files manually rather than using the official installer — currently receive the same generic crash dialog as someone who hit a real bug. That dialog tells them to file a bug report, which is the wrong action entirely: there is no bug to report, the installation is simply incomplete. This creates noise on the issue tracker (maintainers spend time triaging reports that are not defects), and it leaves the user with no actionable guidance at a moment when they just want to know why the tool won't start. The change is small and entirely additive — it detects a specific, well-defined exception type (FileNotFoundException with "Could not load file or assembly" in the message), routes it to a separate dialog that names the missing file, explains the likely cause in plain language, and offers a direct link to the official installer. The generic bug reporter is untouched and handles everything else exactly as before. The cost of merging this is low; the benefit is that a frustrated user gets a clear answer instead of a confusing one.

May have stopped #1157 #1100 and #836 plus others also?

### DIFF
--- a/src/Greenshot/Configuration/LanguageKeys.cs
+++ b/src/Greenshot/Configuration/LanguageKeys.cs
@@ -77,6 +77,12 @@ namespace Greenshot.Configuration
         tooltip_firststart,
         warning,
         warning_hotkeys,
-        update_found
+        update_found,
+        missing_dependency_title,
+        missing_dependency_info,
+        missing_dependency_assembly_label,
+        missing_dependency_download,
+        missing_dependency_show_details,
+        missing_dependency_hide_details
     }
 }
--- a/src/Greenshot/Forms/MissingDependencyForm.Designer.cs
+++ b/src/Greenshot/Forms/MissingDependencyForm.Designer.cs
@@ -1,0 +1,164 @@
+/*
+ * Greenshot - a free and open source screenshot tool
+ * Copyright (C) 2004-2026 Thomas Braun, Jens Klingen, Robin Krom
+ *
+ * For more information see: https://getgreenshot.org/
+ * The Greenshot project is hosted on GitHub https://github.com/greenshot/greenshot
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 1 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+using Greenshot.Base.Controls;
+
+namespace Greenshot.Forms
+{
+    partial class MissingDependencyForm
+    {
+        private System.ComponentModel.IContainer components = null;
+
+        protected override void Dispose(bool disposing)
+        {
+            if (disposing)
+            {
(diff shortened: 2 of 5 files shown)


danrhodes:
I've put this up asan  idea, if people like i, Ii will re-review the Gemini comments 👍

Lakritzator:
Generally it might make sense...
How much certainty is there, that this is the case, and it's not a bad installation done by our installer?

Actually the checksum.SHA256 could come to the rescue, it contains (at least should) all the files with a checksum, we could present a list of all the files in there and check if they are missing and validate if these are correct.

I did notice it is missing in the portable .zip, so I will check why... (just checked, there is much more which doesn't make sense in the portable)

Lakritzator:
Maybe also make it a commandline option, to validate the installation, maybe get that working without that greenshot.exe needs to load any DLLs..

danrhodes:
I've switched off now but I'll do some thinking about this until I can get back on. Time is tight for the next few weeks due to childcare 👍

Lakritzator:
No problem, family first!!!!

I've added a PR to fix the portable and something else: #1166

Lakritzator:
Here is the code which uses the file and validates the files against the hash:

```
using System;
using System.Collections.Generic;
using System.IO;
using System.Security.Cryptography;
using System.Text;
using log4net;

namespace Greenshot.Base.Core
{
    public static class ChecksumValidator
    {
        private static readonly ILog LOG = LogManager.GetLogger(typeof(ChecksumValidator));

        /// <summary>
        /// Validates the installation against a checksum file.
        /// </summary>
        /// <param name="checksumFilePath">Path to the .SHA256 file.</param>
        /// <returns>True if all files match, false otherwise.</returns>
        public static bool ValidateInstallation(string checksumFilePath)
        {
            if (!File.Exists(checksumFilePath))
            {
                LOG.Warn($"Checksum file not found: {checksumFilePath}");
                return false;
            }

            string baseDir = Path.GetDirectoryName(checksumFilePath);
            bool allOk = true;

            try
            {
                foreach (string line in File.ReadLines(checksumFilePath))
                {
                    if (string.IsNullOrWhiteSpace(line)) continue;

                    // Standard sha256sum format: [hash]  [relative_path]
                    // Note: We use two spaces as the separator.
                    string[] parts = line.Split(new[] { "  " }, 2, StringSplitOptions.RemoveEmptyEntries);
                    if (parts.Length != 2) continue;

                    string expectedHash = parts[0].Trim();
                    string relativePath = parts[1].Trim().Replace('/', Path.DirectorySeparatorChar);
                    string fullPath = Path.Combine(baseDir, relativePath);

                    if (!File.Exists(fullPath))
                    {
                        LOG.Error($"Missing file: {relativePath}");
                        allOk = false;
                        continue;
                    }

                    string actualHash = GetFileHash(fullPath);
                    if (!string.Equals(expectedHash, actualHash, StringComparison.OrdinalIgnoreCase))
                    {
                        LOG.Error($"Hash mismatch for {relativePath}! Expected: {expectedHash}, Actual: {actualHash}");
                        allOk = false;
                    }
                }
            }
            catch (Exception ex)
            {
                LOG.Error("Error during checksum validation", ex);
                return false;
            }

            if (allOk) LOG.Info("Installation checksum validation successful.");
            return allOk;
        }

        private static string GetFileHash(string filePath)
        {
            using (var sha256 = SHA256.Create())
            {
                using (var stream = File.OpenRead(filePath))
                {
                    byte[] hashBytes = sha256.ComputeHash(stream);
                    StringBuilder sb = new StringBuilder();
                    foreach (byte b in hashBytes)
                    {
                        sb.Append(b.ToString("x2"));
                    }
                    return sb.ToString();
                }
            }
        }
    }
}
```

PR #1165, opened 2026-04-14
Update AbstractDestination.cs

### LINKED ISSUE #1163: Pasting to Excel when trying to send direct to printer
### Please confirm the following before submitting a Bug Report

### Greenshot version

1.3.315

### Describe the bug

I was attempting to send a screenshot directly to the printer, but it opened Excel and pasted in there. On subsequent attempts it just gave an error.

Software version: 1.3.315 - 420b08b2a9 (64 bit)
Time: 2026-04-14 10:30:43 -05:00
OS: Windows 11 x64 build 26100
GDI object count: 101
User object count: 44

Exception: System.NullReferenceException
Message: Object reference not set to an instance of an object.

Stack:
   at Greenshot.Base.Core.AbstractDestination.<>c__DisplayClass28_0.<ShowPickerMenu>b__3(Object <p0>, EventArgs <p1>)
   at System.Windows.Forms.ToolStripItem.RaiseEvent(Object key, EventArgs e)
   at System.Windows.Forms.ToolStripMenuItem.OnClick(EventArgs e)
   at System.Windows.Forms.ToolStripItem.HandleClick(EventArgs e)
   at System.Windows.Forms.ToolStripItem.HandleMouseUp(MouseEventArgs e)
   at System.Windows.Forms.ToolStrip.OnMouseUp(MouseEventArgs mea)
   at System.Windows.Forms.ToolStripDropDown.OnMouseUp(MouseEventArgs mea)
   at System.Windows.Forms.Control.WmMouseUp(Message& m, MouseButtons button, Int32 clicks)

Configuration dump:

### Additional context

_No response_

---
**Stack Trace Hash:** [1094b529e383]

danrhodes:
Root cause: The ShowPickerMenu method captures surface in a closure shared by all the destination click handlers and the menu close handler. After a successful export, surface.Dispose() is called and surface is set to null. If the click handler fires again (e.g. due to event handler re-entrancy, a double-click, or WndProc processing a queued mouse-up after the menu was already closed), it calls clickedDestination.ExportCapture(true, null, captureDetails) — passing null as the surface — which causes the NullReferenceException.

Fixes made in AbstractDestination.cs:

Line ~268 — Added a null guard in the destination click handler: if surface has already been disposed and set to null, log a warning and return early instead of proceeding with a null surface.

Line ~230 — Added a null check in the Keyboard close handler before calling surface.Dispose(), since the same closure variable could already be null if export succeeded before ESC was pressed.

The "opened Excel" symptom on the first attempt is likely a separate usability issue (user may have hovered over Excel's submenu, which mutated its Tag before clicking Printer), but without the user being able to reproduce it consistently it may be incidental. The crash on subsequent attempts is definitively this null surface bug.

Fixes #1163

### DIFF
--- a/src/Greenshot.Base/Core/AbstractDestination.cs
+++ b/src/Greenshot.Base/Core/AbstractDestination.cs
@@ -225,11 +225,11 @@ namespace Greenshot.Base.Core
                         break;
                     case ToolStripDropDownCloseReason.Keyboard:
                         // Menu closed via keyboard (e.g., ESC key)
-                        if (!captureDetails.HasDestination("Editor"))
+                        if (!captureDetails.HasDestination("Editor") && surface != null)
                         {
                             surface.Dispose();
                             surface = null;
-                        } 
+                        }
                         // We might already be in the disposing process, so queue the disposal to avoid re-entrancy
                         menu.BeginInvoke(new Action(() =>
                         {
@@ -262,6 +262,13 @@ namespace Greenshot.Base.Core
                         IDestination clickedDestination = (IDestination) toolStripMenuItem?.Tag;
                         if (clickedDestination == null)
                         {
+                            return;
+                        }
+
+                        // Guard against re-entrant clicks after the surface has already been disposed
+                        if (surface == null)
+                        {
+                            Log.Warn("Destination click handler invoked after surface was already disposed; ignoring.");
                             return;
                         }
 


Lakritzator (approved):

Lakritzator:
The things you find... 🤓

danrhodes:
> The things you find... 🤓

Geek and proud!

PR #1212, opened 2026-05-05
Update ImageIO.cs

### LINKED ISSUE #1124: Saving screen shot to a .jpg file that already exists
### Please confirm the following before submitting a Bug Report

### Greenshot version

1.3.315-420b-08

### Describe the bug

Software version: 1.3.315 - 420b08b2a9 (64 bit)
Time: 2026-03-27 20:24:04 -04:00
OS: Windows 11 x64 build 26200
GDI object count: 132
User object count: 67

Exception: System.UnauthorizedAccessException
Message: Access to the path is denied.

Stack:
   at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
   at System.IO.FileStream.WriteCore(Byte[] buffer, Int32 offset, Int32 count)
   at System.IO.FileStream.FlushWrite(Boolean calledFromFinalizer)
   at System.IO.FileStream.Dispose(Boolean disposing)
   at System.IO.Stream.Close()
   at Greenshot.Base.Core.ImageIO.Save(ISurface surface, String fullPath, Boolean allowOverwrite, SurfaceOutputSettings outputSettings, Boolean copyPathToClipboard)
   at Greenshot.Base.Core.ImageIO.SaveWithDialog(ISurface surface, ICaptureDetails captureDetails)
   at Greenshot.Destinations.FileWithDialogDestination.ExportCapture(Boolean manuallyInitiated, ISurface surface, ICaptureDetails captureDetails)
   at Greenshot.Base.Core.AbstractDestination.<>c__DisplayClass28_0.<ShowPickerMenu>b__4(Object sender, EventArgs e)
   at System.Windows.Forms.ToolStripItem.RaiseEvent(Object key, EventArgs e)
   at System.Windows.Forms.ToolStripMenuItem.OnClick(EventArgs e)
   at System.Windows.Forms.ToolStripItem.HandleClick(EventArgs e)
   at System.Windows.Forms.ToolStripItem.HandleMouseUp(MouseEventArgs e)
   at System.Windows.Forms.ToolStrip.OnMouseUp(MouseEventArgs mea)
   at System.Windows.Forms.ToolStripDropDown.OnMouseUp(MouseEventArgs mea)
   at System.Windows.Forms.Control.WmMouseUp(Message& m, MouseButtons button, Int32 clicks)

Configuration dump:

### Additional context

_No response_

danrhodes:
Root cause: UnauthorizedAccessException thrown from FileStream.FlushWrite during Dispose (i.e. at the closing } of the using block in ImageIO.cs:323). This happens when the target file already exists and has the read-only attribute set. This is a Windows behaviour — Windows lets you open and truncate a read-only file with FileMode.Create, but denies the final flush when the stream is closed, enforcing the read-only flag at flush/write-commit time rather than at open time.

Fix (ImageIO.cs:314): Before opening the FileStream, check if the existing file is read-only and strip that attribute. The existing allowOverwrite check is preserved — it now sits inside the File.Exists block, so it only runs when the file actually exists. The read-only strip only happens when overwriting is permitted.

Fixes #1124


danrhodes:
Radio silence recently due to other commitments, but made pr's for these, I will review Gemini tomorrow. Night all.

Lakritzator:
I am generally not a fan of overwriting readonly files, especially without asking removing the attribute. I would actually call it misbehavior, I am currently not in favor but I might misunderstand this. When does this even happen?

Can't we just show a clear message, and ask the user to write to another file?

Lakritzator:
I had another look at this, and sorry but I cannot find this correct. We should respect a read only attribute, remove the attribute change and make the error message clear enough that the user understand the issue, using language resources, and I will merge it.

Lakritzator:
@danrhodes I will close this, I would not like it if any normal desktop application changes the read-only attribute on my files, without at least having a clear "I'm going to do 'this', are you okay?", and I don't think Greenshot is for that. We should clearly specify why we cannot save the file, yes... Additionally if we would do this, the effect afterwards is that the file is no longer read-only, but it was read-only because of a "reason".

Hope you are okay with that...

Issue #387, opened 2022-02-23
Add noise to obfuscate-tool


viggotw:
Add noise to the pixelation to avoid vulnerabilities such as [this](https://github.com/BishopFox/unredacter), where a user is able to reconstruct redacted, and possibly sensitive, information.

Lakritzator:
That is funny, I've been thinking about that issue (in general) a couple of days ago.

It might seem very straightforward to do so, but when you think about it, it's actually not soooo easy.

- If I add noise, meaning a random number, you will see the area change with every movement.
- You will never get the WYSIWYG (what you see is what you get) as the image is rendered again when you save.
- Although one can reconstruct text, it does look cooler than "random" pixels.

Anyway, I could live with that it doesn't look as good or it moves, just want to check out what is possible.

Lakritzator:
Actually there was an issue reported before, #248, which made me think about some extra tricks, here OCR as mentioned in #383 might come to the rescue (guess the font size and preset the pixel size big enough)

Christian-Schulz:
How about adding a random noise layer for the entire image that does not move? Each pixelate then uses the same noise layer.

thomasdn:
I second @Christian-Schulz suggestion to generate a static noise layer for the entire image. Such layer could be generated by using a hash of the original image as seed. 
In fact, a static noise layer could be generated once and for all and distributed with the Greenshot binary. It could be something as simple as a random noise layer of, say, 500x500 px that is then just repeated/tiled over the image to match the actual image dimensions.

Lakritzator:
This should be fixed with Greenshot v1.4.192 and later.

PR #1259, opened 2026-06-21
Prevent reversing our pixelate

Lakritzator:
This changes the way the pixelate filter works, and makes the changes via a Cryptography random, prohibiting that the effect is reversed via ML or AI.


Lakritzator:
This addresses #387 and should make it impossible to reverse the pixelate.

PR #622, opened 2025-07-16
Combine multiple grayscale highlights

Kissaki:
## The Premise

Highlight in Grayscale mode grays everything out except for the highlight area. This works fine for a single such highlight.

image

However, when multiple highlights in Grayscale mode are added, they cancel each other out,

image

while overlaps do show up.

image

## The Solution

Instead of applying the grayscale filters one after the other individually, we combine their clipping region and highlight exclusions.

image

## The Implementation

The existing code architecture is designed with independent elements and filters in mind. Combining filters/filter drawing breaks this established code structure, resulting in a somewhat messy implementation crossing the otherwise clear borders.

The implementation does try to structure logic into methods and adds comments where possible, to reduce negative code impact.

This change solves an issue or shortcoming (depending on interpretation). The question is whether this code burden is worth that. Practically speaking, the burden doesn't seem particularly big - the concerns are still relatively isolated.

---

Resolves #329

Please share your thoughts, or whether there are other, better approaches to implementing this in the context of this codebase.

PR #1350, opened 2026-09-06
Single draw for inverted filters with multiple areas

Lakritzator:
This change makes it possible to draw the HighlightFilter or GrayscaleFilter only once for multiple selected areas, as #622 and #512 supposed to do, but in a single change looking at both requirements.

I really do not want to take away the great work that @FF-Brown and @Kissaki did on their PRs, but I wanted a single solution for the 2 filters. This is what I came up with.

Co-authored-by: @Kissaki 
Co-authored-by: @FF-Brown


FF-Brown:
Thanks for doing the work to merge those together and make a better fix 😁

PR #1352, opened 2026-09-07
Experimental: This PR adds recipes, a way to specify the flow of a capture from trigger to result

Lakritzator:
This is an experiment, and only available if the user has manually set IsBetaTester to true.
However, to make this possible, a lot of the core functionality needed to be refactored, as this was a not so tasteful spaghetti code.

This means there might be side effects, even if the user doesn't set IsBetaTester.
**Due to the nature of the changes, this is much needed to bring Greenshot forward.**

I want to do some more testing, to see if more legacy functions will still work before merging this.


Lakritzator:
@copilot add the translations for all the other languages

Lakritzator:
@Christian-Schulz @danrhodes you as heavy users, maybe have a look at this... 
Somewhat technically heavy, but I trust you can understand most of the idea and the usage:
https://github.com/greenshot/greenshot/blob/main/docs/capture-recipes.md

I have some additional ideas to extend this concept, but I needed a good base for it... was a risky change, there will be bugs. Available with 1.4.211 and later.

danrhodes:
I will install tomorrow and feedback 👍

Christian-Schulz:
Huge , this would solve many different feature requests. I will look at it.

Lakritzator:
This feature should make it possible to completely configure how you want Greenshot to handle something, and later share this with others via a to be build github page. Some building blocks, steps, are still missing, the current state is an experiment to see what we need.

This feature is "protected" by the IsBetaTester setting in the greenshot.ini
If set to true, you will have a new option in the context menu to import an recipe file, when doing so an approval window appears.

Btw. There is no way to remove a recipe yet, neither something like debugging.

FF-Brown:
@Lakritzator This sounds amazing! I can't wait to see it in action!

Christian-Schulz:
> Btw. There is no way to remove a recipe yet, neither something like debugging.

I should have read that sooner. Had spent 20 minutes looking for it. 🤦‍♂️

* I’m not sure that importing the .gsrecipe.json files by reference is the right choice. 
Whenever I import something into an app, I would expect the app makes a copy of it.

* Am I right about that? The main critical changes are in `CaptureHelper` that uses the new recipes under the hood.

* Is it possible to extend the recipes that a step could add an Container like a rectangle, an arrow or a Image?

Lakritzator:
> > Btw. There is no way to remove a recipe yet, neither something like debugging.
> I should have read that sooner. Had spent 20 minutes looking for it. 🤦‍♂️

Sorry ;)
 
> * I’m not sure that importing the .gsrecipe.json files by reference is the right choice.
>   Whenever I import something into an app, I would expect the app makes a copy of it.

Hmm, yeah it would be bad to keep them inside a download folder or something, good though, needs clearification.

> * Am I right about that? The main critical changes are in `CaptureHelper` that uses the new recipes under the hood.
Yeah, that's a very critical part, as I replaced the spaghetti code which did the whole "capture flow" with recipes, build the original flow more or less in a new code, and already had some bugs (printscreen with region only captured 1 display)... so critical!

> * Is it possible to extend the recipes that a step could add an Container like a rectangle, an arrow or a Image?

Definitively, that's one of the goals, but before I can do that, I need to kill the Capture/CaptureDetails/Surface mess I have, so I left it for now. Then the impact would be even bigger, maybe if we find it stable enough, I might have a look.

PR #1362, opened 2026-09-09
Fix for loading .greenshot file

Christian-Schulz:
Since v1.4.209 greenshot files have not been loading properly. Greenshot loads it only as image.

Before loading the file we need an instance of a `ISurface`.  But only the `CapturePayload` can create it internally and `CapturePayload` needs an Image.
So i added a hack wich loads the file twice.

I know that the main problem is that the Surface is defined in the Greenshot.Editor project and it is a Windows.Forms control.
And Greenshot.Base should not reference Greenshot.Editor.

In #638  i had the same problem and created an internal domain representation [GreenshotFile](https://github.com/Christian-Schulz/greenshot/blob/feature/new_greenshot_file_format_v2/src/Greenshot.Editor/FileFormat/GreenshotFile.cs) for all data from a Surface.
After loading the file into the domain object it is converted to a Surface [here](https://github.com/Christian-Schulz/greenshot/blob/413471def47c25492e2e6e7d5460b16cb1ae4deb/src/Greenshot.Editor/FileFormat/GreenshotFileVersionHandler.cs#L205)

Such similar domain object could be replace the ISurface in CapturePayload

@Lakritzator what are your thought?

### DIFF
--- a/src/Greenshot.Base/Pipeline/Sources/FileCaptureSource.cs
+++ b/src/Greenshot.Base/Pipeline/Sources/FileCaptureSource.cs
@@ -25,6 +25,7 @@ using System.IO;
 using System.Threading;
 using System.Threading.Tasks;
 using Greenshot.Base.Core;
+using Greenshot.Base.Core.Enums;
 using Greenshot.Base.Interfaces;
 using log4net;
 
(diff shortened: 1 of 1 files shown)


Lakritzator:
Just trying to understand this, as you specify 1.4.209.
Do you mean between Greenshot 1.4.193 and 1.4.209 something changed, therefore it no longer works, or do you mean "as of 1.4.209 this still doesn't work"?

I guess you mean since I introduced the recipes the FileCaptureSource is plainly wrong, I actually forgot to review it (was auto generated by AI). The code needs to be corrected, of course it needs to do it correctly as is and I'll fix it.

Lakritzator:
I'll look at the last commit before it, and fix it with that...
https://github.com/greenshot/greenshot/tree/b631bbcdbd5835c87a8f1fc0861b113d1446a972/src


It is the flow from here: https://github.com/greenshot/greenshot/blob/b631bbcdbd5835c87a8f1fc0861b113d1446a972/src/Greenshot/Helpers/CaptureHelper.cs#L173

And the following part is what was done exactly:
https://github.com/greenshot/greenshot/blob/b631bbcdbd5835c87a8f1fc0861b113d1446a972/src/Greenshot/Helpers/CaptureHelper.cs#L433

To be honest, it should use the IFileFormatHandler logic which would recognize the GreenshotFileFormatHandler can be used, and use this. So it already wasn't the best... Good that we noticed this, it will make maintenance in the future better too.

Christian-Schulz:
> I guess you mean since I introduced the recipes the FileCaptureSource is plainly wrong

Yes, that's exactly what I mean. I only mentioned the version number in case another bug reporter encounters the same problem.

Lakritzator:
#1363 should solve it...

PR #1363, opened 2026-09-10
Different solution for #1362

Lakritzator:


Lakritzator:
This should fix the issue why #1362 was created,

PR #1364, opened 2026-09-12
Recipe phase 2

Lakritzator:
Improved the recipe capability, using a directed graph as foundation, added expressions and make it possible to supply drawables to the capture.

### DIFF
--- a/docs/capture-recipes.md
+++ b/docs/capture-recipes.md
@@ -1,8 +1,8 @@
 # Greenshot Capture Recipes Guide
 
-Greenshot features a modular, recipe-driven capture pipeline. Instead of hardcoded capture flows, every screenshot workflow is an ordered sequence of configurable steps defined as a **Capture Recipe**.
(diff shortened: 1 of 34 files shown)


Lakritzator:
@Christian-Schulz Now my feature is starting to be good... 

> Is it possible to extend the recipes that a step could add an Container like a rectangle, an arrow or a Image?

I would say, with this change, this is more than possible.

Christian-Schulz:
Just read the examples and recipe.schema.json.  I can already see a lot of users writing their own automation scripts.
Tomorrow I'm going to take a much closer look at that.

If you have some tokens left. How about a capture mechanism that could save that as recipes? 😉

Lakritzator:
I'll try to push the editor I have written, it's not finished but I actually need to know what is missing.

> How about a capture mechanism that could save that as recipes?

I'm not sure what you mean.

Lakritzator:
@Christian-Schulz : I now added the editor for the recipes, find it under the recipe menu in build Greenshot 1.4.216 or later.
I know it's not perfect yet, but I think it's ready very awesome.

Christian-Schulz:
> How about a capture mechanism that could save that as recipes?

> I'm not sure what you mean.

I mean, what if you took a screenshot, edited it in the editor, added effects and drawables, and then saved it as a recipe?

Lakritzator:
Got it, technically possible but that won't work very well, as the elements would be positioned for that specific screenshot. Currently we only have fixed coordinates, and no anchor based placement. E.g. 50 pixel left from edge, or 20 px below other element

Lakritzator:
I've added a way to describe exact that in a recipe

PR #1365, opened 2026-09-12
Recipe editor

Lakritzator:


Christian-Schulz:
First impressions.

- Step parameter are not shown properly:
image

- the JSON text ist readonly
- I can't quite confirm that yet, but I have a feeling that the test function doesn't always execute the currently displayed recipe.

Christian-Schulz:
oh, you fixed it 1h ago.

Lakritzator:
@Christian-Schulz I've been VERY active, fixed many issues but don't know if that was fixed 😂
Added user prompts, decisions, more configuration, more options...

There will be issues... but I think this is a feature, which helps to make Greenshot very flexible.
Although currently one needs to understand a lot, I could not imagine my wife to configure such a thing.
I want to be have some good recipes on the website, so community driven.
Then provide a "recipe" download feature, so people can just search and install.

Lakritzator:
I stopped working on it, keeping it stable for a few days, to allow people like you and others, to play around with it, so we can see what is missing. I know that there is no commandline yet, uploads is also very limited, those are somewhat dangerous.

I need to provide some useful examples too, need to think about that. Currently it's a power user thing.

Christian-Schulz:
yeah, i saw it. I played with it the whole day.  The user prompt feature was a good idea.

Christian-Schulz:
A joined flow don't work properly. 
image
The image editor are only shown, if i choose the "With Border" path.

<details>
<summary>example recipe</summary>

```json
{
  "version": "1.0",
  "id": "recipe_46e2e1",
  "name": "New Custom Workflow",
  "description": "Custom visual capture recipe",
  "triggers": [],
  "nodes": [
    {
      "id": "acquire",
      "stepType": "Source",
      "name": "Acquire Region",
      "enabled": true,
      "parameters": {
        "sourceType": "Region",
        "captureMouseCursor": false,
        "preSuppliedRegion": "20,20,20,20"
      }
    },
    {
      "id": "userprompt_1175",
      "stepType": "UserPrompt",
      "name": "UserPrompt",
      "enabled": true,
      "parameters": {
        "title": "Greenshot decision",
        "message": "Please confirm the next step for this capture:",
        "showPreview": true,
        "timeoutSeconds": 0,
        "choices": [
          {
            "key": "With Border",
            "label": "Yes, With Border",
            "style": "Primary",
            "isDefault": true,
            "isCancel": false
          },
          {
            "key": "No",
            "label": "No, Without Border",
            "style": "Secondary",
            "isDefault": false,
            "isCancel": true
          }
        ]
      }
    },
    {
      "id": "effect_d67f",
      "stepType": "Effect",
      "name": "Effect",
      "enabled": true,
      "parameters": {
        "effect": "Border",
        "shadowSize": 10,
        "darkness": 0.6,
        "width": 40,
        "color": "#00FF00"
      }
    },
    {
      "id": "interactiveselection_2005",
      "stepType": "InteractiveSelection",
      "name": "InteractiveSelection",
      "enabled": true,
      "parameters": {
        "selectionMode": "Region",
        "allowWindowSnapping": true
      }
    },
    {
      "id": "editor_a6cd",
      "stepType": "Editor",
      "name": "Editor",
      "enabled": true,
      "parameters": {
        "destination": "Editor"
      }
    }
  ],
  "flow": {
    "startNodes": [
      "acquire"
    ],
    "transitions": {
      "acquire": [
        "interactiveselection_2005"
      ],
      "interactiveselection_2005": [
        "userprompt_1175"
      ],
      "effect_d67f": [
        "editor_a6cd"
      ]
    },
    "conditionalTransitions": [
      {
        "from": "userprompt_1175",
        "branch": "With Border",
        "to": "effect_d67f"
      },
      {
        "from": "userprompt_1175",
        "branch": "No",
        "to": "editor_a6cd"
      }
    ]
  },
  "showInContextMenu": true,
  "isBuiltIn": false,
  "isOverridden": false,
  "filePath": "C:\\Temp\\geenshottemp\\recipes\\userprompt.gsrecipe.json"
}
```
</details>

The editor seen to delete the startNodes list while loading a recipe.json.
image

Lakritzator:
I've made some fixes, "unfortunately" I was already working on other stuff, so I had to include that too...
This means you also now have extension steps available, there will be "issues" with those...

Lakritzator:
@Christian-Schulz Let me know if that recent change solved it for you, or what other issues you notice

Christian-Schulz:
My Issues above are solved.

Next case: I created a flow with two parallel effects and two editor destinations. I think this should be a valid use case.

For a couple milliseconds I could see the output in the first editor. Then the second editor opens. Both effects are applied there and the fist surface disappears. 
image


General question. Do you think the term “recipe” is the right choice? Why not use the term “workflow,” which is more common in my opinion.

Lakritzator:
It's actually not a workflow, or bpmn, but a DAG.

Here is a comparison:
BPMN (Business Process Model and Notation) is a standard for modeling human-centric and long-running business processes, whereas a DAG (Directed Acyclic Graph) is a mathematical structure used in data engineering to model one-directional, task-dependent data pipelines.

Key DifferencesCore Purpose: BPMN coordinates end-to-end business operations across people, systems, and departments. A DAG orchestrates data processing tasks (like ETL jobs in Airflow or Dagster) based on data dependencies. 

Loops and Cycles: BPMN allows loops, meaning a process can go backward to repeat a step or wait for a new event. A DAG is strictly acyclic, meaning it moves forward only and cannot have infinite loops or circular dependencies.

Human Intervention: BPMN easily handles long delays, external messages, and manual human tasks that take days or weeks. DAGs are designed for automated, computational task execution that finishes without human approval. 



But in the end that doesn't matter, I took the name with marketing in my head, sharing recipes sounds way better that sharing a workflow, DAG or bpmn...

Community recipes to get you going...

Lakritzator:
About the bug, that was to be expected, the surface is not something which can be shared. I think I will need to work on a non mutable version of the object, as soon as there is a change a copy is used. Will figure it out

Lakritzator:
I've updated head with fixed for the branching issues: #1370

Christian-Schulz:
My use case with two editor destinations is fixed.

I think it would make sense to move the “Reload recipes” and “Import recipe” menu items from the system tray into the recipe editor window.

PR #1376, opened 2026-09-15
[Experiment] New OutputFormatRegistry

Christian-Schulz:
I created a OutputFormatRegistry as areplacment for emun OutputFormat.

The main goal is to allow the list to be dynamically expanded.  This enables plugins to support additional file types.

In `ImageSharpFileFormatHandler` you see this also allows for file types in the beta test.

To do this, OutputFormat had to be changed from enum to string. This would also require changing all configuration classes.
My first tests looked good.  

This isn't complete yet. 
@Lakritzator  Should I go on? What are your thoughts?


Lakritzator:
I like concepts like that, while working on the recipe functionality, I constantly come into the limitation of enums, when functionality can be extended. So up front, without looking at the details, I am up for it.

I will need to read through it, but my second though after the first part, was how does this handle a file format which has different extensions?

Examples:
- tiff, tif
- jpg, jpeg, maybe even jfif

Additionally, there is also the challenge of the mime-type, which is usually strongly coupled with it, I actually think it makes sense to force a coupling with it, for transports without having an extension.

Then there might be the issue with translations... although I do not know why I would "translated" the description.

Some other ideas:
Consider Implicit conversion, the OutputFormatDefinition can automatically cast to string if you like.
Consider records...

Might it make sense to have the wellknownformats actually an OutputFormatDefinition instead of a string?
Or why do we need those, if we have a registry? (just writing down my thoughts), do these have a bennefit?

Lakritzator:
Sorry about the conflicts... I'm going through the recipe code like crazy... fixing issues in recipes but also with some core parts in Greenshot here and there, as things were just "not complete".

I wish I had more AI tokens, already applied for Codex as an open source provider, maybe I get lucky. I am using gravity with my Google AI Pro account, but the tokens are somewhat limited, I hit my weekly limit after ~3-4 days.. I do think Gemini 3.8 is pretty good, but it is not perfect, sometimes has blind spots (Gemini: we can recognize that the user must install the extension XYZ. MY question: If it is not installed, how do we know about it? Gemini: ah yes... true.)

Lakritzator:
Btw.. "off topic", maybe not related to you directly, but I feel sorry to have to deny the cool PR 1203 for the border option, I am really in favor of community work and seeing a many contributors as possible, denying that makes my heart bleed... but it's just not something we can maintain.

Adding such functionality to the UI might seem like a good solution for the one "requirement" someone has, but I already saw that we have packed way to much in such ways. Recipes are the way out to provide the flexibility, but as spidermans uncle said, with great power comes great responsibility.

Need to consider how I can bring recipes for simple use-cases like the border, but will think about this in the coming days after the functionality is stable (I do not mean the implementation is bug free, just that the concept is "round").

Christian-Schulz:
I hadn't thought about mime types and different file extensions for the same file type.
I'll think about it over the next few days. Right now, I'm busy.

The only reason for using `WellKnownOutputFormats` was to eliminate magic strings in the code for the identitier.
There are many places in the code that responds to a specific file type.

`Description` should probably be `FullName` instead, and I had intended that to be the official name. It may not need to be translated at all.

Christian-Schulz:
Idea for simple use-cases like the border. 

new recipe-trigger  `editor menu`
new source acquisition or source type `current surface`
new destination `current surface`

```mermaid
flowchart LR
  a["recipe-trigger: editor menu"] --> b["source: current surface"]
  b--> c["effect: border"]
  c--> d["..."]
  d--> e["dest: current surface"]
```

Lakritzator:
Hmmm... I'm not sure if I get your idea, but if so... yes maybe?

I guess your comment is not "how to make recipes simple to use" but why not also use recipes to apply to the current capture in the editor? I see no reason why not make that possible... you mean a bit like a template? (A potentially **very** complex one, as I just saw me build a huge one)

(btw... I love mermaid, that's why I added a button to generate mermaid graphs for the recipe in the editor!)

Christian-Schulz:
OK, next time I'll quote more clearly.

> Need to consider how I can bring recipes for "simple use-cases like the border", but will think about this in the coming days after the functionality is stable (I do not mean the implementation is bug free, just that the concept is "round").

yeah, there where some feature request ala "I want store different effect settings and need a button in the toolbar so I can always select the one I need at the moment. I don't want to have to reset the settings every time"    
 
This should be now possible with recipes. We “only” need a new CaptureSourceStep to capture the data from the current editor and a new DestinationStep to apply it to the current editor.

(... yeah  I use mermaid every now and then when I'm analyzing and visualizing data structures with LLMs. But, when it comes to more complex cases, most viewers on GitHub, GitLab, and in VS Code extensions don't work properly.)

Lakritzator:
Just added your "editor trigger" with the latest push/merge (or Greenshot 1.4.227) now I am out of AI tokens for today and my weekly tokens are down to 13% 😢

I use Obsidian more and more, and everything I can visualize is better than nothing, mermaid works "pretty good" there.

Lakritzator:
Read the documentation, I changed the feature toggle in the ini

Lakritzator:
Oh no, forgot about the Issue 906 or more importantly PR 953, the "capture bar", this is also something which conflicts with the Recipe idea... The issue is okay, but the work that was put into 953... Need to think about that...

Lakritzator:
@Christian-Schulz I've added some crazy stuff, now you are able to add the annotation step and use an open editor or existing .greenshot file to define the annotations to add. So it deserializes the information and transform it to configuration for the annotation step. Running the recipe will then, as soon as the step is used, create the elements on the capture-surface.
I had to extend the step configuration and knowledge by a lot, but you can now (almost) configure every single detail.

Unfortunately the editor / drawable containers do not support anchors, so the cannot be "left" bound + 50 px. So this means the import only has absolute locations, but still...

Lakritzator:
Ugh, sorry for the conflicts...

PR #1380, opened 2026-09-16
Added recipe annotation import from open editors or a .greenshot file

Lakritzator:
Moving some editor expert features towards the editor settings elsewhere, introducing an checkbox to enable recipes in the expert settings.
Added a way to import the annotations from the current editor(s), or a .greenshot file.
Improved the way recipes can handle images, either externally linked on the file system (which is not the best idea, but makes it flexible), or embedded as base-64 which increases the filesize of the json a lot.


Lakritzator:
With this change, besides protentional bugs, the recipe feature is starting to be feature rich and (almost) complete for power users.

PR #1388, opened 2026-09-18
feat(installer): Add Portuguese (Brazil) installer translation

eduardomozart:
## Description
This PR adds the Portuguese (Brazil) translation to the Greenshot installer.

### Changes Made
- Mapped `ptBR` locale to LCID `1046` in `lcid.iss`.
- Added `ptBR` language utilizing the built-in `BrazilianPortuguese.isl` file in `setup.iss`.
- Added translated custom strings for installer options (plugins, start options, descriptions) in `setup.iss`, using correct Unicode encoding for accents.


Lakritzator:
That's cool, looks correct.
It does show that I need to think about placing those values elsewhere, the file is getting big.

PR #1390, opened 2026-09-18
feat: Add 'Edit with Greenshot' context menu for image files

### LINKED ISSUE #701: Add to CONTEXT MENU, (right click).
**What**
- Add to context menu, when selecting imgs.

**Why**
- I found that GSEditor works like a charm and covers most cases.

**Describe alternatives you've considered**
- `Open with ...` work, but to many steps.
- Also, tried open with clipboard when right-click taskbar GS icon.
- These steps works, but not intuitive.

**My quick-and-dirty solution**
`DOSKEY gs = greenshot`
- Just good-old CLI, but normal user??

**Additional context**
- My context-menu clutters, but only 1% of it are useful (Open with Code): 

Image

eduardomozart:
# Description
This PR adds a right-click context menu option to Windows Explorer that allows users to open any image file directly in the Greenshot image editor. Closes #701.

Because Windows 11 introduced a strict new architecture for top-level context menus, this implementation is split into two phases to ensure support across all Windows versions.

### 1. Legacy Registry Menu (Windows 10 & Win11 "Show more options")
* Modifies `setup.iss` to add `SystemFileAssociations\image\shell` registry keys.
* Automatically registers the `Greenshot.exe --openfile "%1"` command for all perceived image file types (PNG, JPG, BMP, etc.).

image

### 2. Modern Windows 11 Top-Level Menu
* **Architecture:** Microsoft requires top-level Windows 11 context menus to be backed by an `IExplorerCommand` COM server and an application package identity. 
* Adds `Greenshot.ShellExt`, a lightweight C++ COM DLL that implements the required `IExplorerCommand` interface.
* Uses an **External Location Package (Sparse Package)** via `AppxManifest.xml` to grant Greenshot app identity without fully converting the project to MSIX.
* Automatically registers and unregisters the sparse package via PowerShell during the Inno Setup installation/uninstallation process.
* C++ DLL is integrated directly into the `Greenshot.sln` and builds automatically via MSBuild in the existing GitHub Actions pipeline.

image

*Note: For local testing on Windows 11 (Top-Level Menu), Developer Mode must be enabled since the local build is not signed with the official certificate matching the Publisher in the AppxManifest.*

### Testing
- [x] Tested on Windows 10 (Legacy menu)
- [X] Tested on Windows 11 (Modern top-level menu)
- [X] Verified installer task and uninstaller cleanup (`Get-AppxPackage -Name "Greenshot.ShellExtension"`)

image


Christian-Schulz:
Awesome! 🚀
> Note: For local testing on Windows 11 (Top-Level Menu), Developer Mode must be enabled since the local build is not signed with the official certificate matching the Publisher in the AppxManifest.

What are the implications for the many users who are using the unsigned version and don't have developer mode enabled?
Does the menu item return an error, or is it simply not available?
Does the installer return an error when you select the option?

I just noticed that recently. The argument option `--openfile` doesn't exists. You can just leave that out. The app interprets this as an additional file path and simply ignores it. This option was removed in version 0.8. That was practically yesterday. 🤣

Lakritzator:
This is crazy, because I just wrote the following like ~4 hours ago:
[Architecture Decision Record - Unified Command Line & Custom Protocol Handling.md](https://github.com/user-attachments/files/32396828/Architecture.Decision.Record.-.Unified.Command.Line.Custom.Protocol.Handling.md)

In short, I also had the idea (and already a prototype) to use a simple C exe to handle the things I specified, and have it communicate via a named pipe in Greenshot. This solves the startup time and that a huge exe needs to hand over the information to running instance of itself.

I never considered this, I understand it would feel a bit more integrated in the shell, but it also has a bigger complexity. Don't know if the benefit exceeds the price for the complexity. Still I need to look at it, because I seem to recall that there were other benefits from an application identify.

Lakritzator:
Oh, having it as attachment is somewhat odd.. here as text:

## Architecture Decision Record - Unified Command Line & Custom Protocol Handling
**Status:** Accepted
**Date:** September 2026
**Context:** Greenshot Client Interoperability & IPC

#### 1. Context & Problem Statement

Greenshot needs to process external input from multiple Windows entry points:
- **Custom URL Protocol:** Links clicked in browsers, documentation, or internal tools using `greenshot://<action>?<params>`.
- **Command-Line Invocations (CLI):** File paths, capture commands, or automated flags passed from scripts or terminal sessions.
- **Shell Integration / "Open With":** File paths passed from File Explorer.
- Optionally **Native Messaging** from an browser extension.

Greenshot is a long-running, single-instance .NET desktop application. Launching a full .NET runtime instance for every URL click or CLI invocation introduces significant latency (cold-start penalties), wastes memory, and complicates single-instance mutex handover. Furthermore, command-line arguments and URL payloads in multi-user or Remote Desktop (RDS) environments must never cross Windows user session boundaries.

#### 2. Decision

We will implement a **dedicated, lightweight native Win32 C helper binary (`greenshot-proxy.exe`)** to act as the single ingestion gateway for custom URL schemes, command-line arguments, and file associations.

1. **Direct Entry Point:** All protocol registrations (`HKCR\greenshot`) and file-association verbs point directly to `greenshot-proxy.exe` rather than `Greenshot.exe`.
2. **Session-Isolated Named Pipe IPC:** The proxy communicates with the running Greenshot instance via a Windows Named Pipe scoped to the active caller's Security Identifier:
    `\\.\pipe\Greenshot_<UserSID>`
3. **Conditional Cold-Start:**
    - **If Greenshot is running:** The proxy connects to the pipe, writes the framed payload, and terminates immediately.
    - **If Greenshot is NOT running:** The proxy locates `Greenshot.exe`, launches it via `CreateProcessW`, waits for the Named Pipe to become ready (up to a bounded timeout), sends the payload, and terminates.

#### 3. Architecture & Execution Flow

```
                      [ External Trigger ]
            (greenshot:// link  OR  CLI / Open With)
                               │
                               ▼
                     greenshot-proxy.exe (C, <50 KB)
                               │
                 Resolve Caller Token / User SID
                               │
                 Attempt connection to:
              \\.\pipe\Greenshot_<UserSID>
                               │
             ┌─────────────────┴─────────────────┐
      [Pipe Exists]                       [Pipe Not Found]
             │                                   │
      Greenshot ALIVE                     Greenshot OFFLINE
             │                                   │
             │                           Spawn Greenshot.exe
             │                           Poll WaitNamedPipeW (up to 8s)
             │                                   │
             └─────────────────┬─────────────────┘
                               │
                       Write JSON Payload
                    [4-byte Len] + [UTF-8 JSON]
                               │
                       Proxy Terminates
```

#### 4. Wire Protocol Specification

Communication across `\\.\pipe\Greenshot_<UserSID>` uses a length-prefixed JSON envelope:
- **Prefix:** 4-byte unsigned integer (little-endian) representing payload byte count.
- **Payload:** UTF-8 encoded JSON object.

JSON

```
{
  "version": 1,
  "source": "url_scheme" | "cli" | "open_with",
  "raw_input": "greenshot://capture?target=window",
  "parsed": {
    "action": "capture",
    "parameters": {
      "target": "window"
    }
  }
}
```

#### 5. Registry Configuration

**Custom Protocol Handler (`greenshot://`):** Written by the installer into `HKLM` (or `HKCU` for per-user installations):

Code snippet

```
[HKEY_CURRENT_USER\Software\Classes\greenshot]
@="URL:Greenshot Protocol"
"URL Protocol"=""

[HKEY_CURRENT_USER\Software\Classes\greenshot\shell\open\command]
@="\"C:\\Program Files\\Greenshot\\greenshot-proxy.exe\" \"%1\""
```

**File Type Associations / "Open With":**

Code snippet

```
[HKEY_CURRENT_USER\Software\Classes\Applications\greenshot-proxy.exe\shell\open\command]
@="\"C:\\Program Files\\Greenshot\\greenshot-proxy.exe\" --file \"%1\""
```

#### 6. Security Considerations & Mitigations

- **Command Injection via URL Schemes:**
    Malicious web pages can craft arbitrary URLs (e.g., `greenshot://--arg-injection/path`).
    - _Mitigation:_ The proxy strictly treats `argv[1]` as raw input data. It performs zero shell execution or dynamic command building. It packages the raw string into the JSON envelope and delegates semantic parsing entirely to Greenshot core.
- **Cross-Session Spoofing & Snooping:**
    - _Mitigation:_ The pipe name incorporates `ConvertSidToStringSidW(TokenUser)`. Greenshot’s `NamedPipeServerStream` assigns an explicit DACL granting access solely to `TokenUser` and `SYSTEM`, blocking access from other user sessions or lower-integrity contexts.        
- **Denial of Service (Pipe Blocking):**
    - _Mitigation:_ If the pipe returns `ERROR_PIPE_BUSY`, the proxy calls `WaitNamedPipeW` with a strict 2000 ms timeout rather than waiting indefinitely.

#### 7. Consequences

- **Positive:**
    - Sub-millisecond proxy startup time ensures URLs and CLI commands feel instantaneous.
    - Eliminates multiple `.NET` runtime instances competing for locks or spawning duplicate tray icons.
    - Centralizes all external input validation into a single pipe listener inside Greenshot.
- **Negative:**
    - Introduces an additional native C binary that must be compiled, signed, and maintained in the deployment pipeline alongside the managed .NET solution.

Lakritzator:
I even already had a **prototyp** for the url handler 😂

```
#define WIN32_LEAN_AND_MEAN
#include <windows.h>
#include <shellapi.h>

#define PIPE_NAME "\\\\.\\pipe\\Greenshot_UrlHandler_Pipe"

// Helper to send data over the pipe
static BOOL TryForwardUrl(const char* url) {
    if (!WaitNamedPipeA(PIPE_NAME, 300)) {
        return FALSE;
    }

    HANDLE hPipe = CreateFileA(
        PIPE_NAME,
        GENERIC_WRITE,
        0,
        NULL,
        OPEN_EXISTING,
        0,
        NULL
    );

    if (hPipe == INVALID_HANDLE_VALUE) {
        return FALSE;
    }

    DWORD bytesWritten = 0;
    DWORD len = 0;
    while (url[len]) len++;

    BOOL success = WriteFile(hPipe, url, len, &bytesWritten, NULL);
    FlushFileBuffers(hPipe);
    CloseHandle(hPipe);

    return success;
}

// Launches Greenshot.exe from the current executable's folder
static BOOL LaunchGreenshot() {
    char exePath[MAX_PATH];
    DWORD len = GetModuleFileNameA(NULL, exePath, MAX_PATH);
    if (len == 0 || len >= MAX_PATH) return FALSE;

    // Strip "Greenshot.UrlHandler.exe" to get the directory
    for (int i = (int)len - 1; i >= 0; i--) {
        if (exePath[i] == '\\') {
            exePath[i + 1] = '\0';
            break;
        }
    }

    // Append "Greenshot.exe"
    char targetPath[MAX_PATH];
    lstrcpyA(targetPath, exePath);
    lstrcatA(targetPath, "Greenshot.exe");

    SHELLEXECUTEINFOA sei = { sizeof(sei) };
    sei.fMask = SEE_MASK_FLAG_NO_UI;
    sei.lpFile = targetPath;
    sei.nShow = SW_SHOWNORMAL;

    return ShellExecuteExA(&sei);
}

int WINAPI WinMain(HINSTANCE hInstance, HINSTANCE hPrevInstance, LPSTR lpCmdLine, int nCmdShow) {
    // 1. Trim leading whitespace/quotes
    char* urlArg = lpCmdLine;
    while (*urlArg == ' ' || *urlArg == '\t') urlArg++;

    if (*urlArg == '\0') {
        return 0;
    }

    // 2. Loop until sent or user gives up
    while (!TryForwardUrl(urlArg)) {
        int choice = MessageBoxA(
            NULL,
            "Greenshot does not appear to be running.\n\n"
            "Would you like to start Greenshot and process this link?",
            "Greenshot",
            MB_YESNO | MB_ICONQUESTION | MB_DEFBUTTON1 | MB_TOPMOST
        );

        if (choice == IDYES) {
            LaunchGreenshot();

            // Wait for Greenshot to start and open its pipe (poll up to 10 seconds)
            BOOL connected = FALSE;
            for (int i = 0; i < 20; i++) {
                Sleep(500);
                if (TryForwardUrl(urlArg)) {
                    connected = TRUE;
                    break;
                }
            }

            if (connected) {
                return 0; // Success
            }

            // If it timed out, ask if the user wants to retry
            int retryChoice = MessageBoxA(
                NULL,
                "Greenshot started, but was not ready in time.\n\n"
                "Would you like to try connecting again?",
                "Greenshot",
                MB_RETRYCANCEL | MB_ICONWARNING | MB_DEFBUTTON1 | MB_TOPMOST
            );

            if (retryChoice != IDRETRY) {
                return 1;
            }
            // If IDRETRY, the while loop runs again
        }
        else {
            // User selected NO / Cancel
            return 1;
        }
    }

    return 0;
}
```

Lakritzator:
I at least find the concept very fascinating, but I need to think it through.
What we do need to do, implement that Pipe communication, that might make it easier for a C(PP) exe to communicate with the running instance... so I will look at that first.

Lakritzator:
You should really stop using force pushes to the repository, this breaks a lot of processes which run in the background.

eduardomozart:
> Awesome! 🚀
> 
> > Note: For local testing on Windows 11 (Top-Level Menu), Developer Mode must be enabled since the local build is not signed with the official certificate matching the Publisher in the AppxManifest.
> 
> What are the implications for the many users who are using the unsigned version and don't have developer mode enabled? Does the menu item return an error, or is it simply not available? Does the installer return an error when you select the option?
> 
> I just noticed that recently. The argument option `--openfile` doesn't exists. You can just leave that out. The app interprets this as an additional file path and simply ignores it. This option was removed in version 0.8. That was practically yesterday. 🤣

@Christian-Schulz 
To answer your questions about the unsigned/non-developer mode implications: I added an installation fallback (Legacy mode) for users with local builds or with Developer Mode disabled. This installs only the classic registry-based context menu (which requires the user to click "Show more options" on Windows 11). If the user is running an official (digitally signed) build, or if they have Developer Mode enabled, the installer will automatically register the modern Windows 11 Explorer extension.

Also, thanks for the heads up about the `--openfile` argument! In this updated architecture, we don't pass any command-line arguments to `Greenshot.exe` when spawning it. The proxy just sends the file path directly to the running instance via the Named Pipe!

eduardomozart:
> This is crazy, because I just wrote the following like ~4 hours ago: [Architecture Decision Record - Unified Command Line & Custom Protocol Handling.md](https://github.com/user-attachments/files/32396828/Architecture.Decision.Record.-.Unified.Command.Line.Custom.Protocol.Handling.md)
> 
> In short, I also had the idea (and already a prototype) to use a simple C exe to handle the things I specified, and have it communicate via a named pipe in Greenshot. This solves the startup time and that a huge exe needs to hand over the information to running instance of itself.
> 
> I never considered this, I understand it would feel a bit more integrated in the shell, but it also has a bigger complexity. Don't know if the benefit exceeds the price for the complexity. Still I need to look at it, because I seem to recall that there were other benefits from an application identify.

@Lakritzator
Thank you for sharing that ADR! I've actually adapted this PR to closely follow your exact specifications from ADR 002, and it really turned out to be the cleanest way to implement this feature.

Because the modern Windows 11 shell extension is forced to execute inside an isolated MSIX AppContainer context (due to the sparse package identity), the only reliable way to interact with the main Greenshot instance running on the user's desktop is through Named Pipes. 

Using the lightweight C++ `greenshot-proxy.exe` as the `ExeServer` solves both the cold-start penalty and the sandbox restrictions. If Greenshot isn't running, the proxy escapes the AppContainer (by delegating the launch to the desktop shell to spawn the heavy .NET app outside the sandbox), waits for the `\\.\pipe\Greenshot_<SID>` named pipe to become available, sends the JSON envelope, and immediately exits. It perfectly aligns with your vision of a unified, lightweight ingestion gateway!

PR #1396, opened 2026-09-19
New CaptureSource and destination for current editor

Christian-Schulz:
I've expanded the possibilities using the editor trigger.

The following is now possible
- use current editor as souce and any destination (incl. current editor)
- use any source and replace surface in current editor
- use any source and any destination
- using a recipe without source
- using a recipe without destination
- using a recipe without source and without destination

I changed the editor destination step parameter ReuseEditor (bool) to TargetEditor (enum)
possible values are `NewEditor`, `AvailableEditor` and `CurrentEditor`.
`AvailableEditor` is the replacement for `ReuseEditor==true`

I added an new CaptureSourceType `CurrentEditor`.
The `CurrentEditorCaptureSource` replaces the need to set the `context.Payload` in the trigger.

If the recipe has no soucre and no destination the trigger creates it at all start nodes/all end nodes.

image


Lakritzator:
Hey, I'll need to look at it tomorrow, way past midnight for me... 
But before this change, if I take a recipe and give it the editor menu as trigger, I was able to add stuff to the current editor, so you can select it from the editor menu.

You should even be able open a new one. What is different with your solution?

image

That would add a text on the current surface in the editor.

Christian-Schulz:
Youre example is still possible.
I reimplemented it in a different way

My solution is more flexible.
In your solution, the current surface was passed as payload for each recipe. That made it difficult to use other sources with editor trigger.

To be honest, I'd say that a capture source must always generate a new payload, even if a payload was provided as input.
(it is called "source")
The clipboard trigger does not currently follow this rule.

Lakritzator:
My kid (who had special needs) is currently draining my energy, can't concentrate... (Very loud, obnoxious) Having a bad phase, was first school week.

I feel like you are saying we need to formalize something to define using the current editor as source,and also as destination. Which I think sounds good.  But then I'm reading something about automatically doing something...

Before I get my brain energy back, and hope to be able to understand what your suggestion is, I wanted to mention that my plan was to be able to have a recipe call others almost like a subroutine. That's why the trigger should have been able to supply the payload, and I had a recipe which only placed annotations to test with (no source, not even a destination). I didn't think it through yet. Maybe you can think about that, if your change can make that possible?

Lakritzator:
Btw we need to be careful with what we do, I've been doing work to gate (protecte) the recipes, providing the users with a "before we enable this, this is what it does. Are you okay with that". Working together with a security advisor, he pointed out that we can currently use a variable to specify the destination, thus changing it at runtime, which fails the gate.

Christian-Schulz:
Take your time , family comes first.

( Just to make sure we're on the same page. In the recipe, they're called “nodes,” and in the pipeline, they're called “steps” ,  both are exactly the same thing, right? )

I use the recipe Editor to vizualize my thoughts and use nodes.

image

I would formalize the rules for recipes/piplines a little more.
- Every pipeline must have at least one active source node
- Every pipeline must have at least one active destination node
- In a recipe, it's allowed to leave out the source node and/or destination node.
- If there is no source node in the recipe the trigger manager created on and connected it to all start nodes in den recipe
- If there is no destination node in the the trigger manager created on and connected it to all end nodes in den recipe

other cases:
(no active destination)
image
(with active destination and active source)
image

- If the trigger manager adds a new source node. This would be e new  'TriggerCaptureSoure` . It determines which is the default source based on the trigger type and set the context.payload. For example, for a clipboard trigger, the clipboard would be the source and  for an editor trigger, it would be the current editor, and so on.
---
More rules:
- Only source nodes are allowed to create a new context.payload
- Every source node creates a new context.payload
- No source node should rely on a pre-supplied payload. Or in other words, a thing "pre-supplied payload" does not exists for source nodes. 
- Every other node type wich want to process the payload have to skip if no payload are available. 


> I wanted to mention that my plan was to be able to have a recipe call others almost like a subroutine. That's why the trigger should have been able to supply the payload, and I had a recipe which only placed annotations to test with (no source, not even a destination).

I'm with you. A subroutine is something that performs a task. To do so, it requires an input and an output. 
You use Context.Payload as both the input and the output. This conflicts with other source nodes in the pipeline and is not flexible enough to control the destination.

Are you familiar with [Blender's node system](https://docs.blender.org/manual/en/latest/interface/controls/nodes/index.html)? The smallest valid system there consists of one input node and one output node.

Maybe we should do something similar here. That means your example recipe with just one annotation node wouldn't be allowed.

Lakritzator:
For me a recipe is the whole thing, that which in the end lands in the .recipe.json
The nodes are the steps, and then there are the pins (inbound / outbound) and finally the connectors are the lines.

I think it's just messed up in my mind, as I probably overread some word... due to concentration issues, as explained.
So "If there is no source node in the recipe the trigger manager created on and connected it to all start nodes in den recipe"
Where does it create this? In memory? I was just trying to understand what you mean with that something is created automatically, is this just your way of making the code consistent, or something to communicate to the user and it will be visible? I guess you mean the first, you are formalizing "what" is actually responsible for creating the payload. Our recipe system will use defaults to create the responsible thing, and the thing will do the right work, in a better definable state? Is it just about refining the rules, or is it about the editor?

Asked differently:
Does this mean, that someone can still draw a single step to add annotations, and define the editor trigger to make it available there? Or do they have to do more? 

I actually understand that you want to refine the rules, and that for the user practically nothing changes...
If that is correct, then I think we are on the same level, if not, I need to see a doctor 😂

Christian-Schulz:
> Does this mean, that someone can still draw a single step to add annotations, and define the editor trigger to make it available there?

Yes

> Or do they have to do more?

No

> Where does it create this? In memory?

Yes

> is this just your way of making the code consistent

Yes

> Our recipe system will use defaults to create the responsible thing, and the thing will do the right work, in a better definable state? 

Yes


> Is it just about refining the rules,

yes

>  or is it about the editor?

No, the only changes in the Editor are the new/changed Parameters


My current changes are only doing this for editor trigger. 
Next, I'll generalize that and implement it in the Trigger Manager.
Give me a little time.

Christian-Schulz:
I think it's ready.

I've generalized the preparation for all recipes that are executed by a trigger.
The trigger manager adds a source node  and destination nodes based on the trigger type. It only adds it if nessesary.

Rules for source node:

``` 
var sourceType = trigger.TriggerType switch
{
    TriggerConfig.TypeClipboard => CaptureSourceType.Clipboard,
    TriggerConfig.TypeEditor => CaptureSourceType.CurrentEditor,
    _ => CaptureSourceType.FullScreen
};
return RecipeStepConfig.CreateSource(sourceNodeId, sourceType);
```

The default source uses the capture mode from user settings.

Rules for destination:

```
switch (trigger.TriggerType)
{
    case TriggerConfig.TypeEditor:
    {
        var editorNode = RecipeStepConfig.CreateEditor(editorNodeId);
        editorNode.Set("TargetEditor", nameof(TargetEditor.CurrentEditor));
        return editorNode;
    }
    default:
    {
        var editorNode = RecipeStepConfig.CreateDestinations(editorNodeId);
        return editorNode;
    }
}
```
The default destination uses the configured destination from user settings.

---
I also added the same mechanism for the test button in the recipe editor. The button behave now exactely like the context menu trigger.  This mean you can now test run a recipe with a single annotation node.


----
I've also simplified how the triggers are called. The triggers now simply call their own `Fire()` method and pass all the necessary parameters to it, if there are any.

I testet editor- , context- , hotkey-  and clipboard-trigger.

src/Greenshot.Base/Core/Enums/TargetEditor.cs
```diff
@@ -0,0 +1,30 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+namespace Greenshot.Base.Core.Enums
+{
+    public enum TargetEditor
+    {
+        NewEditor,
+        AvailableEditor,
```
Lakritzator, line 27: It would be nice to understand what the difference between "AvailableEditor" vs "CurrentEditor" is. Not a blocker...

src/Greenshot.Base/Recipes/CaptureRecipe.cs
```diff
@@ -171,6 +171,16 @@ public bool HasEditorDestination()
             return false;
         }
 
+        /// <summary>
+        /// Determines whether the recipe contains any source step.
+        /// </summary>
+        public bool HasSourceStep()
```
Lakritzator, line 177: How would this "scale" to extensions providing sources? (Not a blocker!)

src/Greenshot.Base/Recipes/CaptureSourceType.cs
```diff
@@ -33,6 +33,7 @@ public enum CaptureSourceType
         LastRegion,
         Clipboard,
         File,
-        TextOcr
+        TextOcr,
+        CurrentEditor
```
Lakritzator, line 37: CurrentEditor vs ActiveEditor? (No blocker!)

src/Greenshot.Base/Triggers/IEditorTrigger.cs
```diff
@@ -32,5 +29,7 @@ public interface IEditorTrigger : ITrigger
         string MenuItemText { get; set; }
         string Group { get; set; }
         int Order { get; set; }
+
+        void Fire(object editorForm);
```
Lakritzator: Just reading the interface makes me wonder, what does "fire" actually achieve? (no blocker!)

src/Greenshot.Base/Triggers/TriggerRecipePreparer.cs
```diff
@@ -0,0 +1,198 @@
...
+
+        /// <summary>
+        /// Ensures that the given recipe has at least one start source node.
+        /// If the recipe does not have a source node, a trigger-specific source node is added
+        /// and connected to the effective start nodes of the given recipe.
+        /// It does not modify the recipe if it already has a source node.
+        /// </summary>
+        /// <param name="recipe">The recipe to inspect.</param>
+        /// <param name="trigger">The trigger that determines the source type.</param>
+        /// <returns>The modified recipe with the required start source node added, or the original recipe if no changes were needed.</returns>
```
Lakritzator: One thing that always triggers me in a code review is a method which takes an "object graph" / class instance, and returns the same type... does cause side-effects by modifying the or cleanly return a clone. You return a cloned object, which is correct. But the documentation says it returns the "modified recipe", which is somewhat misleading. Same for the EnsureRequiredDestination

src/Greenshot.Tests/Recipes/TriggerIntegrationTests.cs
```diff
@@ -108,29 +108,5 @@ public void CaptureRecipe_HasDestinationStep_And_HasEditorDestination_DetectsCor
             Assert.True(recipeWithDestinationsEditor.HasEditorDestination());
         }
 
-        [Fact]
-        public async Task SourceAcquisitionStep_WhenPayloadAlreadyProvided_SkipsAcquisition()
```
Lakritzator, line 112: Can you explain why you deleted this? (no blocker!)
Moved to TriggerRecipePreparerTests?

src/Greenshot/Forms/MainForm.cs
```diff
@@ -771,13 +771,7 @@ private void UpdateRecipesMenu()
 
                 item.Click += (s, ev) =>
                 {
-                    Dispatcher.CurrentDispatcher.BeginInvoke(() =>
```
Lakritzator, line 774: So you removed the dispatcher, and replaced the execute with Fire, but inside fire there is no dispatcher etc... sure this is correct? I needed this, otherwise I wouldn't have added it.... but I trust you tested most, and we can always add it later, so no blocker there...

src/Greenshot/Triggers/EditorTrigger.cs
```diff
@@ -67,9 +70,15 @@ public override void Stop()
         /// <summary>
         /// Invoked when the user clicks this recipe item in the Image Editor menu.
         /// </summary>
-        public void Fire(IDictionary<string, object> parameters = null)
+        /// <param name="editorForm">The instance of the editor form where the trigger was fired</param>
+        public void Fire(object editorForm)
```
Lakritzator: I'm a bit confused about the different "Fire" method implementations, ContextMenuTrigger.Fire used to have a Dictionary, not anymore (Fire()) and this also had one, not anymore (Fire(object)). I also do not like "object", so if we can use at least an interface here?

Lakritzator:
So the only thing which slightly triggered me in the code review, was the Fire inconsistency, and that one of them uses object.

src/Greenshot.Base/Core/Enums/TargetEditor.cs
Christian-Schulz: CurrentEditor only makes sense if the pipeline is started by an editor trigger. Because only in this case a dedicated editor instance is present the pipeline. 
In all other cases it would fallback to AvailableEditor.

AvailableEditor is the same as the greenshort setting `ReuseEditor==true` . 
Which checks all open editors and looks for the first with a surface wich is unmodified.
If no editor is found during this process, a new one will be opened.

src/Greenshot.Base/Recipes/CaptureSourceType.cs
Christian-Schulz: same like  in enum TargetEditor.
CurrentEditor only makes sense if the pipeline is started by an editor trigger. 

But in the case of source acquisition there is no fallback and the pipeline will be canceled.

src/Greenshot.Base/Triggers/IEditorTrigger.cs
Christian-Schulz: That is the function that ultimately triggers the trigger and starts the pipeline. 
It contains all the parameters that the specific trigger requires.

src/Greenshot.Tests/Recipes/TriggerIntegrationTests.cs
Christian-Schulz: I changed the behavior of the step. The step is no longer skipped.

src/Greenshot/Forms/MainForm.cs
Christian-Schulz: Hmmm, I think the trigger dont have to be run on this context, only some of the internal function wich this actually need. 
If this break, the error is somewhere else. Let me do some test more thats interacts with the UI.

Christian-Schulz:
| So the only thing which slightly triggered me in the code review, was the Fire inconsistency, and that one of them uses object.

My idea was that the caller of the trigger shouldn't need to know about the internal implementation. In other words, that there is a dictionary of parameters there.
The `Fire()` method defines everything necessary for the specific trigger.
That is also why there are different parameters for different trigger types.
The Fire method also checks whether all required parameters are valid, otherwise, it terminates immediately.

I wanted to be type safe, but the FormEditor was in a different namespace / project.  But you're right, I could use the interface.

Lakritzator:
One thing I didn't test or look at, is something I included initially but forgot about, a way to override the build in recipes. Did you already do that?

src/Greenshot/Forms/MainForm.cs
Christian-Schulz: i reverted it, it could potetianally happen that the context menu closes after the capture.

Christian-Schulz:
> One thing I didn't test or look at, is something I included initially but forgot about, a way to override the build in recipes. Did you already do that?

I've seen it, but I've never tried it.

src/Greenshot.Base/Recipes/CaptureRecipe.cs
Christian-Schulz: yes, but only if they use the step type "Source".

Lakritzator (approved):

Lakritzator:
Okay, I should really test this, as this is one of the needed features, enhance but now change the build in functionality. Maybe I should just make it simpler, like copy from and disable the original flow...

